PR #3, opened 2019-11-07
Add basic API reference docs generation

munnerz:
Adds initial work for the reference docs generation. This will likely make `scripts/verify-release` fail as I'm going to have to update the go templates used for generating the docs to generate markdown, and generally fix them up 😄

### DIFF
--- a/scripts/gendocs/generate
+++ b/scripts/gendocs/generate
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Jetstack cert-manager contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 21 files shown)


munnerz:
All passing 🎉  I think it's worth merging this as-is so we can improve on the whole site once we've got some content 😄

scripts/gendocs/generate
```diff
@@ -0,0 +1,77 @@
...
+		-template-dir "${REPO_ROOT}/scripts/gendocs/templates" \
+		-api-dir "./pkg/apis" \
+		-out-file "${REPO_ROOT}/content/${outputdir}/reference/api-docs.md" \
+		"$@"
+	rm -rf vendor/
+	popd
+}
+
+genversion "master" "docs"
+# genversion "release-0.11" "v0.11-docs"
```
JoshVanL: Want to clean this up?

munnerz: Left it there and added a comment explaining to use it as a template when adding new versions 😄

JoshVanL: :+1:

PR #18, opened 2019-11-13
Adds configuration acme dns01 webhook

JoshVanL:
Signed-off-by: JoshVanL <vleeuwenjoshua@gmail.com>
/assign @munnerz


munnerz:
It seems like this info is duplicated? https://deploy-preview-18--cert-manager.netlify.com/docs/configuration/acme/dns01/#webhook

It may be best to change the section on the DNS01 page to just list the supported external providers and link to the further docs page which is more geared around actually configuring the dns01 provider type with an example using maybe smallstep?

PR #59, opened 2019-12-12
Update certificate.md

michaelgregson:
keyAlgorithm value should be lowercase

Signed-off-by: Michael Gregson <michael.gregson@u-techsystems.com>


JoshVanL:
Thanks for that :smile: 
/lgtm
/approve
/ok-to-test

PR #60, opened 2019-12-12
CM doesn't refresh token used to authenticate against Vault

pires:


munnerz:
🤦‍♂ looks like this was fixed in #61, which was submitted shortly after your own PR... Thanks though 😄

PR #68, opened 2019-12-19
Add helm-3+ install cmd

therealkevinard:
Helm 3 has a slightly different `install` syntax. Could send newcomers down a confusing path.

### DIFF
--- a/content/en/docs/installation/kubernetes.md
+++ b/content/en/docs/installation/kubernetes.md
@@ -133,12 +133,20 @@ $ helm repo update
 
 Install the cert-manager Helm chart.
 ```bash
+# Helm < v3
 $ helm install \
   --name cert-manager \
   --namespace cert-manager \
   --version v0.13.0-alpha.0 \
   jetstack/cert-manager
+  
+# Helm v3+
+$ helm install \
+  cert-manager jetstack/cert-manager \
+  --namespace cert-manager \
+  --version v0.12.0
 ```
+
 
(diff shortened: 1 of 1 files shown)


meyskens:
Fixes #103

munnerz:
Can you also update the installation guide's full body/content? There's still a pre-req for Tiller to be installed, and we still have a Foreword that goes further into this (and why it needs to be installed): https://github.com/cert-manager/website/blob/master/content/en/docs/installation/kubernetes.md#installing-with-helm

It'd be good to also link to the "Installing Helm" guide still though, in case anyone hasn't got it set up on their local machine or is not familiar with concepts and terms we reference 😄 https://helm.sh/docs/intro/install/

therealkevinard:
Advice on content flow: should helm 3 be assumed for the full body, or should the alternate steps be noted inline separately? 

My thought: someone intentionally using helm 2 is almost certainly at least a little seasoned (eg: most likely, H2 is a business requirement from some months ago), so maybe a callout/blockqoute with a heads-up would be enough for 2, with the rest of the content assuming 3. Thoughts?

munnerz:
This has now been fixed in #110 - thanks for the PR though! 😄

munnerz:
> Advice on content flow: should helm 3 be assumed for the full body, or should the alternate steps be noted inline separately?

We've opted to take this approach and make Helm 2 the 'special case' - if you take a look at the changes in that PR, and if you have any feedback/comments/amendments please open another PR!

PR #73, opened 2020-01-06
Update ingress.md

adamgajzlerowicz:
update to version 3


content/en/docs/tutorials/acme/ingress.md
```diff
@@ -90,7 +90,7 @@ You can get more details about `nginx-ingress` and how it works from the
 Use `helm` to install an NGINX Ingress controller:
 
 ```bash
-$ helm install stable/nginx-ingress --name quickstart
+$ helm install quickstart stable/nginx-ingress
```
munnerz: Could you add a note here to say:

```
# If you are using Helm v2, use:
helm install stable/nginx-ingress --name quickstart
```

Else people may get tripped up if they are running older Helm versions! 😄

PR #88, opened 2020-01-20
Update kubernetes installation page

nirgn975:
In Helm v3, the release name is now mandatory as part of the command.
The `name` flag is not exist anymore.

### DIFF
--- a/content/en/docs/installation/kubernetes.md
+++ b/content/en/docs/installation/kubernetes.md
@@ -134,7 +134,7 @@ $ helm repo update
 Install the cert-manager Helm chart.
 ```bash
 $ helm install \
-  --name cert-manager \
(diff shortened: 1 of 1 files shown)


meyskens:
@nirgn975 could we maybe add a note that `--name` is needed when using Helm 2, I imagine a lot of people still are on 2

munnerz:
This is a dupe of #68 and #75, and my comments over there also apply here 😄 https://github.com/cert-manager/website/pull/68#issuecomment-583324899 may be worth working out which PR should be pushed through and going from there :) I don't mind either way but we do need to update the full install guide as part of migrating to Helm 3

munnerz:
This has now been fixed in #110 - thanks for the PR though! 😄

PR #93, opened 2020-01-25
Fixing duraton -> duration typo

dhermes:


dhermes:
`/approve`

content/en/docs/concepts/certificaterequest.md
```diff
@@ -27,7 +27,7 @@ spec:
   - signing
   - digital signature
   - server auth
-  duraton: 90d
+  duration: 90d
   issuerRef:
     name: ca-issuer
```
dhermes: While we're at it, the line

> ... attempt to request the `Issuer` `letsencrypt-prod` in the default issuer group `cert-manager.io` ...

seems to be incorrect. Should it not refer to `ca-issuer` (from line 32) as the issuer?

munnerz: :+1: yep it should - nice spot 😄

PR #96, opened 2020-01-31
Fixed variable in AzureDNS documentation

Gerrit-K:


munnerz:
Thanks for this PR 😄 

I'm not too familiar with Azure's API/CLI/things needed to be configured - could you provide a description as to what changes you've made and why?

/assign @Gerrit-K

Gerrit-K:
Of course! In my first commit, I just wanted to fix an echo output where the label was the subscription id but the output was incorrectly set to the password of the created service principal. 

After fixing that to display the subscription id, I realized that the variable actually wasn't available. In my second commit, I therefore added a command to initially retrieve that value and slightly adjusted a text further down the page.

Hope that helps.

/assign @munnerz

munnerz:
Cool, thanks makes sense now that I look again..

Thank you!

/lgtm
/approve
/ok-to-test

PR #101, opened 2020-02-04
Fix typo "singing" -> "signing"

ldealmei:
Quick typo fix in the docs.


ldealmei:
@meyskens do I create another PR with the dco, or do you prefer I wait someone to override the dco?

PR #102, opened 2020-02-04
Add key field under privateKeySecretRef

hytromo:
I am not 100% sure, but it worked for me. As a newbie this sent me on a search whether k8s secrets can hold only values without keys, but I don't think it's possible.

In other parts of the documentation the name *and* key are specified on secret refs, like here: https://cert-manager.io/docs/configuration/acme/dns01/google/#create-an-issuer-that-uses-clouddns

### DIFF
--- a/content/en/docs/configuration/acme/_index.md
+++ b/content/en/docs/configuration/acme/_index.md
@@ -61,9 +61,8 @@ spec:
     email: user@example.com
     server: https://acme-staging-v02.api.letsencrypt.org/directory
     privateKeySecretRef:
-      # Secret resource used to store the account's private key.
-      name: example-issuer-account-secret
-      key: example-issuer-account-secret-key
+      # Secret resource that will be used to store the account's private key.
+      name: example-issuer-account-key
     # Add a single challenge solver, HTTP01 using nginx
     solvers:
     - http01:
--- a/content/en/docs/configuration/acme/_index.md
+++ b/content/en/docs/configuration/acme/_index.md
@@ -62,7 +62,8 @@ spec:
     server: https://acme-staging-v02.api.letsencrypt.org/directory
     privateKeySecretRef:
       # Secret resource used to store the account's private key.
-      name: example-issuer-account-key
+      name: example-issuer-account-secret
+      key: example-issuer-account-secret-key
     # Add a single challenge solver, HTTP01 using nginx
     solvers:
     - http01:


hytromo:
I saw the DCO requirement, I hope it's not required for something so trivial.

munnerz:
This field isn't actually required and we default to `tls.key` if it's not provided. The option is left here to allow people to override it if they explicitly need to.

I am not sure documenting it here is that useful given the actual `privateKeySecretRef` is also auto-generated anyway. What led you to actually *need* to set this value?

> I saw the DCO requirement, I hope it's not required for something so trivial.

For minor changes like this, nope - it can be manually overridden when we come to merging 😄

hytromo:
> I am not sure documenting it here is that useful given the actual `privateKeySecretRef` is also auto-generated anyway. What led you to actually _need_ to set this value?

As far as I understand from the documentation, I need to have a Secret storing the private key of my let's encrypt account (as stated in the comment `# Secret resource used to store the account's private key.`). I can't see how this Secret can be autogenerated, I am probably missing something.

What I currently do is generate a Secret resource with a specific `name` and `key` that holds my account's private key and then I generate an Issuer where I update the `privateKeySecretRef` with that name and key.

This is my Secret resource (helm template)

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: {{ .Values.secrets.letsEncryptPrivateKey.name }}
  labels:
    release: {{ .Release.Name | quote }} 
type: Opaque
data:
  {{ .Values.secrets.letsEncryptPrivateKey.key }}: |-
    {{ .Files.Get "files/domain-name.com.key" | b64enc }}
```

(I got this private key through acme.sh, can I make cert-manager generate it itself?)

And this is a snippet from my Issuer resource:
 ```yaml
   privateKeySecretRef:
      name: {{ .Values.secrets.letsEncryptPrivateKey.name }}
      key: {{ .Values.secrets.letsEncryptPrivateKey.key }}
```

The current setup works well without issues, but I may be doing some extra steps that can be automated?

munnerz:
> I can't see how this Secret can be autogenerated, I am probably missing something.

So cert-manager will actually generate this private key for you and store it in the named secret if it doesn't exist. You *can* do what you've done, but it does involve additional steps you'd otherwise not need.

Instead of this patch, could you perhaps add a comment to this field that explains that the private key will be automatically generated/managed if it doesn't already exist? Basically anything that would have helped you to not get into this situation in the first place 😅

PR #104, opened 2020-02-05
Add search

meyskens:
This adds Algolia based search to the website.
It currently works but needs to have some style and functionality improvements before being able to be used.

TODO:
- [x] Style search box
- [x] Add Algolia logo in results (requirement on the free plan)
- [x] Set up indexer in Netifly
- [x] Add more context to rendered results
- [x] fix missing JS on home page
- [x] fix missing links in some search results
- [x] Support multi version


meyskens:
All that is left is setting it up in the run in Netifly, something that @munnerz needs to do (can you make sure it doesn't run on PR builds)

munnerz:
So I'm looking into how to securely set this up in Netlify so that the key is only available during non-PR builds, so that PRs created against this repo can't be used to leak the key.

I came across this: https://community.netlify.com/t/common-issue-using-environment-variables-on-netlify-correctly/267/5

So it seems there isn't a native way to allow this via Netlify.

They *do* expose a `PULL_REQUEST` var that can be used to determine if a build is a pull request, but that does not address the security concerns as anyone could remove this check.

The next best thing I can think of is to encode a 'bootstrap script' into the Netlify build settings 'command' box, which is responsible for either unsetting the environment var (which may still not completely mask the credentials properly, there could be remnants left over that are accessible from within the build) or otherwise somehow fetching/decrypting the secret key. Both of these sound quite non-ideal...

As an alternative, we could set up a postsubmit or periodic job in Prow to rebuild the index from our own infrastructure separately.

munnerz:
Another option would be a webhook that gets called by Netlify after a build has completed... https://docs.netlify.com/site-deploys/notifications/#outgoing-webhooks-and-notifications

meyskens:
CI job added in https://github.com/jetstack/testing/pull/330

PR #106, opened 2020-02-06
Remove commonName where possible

meyskens:
This removes the use of commonName in favour of adding it's value to dnsNames.
Where it doesn't seem right to not meantion commonName a note is added.

Signed-off-by: Maartje Eyskens <maartje@eyskens.me>


meyskens:
Fixes https://github.com/jetstack/cert-manager/issues/2550

PR #109, opened 2020-02-10
Update ingress.md to improve readability

neverping:
This commit will improve readability on ingress.md, because one of the sample output  was not properly escaped during the 'step 2' stage.

And in another output I have omitted part of the text because it was not finishing up the "{" statement and it was producing some visual garbage on the live website.

### DIFF
--- a/content/en/docs/tutorials/acme/ingress.md
+++ b/content/en/docs/tutorials/acme/ingress.md
@@ -132,7 +132,7 @@ quickstart-nginx-ingress  0s
 ==> v1/Service
 quickstart-nginx-ingress-controller       0s
 quickstart-nginx-ingress-default-backend  0s
-
+```
 
 NOTES:
 The nginx-ingress controller has been installed.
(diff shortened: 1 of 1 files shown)


content/en/docs/tutorials/acme/ingress.md
```diff
@@ -132,7 +132,7 @@ quickstart-nginx-ingress  0s
 ==> v1/Service
 quickstart-nginx-ingress-controller       0s
 quickstart-nginx-ingress-default-backend  0s
-
+```
```
munnerz: I think these back-ticks are meant to exist *after* the `NOTES` section, as the NOTES are output from the `helm install` command 😄

neverping: Oh, I see it now! I had the impression that was not escaped.

munnerz:
Thanks! 😄 

/lgtm
/approve

PR #110, opened 2020-02-11
Update Helm install instructions

### LINKED ISSUE #107: Restructure Helm installation
We should restructure the Helm installation documentation to note that helm 3 exists but still keep the Tiller related notes somewhere in there

meyskens:
This updates all install instructions to Helm v3 with keeping a side note on Helm v2

Fixes #107

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -139,3 +139,5 @@ WIP
 x509
 YAML
 YAMLs
+v2
+v3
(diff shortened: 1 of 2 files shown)


munnerz:
Looks like this is failing due to `v2`, `v3` etc not being valid words according to our spellchecker:

```
    content/en/docs/installation/kubernetes.md
       83 | - Helm v2 or v3 installed 
       83 | - Helm v2 or v3 installed 
       85 | ### Note: Helm v2 
       87 | loying cert-manager with Helm v2, you must ensure 
```

We can either add exceptions for these two, or otherwise wrap with backticks so the checker quietens down 😄

PR #115, opened 2020-02-13
Add custom fields for venafi

meyskens:
This adds a issuer specific annotation section for vanafi explainging the custom fields annotation

### DIFF
--- a/content/en/docs/configuration/venafi.md
+++ b/content/en/docs/configuration/venafi.md
@@ -177,7 +177,7 @@ Starting `v0.14` you can pass custom fields to Venafi (TPP version `v19.2` and h
 The value is a JSON encoded array of custom field objects having a `name` and `value` key.
 For example:
 ```yaml
-apiVersion: cert-manager.io/v1alpha3
+apiVersion: cert-manager.io/v1alpha2
 kind: Certificate
 metadata:
   name: example-com-certificate
--- a/content/en/docs/configuration/venafi.md
+++ b/content/en/docs/configuration/venafi.md
@@ -173,6 +173,19 @@ more information on how to create Certificate resources.
 # Issuer specific annotations
 
 ## Custom Fields
-Starting `v0.14` you can pass custom fields to Venafi (TPP version `v19.3` and higher) using the `venafi.cert-manager.io/custom-fields` annotation on Certificate resources.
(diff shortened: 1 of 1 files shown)


content/en/docs/configuration/venafi.md
```diff
@@ -169,3 +169,10 @@ You are now ready to issue certificates using the newly provisioned Venafi
 
 Read the [Issuing Certificates](../../usage/certificate/) document for
 more information on how to create Certificate resources.
+
+# Issuer specific annotations
+
+## Custom Fields
+Starting `v0.14` you can pass custom fields to Venafi (TPP version `v19.3` and higher) using the `venafi.cert-manager.io/custom-fields` annotation on Certificate resources.
```
munnerz: Can be changed to `19.2` now 🎉

content/en/docs/configuration/venafi.md
```diff
@@ -169,3 +169,10 @@ You are now ready to issue certificates using the newly provisioned Venafi
...
+The value is a JSON encoded array of custom field objects having a `name` and `value` key.
+For example: `venafi.cert-manager.io/custom-fields: '[{"name": "field-name", "value": "vield value"}]'`
```
munnerz: I think it may be useful to include a 'snippet' that includes the rest of the object too, e.g:

```
kind: Certificate
metadata:
  name: example-com-certificate
  annotations:
    venafi.cert-manager.io/custom-fields: '[{"name": "field-name", "value": "vield value"}]'
...
```

It _may_ also be worth including more than one custom field in this example too? Although it may be a bit confusing all on one line.. I think this is also a valid way to write it though:

```
kind: Certificate
metadata:
  name: example-com-certificate
  annotations:
    venafi.cert-manager.io/custom-fields: |-
      [
        {"name": "field-name", "value": "vield value"},
        {"name": "field-name-2", "value": "vield value"}
      ]
...
```

This has also just raised/made me think.. do we handle the case where multiple custom field entries specify the same `name`? What is the behaviour in that case? 🤔

munnerz:
Ping on the above @meyskens 😄

munnerz:
Updated to target `master`

/test all

content/en/docs/configuration/venafi.md
meyskens: We do pass it in this way to Venafi, so their platform defines the behavior on multiple fields with the same name

meyskens:
Implemented feedback, back to you @munnerz

PR #117, opened 2020-02-18
Set version to `0.13.0`

masih:
The version in documents is not released yet and results in
404 when executed.


munnerz:
v0.13.1 is in the process of being released - the PR (#116) that updated the version should have been merged *after* the release, but given we are talking about 30 minutes here I don't think it's worth flipping this back 😄

masih:
That makes sense thanks @munnerz

munnerz:
https://github.com/jetstack/cert-manager/releases/tag/v0.13.1 😄 it'll take about 30 minutes to roll out to the Helm Hub - thanks for opening this PR all the same, you have a keen/quick eye!

PR #119, opened 2020-02-19
Fix `helm install` command output

saalaa:
I believe the output of the `helm install` command had been mistaken actual documentation.


munnerz:
🤦‍♂  you are absolutely correct.. thanks for doing this!

/lgtm
/approve
/ok-to-test
/override dco

PR #126, opened 2020-02-20
Fix spelling error

mattshin:
I assumed the intended word was "lose" over "loosen", please let me know if we should change to that instead.


munnerz:
You are right! Thanks 😄 

/lgtm
/approve
/ok-to-test
/override dco

mattshin:
@munnerz whoops didn't see that you already approved - i amended the commit with a sign-off, do you mind approving again?

PR #128, opened 2020-02-24
Add description of Ingress annotations for out-of-tree Issuers

maxweis:
These Ingress annotations are necessary for ingress-shim to create Certificates with out-of-tree Issuers. As such, documentation for this is helpful.


munnerz:
Thanks for this 😄 

/lgtm
/approve
/ok-to-test

PR #129, opened 2020-02-26
Fix typo in route53.md

ianstanton:
- I believe `bpolicy` should read `policy` here


ianstanton:
Tried to add `DCO signoff`. Looks like it isn't being recognized.

munnerz:
It's because you added merge commits into your branch, which themselves don't have DCO signoff messages. Merge commits cause problems round here 😅 

/lgtm
/approve

PR #132, opened 2020-02-26
Use DocSearch

meyskens:
This is a PR for using Algolia's DocSearch instead of InstantSearch.

The configuration is at https://github.com/algolia/docsearch-configs/blob/master/configs/cert-manager.json

This PR is now ready for full review

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -1,5 +1,5 @@
 var versionName = "docs"
-let versionMatch = window.location.href.match(/v(.*)-docs\//) 
+let versionMatch = window.location.pathname.match(/v(.*)-docs\//) 
 if (versionMatch && versionMatch[1]) { // check if we matched a version
   versionName = `v${versionMatch[1]}-docs`
 }
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -1,34 +1,15 @@
-var client = algoliasearch('18N9PEKHUC', '29ab24453d2b70065a76b4e1a6a82abc');
-var indexName = "cert-manager-latest"
+var versionName = "docs"
 let versionMatch = window.location.href.match(/v(.*)-docs\//) 
 if (versionMatch && versionMatch[1]) { // check if we matched a version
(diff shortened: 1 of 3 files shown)


meyskens:
Version based indexing added in https://github.com/algolia/docsearch-configs/pull/1744
TODO: someway to handle automated version adding

munnerz:
This does look a lot simpler 😄 

That said, it appears to have rendered incorrectly for me on the deploy preview: 
Screenshot 2020-03-03 at 08 52 28

Any ideas? I did try refreshing to see if it was a stale cache but it didn't seem to help!

munnerz:
Also seems like there may be a few dupes in there too 🙈

meyskens:
@munnerz test it on https://deploy-preview-132--cert-manager.netlify.com/docs/ i messed up putting the CSS in the correct place, there might be dupes because I pushed versioning but haven't updated the JS yet

munnerz:
Nice! Yes, way better 😄 

Screenshot 2020-03-03 at 09 00 55

meyskens:
Index for 0.13 added in https://github.com/algolia/docsearch-configs/pull/1788

munnerz:
Do we want to index `next-docs` too? We don't actually currently index this today as it doesn't match the regex in the `scripts/index` file. Could be done later, and from what I understand only requires a change over in the algolia repo and not here

munnerz:
Also, when searching here: https://deploy-preview-132--cert-manager.netlify.com/v0.12-docs/ I expected to see some results (as only v0.13 is being added here, v0.12 already existed: https://github.com/algolia/docsearch-configs/pull/1788/files)

Is that expected?

meyskens:
@munnerz `facetFilters: ["version:view-132--cert-manager.netlify.com/v0.12-docs"]"}]}`
seems to have made a bug with the deploy preview, taking a look

meyskens:
Fixed this by using the path specified by the browser.

PR #158, opened 2020-03-15
Remove CRD installation step for Helm chart

longngn:
Because CRDs are already included in the chart: https://github.com/jetstack/cert-manager/tree/master/deploy/charts/cert-manager/crds

### DIFF
--- a/content/en/docs/installation/kubernetes.md
+++ b/content/en/docs/installation/kubernetes.md
@@ -111,18 +111,12 @@ docs](https://github.com/helm/helm/blob/240e539cec44e2b746b3541529d41f4ba01e77df
 
 In order to install the Helm chart, you must follow these steps.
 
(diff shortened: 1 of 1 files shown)


meyskens:
While the CRDs are in the charts folder they are not part of the actual charts release and still need to be installed.

longngn:
> While the CRDs are in the charts folder they are not part of the actual charts release and still need to be installed.

Why are they not part of the actual release? Is there any plan to include CRDs in the chart in the future?

munnerz:
We're waiting on a response of this issue before we can confidently include CRDs into the Helm chart again: https://github.com/helm/helm/issues/7735

If you've got any thoughts/opinions/insight, please add a comment to that issue! It'd be greatly appreciated 🙌 

I'm going to close this PR for now, as this _is_ a required step :)

PR #164, opened 2020-03-20
Includes the CRL Distribution Points option

srbraun:
EDIT:

Documentation for https://github.com/jetstack/cert-manager/pull/2625

### DIFF
--- a/content/en/docs/configuration/ca.md
+++ b/content/en/docs/configuration/ca.md
@@ -55,7 +55,7 @@ spec:
   ca:
     secretName: ca-key-pair
 ```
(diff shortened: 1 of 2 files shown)


content/en/docs/configuration/ca.md
```diff
@@ -55,7 +55,15 @@ spec:
   ca:
     secretName: ca-key-pair
 ```
-
+Optionally, you can specify (CRL)[https://en.wikipedia.org/wiki/Certificate_revocation_list] Distribution Points. An array of `Strings` each of which identifies the location of the CRL from which the revocation of this certificate can be checked:
```
meyskens: Any particular reason we backtick Strings? I think just `strings` (with not backticks) will do in this case

srbraun: done

munnerz:
Thanks for putting this feature together! 😄 

/lgtm
/approve

srbraun:
> Thanks for putting this feature together! smile

Thank you @munnerz  for the support.

PR #169, opened 2020-03-25
Fix typo in the docs

roy-work:
There's no such thing as a "USI" SAN, it's a URI.

Also, drop the "SAN" here; we don't repeat it for the other classes of "SAN" (e.g., "DNS Name SAN, IP Address SAN", and we're not explicitly naming the key (i.e., `uriSAN`).

### DIFF
--- a/content/en/docs/usage/certificate.md
+++ b/content/en/docs/usage/certificate.md
@@ -50,7 +50,7 @@ spec:
   usages:
     - server auth
     - client auth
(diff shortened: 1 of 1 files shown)


roy-work:
Sorry about the slow follow up w.r.t. DCO: I wrote this in the context of my job, so I had to follow up internally about it. My employer is fine w/ this change being upstreamed. It's too late to alter the commit, of course, but I wanted to note that here.

Signed-off-by: Roy Wellington Ⅳ <roy@commure.com>

PR #183, opened 2020-04-22
Explain usage of cnameStrategy: Follow in greater detail

unbreakab1e:
Hello! 

Recently we started using cert-manager for our development clusters and I had a hard time wrapping my head around getting `cnameStrategy: Follow` to work. 
Surprisingly, the most comprehensible doc for this case was found in the acme.sh project, but it is not easy to find if you are using an Average Joe google-fu.
So please consider merging this pull request which explains how to use `cnameStrategy: Follow` in greater detail.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -113,6 +113,7 @@ Smallstep
 SOA
 subdomain
 subdomains
+(sub)domains
 subresource
 templating
 Tiller
--- a/content/en/docs/configuration/acme/dns01/_index.md
+++ b/content/en/docs/configuration/acme/dns01/_index.md
@@ -75,11 +75,22 @@ By default, cert-manager will not follow CNAME records pointing to subdomains.
 
 If granting cert-manager access to the root DNS zone is not desired, then the
 `_acme-challenge.example.com` subdomain can instead be delegated to some other,
-less privileged domain.
-Once a CNAME record has been configured to point at the desired domain, and the
-DNS configuration/credentials for the zone that *should be updated* have been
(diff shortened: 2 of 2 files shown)


meyskens:
Thanks for this, it looks good however the zone IDs like `XXXXXXXXXXXXXXXXXXXX` become confusing. Not too familiar with route53 but it in other providers like CloudDNS a zone name called `less-privileges-domain-com` are possible, could we use those so the YAML is more easy to read.
Thanks!

unbreakab1e:
Well, in Route53 hostedZoneID is an autogenerated upper-case alphanumeric string. 
In turn, I am not familiar with CloudDNS and thus, I am afraid I would not be able to make a plausible mock-up of the yaml.

meyskens:
Looked into this and it seems that Route53 is a rare one where we allow the Zone ID to be set. 
How about we do it like 
```yaml
         accessKeyID: <Access ID for domain less-privileges-domain.com here>
          hostedZoneID: <Zone ID for domain less-privileges-domain.com here>
```

unbreakab1e:
Looks perfectly reasonable, I've updated the PR, please have a look

meyskens:
Thanks!

/lgtm
/ok-to-test
/approve

meyskens:
@unbreakab1e could you add `(sub)domains` to .spelling? Thanks!

meyskens:
Sorry seems to be `multitide` needs to be `multitude`

PR #185, opened 2020-04-22
docs(route53): Adding EKS IRSA method

sc250024:
# Description

Someone emailed me personally asking about my comment on this previous issue: https://github.com/jetstack/cert-manager/issues/2147#issuecomment-590366437. I decided if that was the case, the documentation could use an update about how the EKS IRSA method works.

This is the PR that @munnerz asked for here: https://github.com/jetstack/cert-manager/issues/2147#issuecomment-590779944

# Features / Fixes

- Adds explicit documentation on the EKS IRSA method.
- Clarifies some of the existing AWS role information docs.

### DIFF
--- a/content/en/docs/configuration/acme/dns01/route53.md
+++ b/content/en/docs/configuration/acme/dns01/route53.md
@@ -13,7 +13,7 @@ how cert-manager handles DNS01 challenges.
 > Note: This guide assumes that your cluster is hosted on Amazon Web Services
 > (AWS) and that you already have a hosted zone in Route53.
 
-## Set up a IAM Role
+## Set up an IAM Role
 
 cert-manager needs to be able to add records to Route53 in order to solve the
 DNS01 challenge. To enable this, create a IAM policy with the following
@@ -21,27 +21,27 @@ permissions:
 
 ```json
 {
-    "Version": "2012-10-17",
-    "Statement": [
-        {
-            "Effect": "Allow",
-            "Action": "route53:GetChange",
-            "Resource": "arn:aws:route53:::change/*"
-        },
-        {
-            "Effect": "Allow",
-            "Action": [
-              "route53:ChangeResourceRecordSets",
-              "route53:ListResourceRecordSets"
-            ],
-            "Resource": "arn:aws:route53:::hostedzone/*"
-        },
-        {
-            "Effect": "Allow",
-            "Action": "route53:ListHostedZonesByName",
-            "Resource": "*"
-        }
-    ]
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": "route53:GetChange",
+      "Resource": "arn:aws:route53:::change/*"
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "route53:ChangeResourceRecordSets",
+        "route53:ListResourceRecordSets"
+      ],
+      "Resource": "arn:aws:route53:::hostedzone/*"
+    },
+    {
+      "Effect": "Allow",
+      "Action": "route53:ListHostedZonesByName",
+      "Resource": "*"
+    }
+  ]
 }
 ```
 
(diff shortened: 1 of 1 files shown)


TBBle:
Is it worth mentioning that users of eksctl can declare the Service Account with appropriate permissions using eksctl, and then pass that pre-existing user to cert-manager's Helm chart?

It's a lot simpler than managing the roles and SA annotations by hand, I've found.

TBBle:
Hmm. This doesn't mention the `fsGroup` change from https://github.com/jetstack/cert-manager/issues/2147; unless I messed up and it's not necessary?

sc250024:
> Hmm. This doesn't mention the `fsGroup` change from [jetstack/cert-manager#2147](https://github.com/jetstack/cert-manager/issues/2147); unless I messed up and it's not necessary?

Not sure that's related to this? This involves dealing with the AWS IAM portions of `cert-manager`, whereas that issue is how the Helm chart / Kubernetes manifest is written to properly account for the permissions on the `ServiceAccount` token.

sc250024:
> Is it worth mentioning that users of eksctl can declare the Service Account with appropriate permissions using eksctl, and then pass that pre-existing user to cert-manager's Helm chart? (If not using Helm, eksctl can also modify an existing service account with the appropriate annotation)
> 
> It's a lot simpler than managing the roles and SA annotations by hand, I've found.

I'm not against `eksctl`, but my trust of CLI tools to write properly fine-grained IAM roles doesn't extend very far. Interally we use Terraform, so IAM roles / policies need to be properly defined in code, not "Run the `eksctl` command."

TBBle:
In the eksctl use-case, _you_ provide the policy, i.e. the relevant snippet from my setup:
```yaml
    # This account is to support cert-manager making updates to the hosted zone
    # in Route53. This allows using an ACME system to auto-issue TLS certificates to https services.
    # Policy definition per cert-manager Route53 documentation, and also by examining the
    # "CertManager" instance policy in the eksctl source
    # https://cert-manager.io/docs/configuration/acme/dns01/route53/
    # https://github.com/weaveworks/eksctl/blob/32027355678915365c68053e9293900b951cbc04/pkg/cfn/builder/iam_helper.go#L55
    - metadata:
        name: cert-manager
        namespace: cert-manager
      attachPolicy:
        Version: "2012-10-17"
        Statement:
          - Effect: Allow
            Action:
              - route53:ChangeResourceRecordSets
              - route53:ListResourceRecordSets
            Resource: "arn:aws:route53:::hostedzone/<MYHOSTEDZONE>"
          - Effect: Allow
            Action:
              # route53:GetChange is here, see https://github.com/jetstack/cert-manager/issues/2827
              - route53:GetChange
              - route53:ListHostedZonesByName
            Resource: "*"

```

That policy is copied out of the eksctl source, because it's not "generating" the policy, it has the correct policy defined in its source, and is just creating the policy and role object for you, and the service account, and tying them together for you to save having to do a bunch of manual work in two systems.

It's also possible to provide it the ARN of a Policy you've written yourself, if your pipeline supports it, and it'll just create the Role and Service Account for you.

In the future, I hope eksctl will support "policy addons" (basically naming that policy rather than writing it out explicitly) for IRSA, as it does currently for setting up instance roles. Still perfectly fine if you want to not trust it, the option to write your own remains, but it's much clearer when our pipeline _is_ as simple as `run eksctl with this config` to be able to say "turn on the cert-manager policy" compared to "here is a policy that matches the cert-manager needs".

The other advantage of eksctl is that AWS's own docs point you to it, so it's already on the table if you're using EKS, and not using some other infrastructure tool like Terraform already.

You can see all the eksctl addon policies in [iam_helper.go](https://github.com/weaveworks/eksctl/blob/32027355678915365c68053e9293900b951cbc04/pkg/cfn/builder/iam_helper.go). They're relatively fine-grained and hand-written, although they can't be customised for things like the Resource sadly.

TBBle:
For `fsGroup`, I'm not sure where _else_ to mention it in the docs, since it's specific to IRSA, and for whatever reason was not made the default for cert-manager -- which makes sense, you probably don't want that setting in other contexts, or it may need to be a _different_ fsGroup in other contexts.

TBBle:
While you're touching things, per https://github.com/jetstack/cert-manager/issues/2827, how do you feel about changing the `Resource` for `GetChanges` to `"*"`? I was going to wait for this PR to land and then put that change up for PR too, and am happy to do that if you're not interested in that now.

sc250024:
> While you're touching things, per [jetstack/cert-manager#2827](https://github.com/jetstack/cert-manager/issues/2827), how do you feel about changing the `Resource` for `GetChanges` to `"*"`? I was going to wait for this PR to land and then put that change up for PR too, and am happy to do that if you're not interested in that now.

I feel that it's scope creeping a bit. Since @munnerz asked for the PR, and @meyskens is actually doing the reviewing, I'll let them make a judgement call on that.

JoshVanL:
Awesome, thanks for that addition @sc250024 

/lgtm
/approve

PR #207, opened 2020-05-06
Update install guide to mention installCRDs option

munnerz:
Signed-off-by: James Munnelly <james@munnelly.eu>

/assign @meyskens 
/milestone v0.15

### DIFF
--- a/content/en/docs/installation/kubernetes.md
+++ b/content/en/docs/installation/kubernetes.md
@@ -42,7 +42,7 @@ $ kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/rel
 ```
 
 > **Note**: If you're using a Kubernetes version below `v1.15` you will need to install the legacy version of the manifests.
-> This version does not have API version conversion and only supports `cert-manager.io/v1alpha2` API resources. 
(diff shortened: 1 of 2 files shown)


content/en/docs/installation/kubernetes.md
```diff
@@ -131,26 +118,59 @@ Add the Jetstack Helm repository.
...
 
 # Helm v2
 $ helm install \
   --name cert-manager \
   --namespace cert-manager \
   --version v0.15.0 \
-  jetstack/cert-manager
-
+  jetstack/cert-manager \
+  # --set installCRDs=true
```
meyskens, line 173: Did we test this with Helm 2? I do remember we specifically targeted Helm 3 with this.

munnerz: We're pretty much falling back on Helm's default behaviour by including the CRDs as we do - we don't use the `crd-install` annotation, and the way that Helm handles CRDs in the `templates/` directory has not changed between 2 and 3 :).

meyskens: okay!

PR #221, opened 2020-05-13
Fix GKE RBAC note in installation instructions.

mltsy:
Somehow this note got split into two parts with another note inserted in between.

### DIFF
--- a/content/en/docs/installation/kubernetes.md
+++ b/content/en/docs/installation/kubernetes.md
@@ -59,11 +59,11 @@ $ kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/rel
 > 'elevate' your own privileges to that of a 'cluster-admin' **before** running
 > the above command. If you have already run the above command, you should run
 > them again after elevating your permissions:
->  ```bash
->  kubectl create clusterrolebinding cluster-admin-binding \
->    --clusterrole=cluster-admin \
->    --user=$(gcloud config get-value core/account)
->  ```
+```bash
+  kubectl create clusterrolebinding cluster-admin-binding \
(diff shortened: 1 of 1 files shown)


JoshVanL:
Thanks for that! @mltsy  :sweat_smile: 

It looks like the formatting is a bit off though. You can test the website locally with `./scripts/server` at the repo root, or by checking the `deploy/netlify` details in the CI information below. 

I think it should be the case of added an extra space before the ```bash

/unassign
/assign @mltsy

mltsy:
Oh!  Interesting - it looked fine in both the VSCode preview and the Github preview.
Thanks for the tips - I tried running `./scripts/server` and got an error:
```
Building sites … ERROR 2020/05/14 09:32:04 Transformation failed: SCSS processing failed: file "stdin", line 6, col 1: File to import not found or unreadable: ../vendor/bootstrap/scss/bootstrap.
```

But I tried with the extra space you suggested and will see how it comes out in `deploy/netlify`... 🤞

mltsy:
Well that first one didn't work... I'm not sure why - it seems like the markdown processor being used (wherever that is) is not following standard or github markdown specs... must be another flavor I'm not familiar with...?

mltsy:
Huh, that didn't work either - well I'm a little stumped!  I guess I'll just remove the code-block from the blockquote, which makes it read a little weird, but at least doesn't mess up the whole document 😅  If you happen to know what kind of markdown we're supposed to be using here, maybe we can fix it!

mltsy:
Wow!  Somehow giving up and removing the blockquote actually made it to what I was trying to do in the first place with standard markdown, and put the code block inside the blockquote! 😆

PR #224, opened 2020-05-18
Fixing documentation issue discovered during nginx walkthrough

johndietz:
During the nginx walkthrough I discovered that there's an incorrect command
`kubectl describe certificate`
which should be changed to
`kubectl describe certificate quickstart-example-tls`
... oddly, down in the results of that command, it looks like part of the correct command got incorrectly pasted in the documented results, so cleaning that up as well.


johndietz:
not sure why it's saying i'm missing dco signoff, my last two commits both include 
```
signed-off-by: John Dietz <jdietz@virtru.com>
```

meyskens:
@johndietz the 2nd one has a lower case `S`

PR #226, opened 2020-05-20
Add clarification in EKS OIDC definition

atmosx:
Clarify that we need to remove the IAM role from `issuer` definition, to avoid `cert-manager` hitting an `sts:AssumeRole` permissions error.


content/en/docs/configuration/acme/dns01/route53.md
```diff
@@ -225,6 +225,33 @@ metadata:
     eks.amazonaws.com/role-arn: arn:aws:iam::XXXXXXXXXXX:role/cert-manager
 ```
 
+Make sure there is no `role: arn...` field on the issuer definition, otherwise you might end up with `sts:assumerole` permission error. The issuer or clusterissuer resource definition in case of EKS OIDC should look a like this:
```
meyskens, line 228:
```suggestion
Make sure there is no `eks.amazonaws.com/role-arn` annotation on the issuer definition, otherwise you might end up with `sts:assumerole` permission error. The issuer or clusterissuer resource definition in case of EKS OIDC should look a like this:
```

meyskens:
Thanks for this! just one suggestion to use the specific terminology so users are less likely to be confused :)

content/en/docs/configuration/acme/dns01/route53.md
atmosx: Thank you for the suggestion @meyskens !

I modified wording and fixed a few more typos. Hopefully looks better now, ty!

content/en/docs/configuration/acme/dns01/route53.md
```diff
@@ -225,6 +225,35 @@ metadata:
...
+          - "foobar.cloud"
+          - "*.foobar.cloud"
+      dns01:
+        route53:
+          region: us-east-1
+          hostedZoneID: ABCD123
+
+```
+
+To check for errors, create a `certificate` resource to kickstart the process and monitor  `cert-manager` pod logs.
```
meyskens, line 255: It is not advice to use the cert-manager logs for investigating issues, cert-manager reports all events back to Kubernetes which you can see using `kubectl describe`

atmosx: > It is not advice to use the cert-manager logs for investigating issues, cert-manager reports all events back to Kubernetes which you can see using `kubectl describe`

Hi! I see, I did not recall checking the output of `describe` tbh. Unfortunately I won't be able to test the scenario anytime soon. What about closing the PR and re-openning it when we can test the error log appears when describing the resource? I might be able to do so in a few weeks.

meyskens: Is good! feel free to close it!

PR #233, opened 2020-06-06
Document secret ownership behaviour

haines:
I found the default behaviour surprising - I expected that when I deleted a certificate, the secret would also be removed.

It seems like I'm not the only one who has run into this (e.g. https://github.com/jetstack/cert-manager/issues/1944), so assuming changing the default behaviour is not an option, it'd be good to document it!


meyskens:
Sorry for the delay! This one slipped my GitHub inbox...
Looks great! Thanks for this!

/lgtm
/approve

haines:
No worries @meyskens, thanks for reviewing!

PR #254, opened 2020-07-18
Add DNS01RecursiveNameserversOnly to docs

rtgibbons:
I found this to be true during my install. After reading the source I found reference to the option for --dns01-recursive-nameservers-only - This was required to have cert-manager use the provide recursive nameservers and not the define NS record on the domain i was trying to have ACME generate certificates on.

### DIFF
--- a/content/en/docs/configuration/acme/dns01/_index.md
+++ b/content/en/docs/configuration/acme/dns01/_index.md
@@ -51,14 +51,15 @@ read the multiple-solver-types section.
 ## Setting Nameservers for DNS01 Self Check
 
 cert-manager will check the correct DNS records exist before attempting a DNS01
-challenge.  By default, the DNS servers for this check will be taken from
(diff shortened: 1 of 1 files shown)


meyskens:
Thanks for this! This sure would be good to require to have in the documentation, however it should be a separate case listed and is not required for every user that needs alternate DNS01 revolvers.
Could you add this is a separate case in the document? That would be great!

rtgibbons:
No problem.

@meyskens I've updated the documentation talk about how DNS01 validation checks work based on how I understand it from the code, and provided clarification around each options. 

I believe these two options should reside in the same section based on how the interact with DNS01 self checks with regards to nameservers. And I left the examples with both options to show that `--dns01-recursive-namerservers-only` is just a flag and then how to supply multiple args via helm.

Let me know what you think.

meyskens:
Looks good! Thanks a lot!

/lgtm
/approve
/ok-to-test

PR #255, opened 2020-07-19
Add reference to .data.ca.crt in secret property

kylos101:
Hi, adding a `ca.crt` property to the secret appears to be a common feature of certs provisioned by cert-manager. 

I am updating the docs so that folks looking to take advantage of this secret can see that it contains the related ca.crt, too. 

Let me know if this is correct, or if I should change anything?


kylos101:
Fighting with GIT...will make a new PR with a signed commit.

PR #256, opened 2020-07-19
Add reference to .data.ca.crt in secret property

kylos101:
Signed-off-by: Kyle Brennan <kylos101@gmail.com>

Hi there, just adding a reference to the docs so that it mentions that the Certificate's Secret will also include a reference to the related CA certificate via a ca.crt property. 

Let me know if this is okay, or if you'd like any changes? Thank you!

### DIFF
--- a/content/en/docs/concepts/certificate.md
+++ b/content/en/docs/concepts/certificate.md
@@ -37,10 +37,10 @@ spec:
 
 This `Certificate` will tell cert-manager to attempt to use the `Issuer` named
 `letsencrypt-prod` to obtain a certificate key pair for the `foo.example.com`
-and `bar.example.com` domains. If successful, the resulting key and certificate
-will be stored in a secret named `acme-crt-secret` with keys of `tls.key` and
-`tls.crt` respectively. This secret will live in the same namespace as the
-`Certificate` resource.
+and `bar.example.com` domains. If successful, the corresponding CA certificate,
+and resulting TLS key and certificate will be stored in a secret named 
+`acme-crt-secret`, with keys of `ca.crt`, `tls.key`, and `tls.crt` respectively.
+This secret will live in the same namespace as the `Certificate` resource. 
 
 The `dnsNames` field specifies a list of [`Subject Alternative
 Names`](https://en.wikipedia.org/wiki/Subject_Alternative_Name) to be associated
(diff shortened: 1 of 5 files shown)


content/en/docs/concepts/certificate.md
```diff
@@ -37,10 +37,10 @@ spec:
 
 This `Certificate` will tell cert-manager to attempt to use the `Issuer` named
 `letsencrypt-prod` to obtain a certificate key pair for the `foo.example.com`
-and `bar.example.com` domains. If successful, the resulting key and certificate
-will be stored in a secret named `acme-crt-secret` with keys of `tls.key` and
-`tls.crt` respectively. This secret will live in the same namespace as the
-`Certificate` resource.
+and `bar.example.com` domains. If successful, the corresponding CA certificate,
```
meyskens, line 40: The reason that this probably never made it into the documentation is that not all Issuers support the adding of the CA certificate, for example most importantly this will not work with ACME.
We should document this here for sure but with a note that cert-manager will only do this if the CA is known

kylos101: Hi @meyskens, I had no idea, thank you so much for clarifying! When using ACME, would the `ca.crt` key exist, and the value be empty? Or would the key not exist at all?

meyskens: In the current release it will not exist, it used to be empty but that changed.

kylos101: Gotcha, so, if version >= 0.16, it will not exist when the CA is unknown. If version <= 0.15, it could exist with or without a value (depending on whether the CA is known)?

meyskens: Sorry i replied but forgot to sent. Yup that's correct this was added in https://github.com/jetstack/cert-manager/pull/2947. Since the documentation is versioned can you describe the 0.16 behaviour in here?

kylos101: No prob, @meyskens, I really appreciate your help! I just added 8746d15, let me know if you need anything else?

meyskens:
/override dco
/approve
/ok-to-test
/lgtm

Thank you very much!

meyskens:
the dco override is having a bad day :(

PR #258, opened 2020-07-21
Added more troubleshooting info

jslay88:
In regards to https://github.com/jetstack/cert-manager/issues/2980

### DIFF
--- a/content/en/docs/faq/acme.md
+++ b/content/en/docs/faq/acme.md
@@ -131,8 +131,8 @@ You will also see any errors coming from your DNS provider here.
 ### HTTP01 troubleshooting
 First of all check if you can see the challenge URL from the public internet, if this does not work check your Ingress and firewall configuration as well as the service and pod cert-manager created to solve the ACME challenge.
 If this does work check if your cluster can see it too. It is important to test this from inside a Pod. If you get a connection error it is suggested to check the cluster's network configuration.
-If you receive a `tls: handshake failure`, try setting the annotation `cert-manager.io/issue-temporary-certificate: "true"` on your application's ingress.
-If you still are having issues, there may be a conflict with having multiple ingress using the same hostname, in this case, the annotation `acme.cert-manager.io/http01-edit-in-place: "true"` is likely required.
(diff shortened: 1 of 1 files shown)


content/en/docs/faq/acme.md
```diff
@@ -131,6 +131,8 @@ You will also see any errors coming from your DNS provider here.
 ### HTTP01 troubleshooting
 First of all check if you can see the challenge URL from the public internet, if this does not work check your Ingress and firewall configuration as well as the service and pod cert-manager created to solve the ACME challenge.
 If this does work check if your cluster can see it too. It is important to test this from inside a Pod. If you get a connection error it is suggested to check the cluster's network configuration.
+If you receive a `tls: handshake failure`, try setting the annotation `cert-manager.io/issue-temporary-certificate: "true"` on your application's ingress.
```
meyskens, line 134:
```suggestion
If you receive a `tls: handshake failure`, try setting the annotation `cert-manager.io/issue-temporary-certificate: "true"` on the Ingress or Certificate resource. This will issue a temporary self signed certificate for the ingress controller to use before the actual certificate is issued.
```

content/en/docs/faq/acme.md
```diff
@@ -131,6 +131,8 @@ You will also see any errors coming from your DNS provider here.
...
+If you still are having issues, there may be a conflict with having multiple ingress using the same hostname, in this case, the annotation `acme.cert-manager.io/http01-edit-in-place: "true"` is likely required.
```
meyskens, line 135: Multiple ingresses for the same hostname is a valid thing to do, I think we should sightly change the working here. 
```suggestion
If you still are having issues, there may be an issue with your ingress controller handling multiple resources for the same hostname, in this case, the annotation `acme.cert-manager.io/http01-edit-in-place: "true"` is likely required.
```

meyskens:
Thanks for this! Changed the wording a bit to be more general and not ingress-shim specific. Let me know what you think then happy to merge :) 

/override dco
/ok-to-test

content/en/docs/faq/acme.md
jslay88: All good to me.

content/en/docs/faq/acme.md
jslay88: All good to me.

PR #266, opened 2020-07-27
Fix link to kubectl on Darwin

Nowaker:


meyskens:
Thanks! that might work better :) 

/override dco
/approve
/lgtm
/ok-to-test

PR #267, opened 2020-07-27
Add notice on sub-charting cert-manager

meyskens:
A short but powerful notice not to use cert-manager as a sub-chart with Helm


meyskens:
@james-w can you retry

PR #280, opened 2020-08-06
Update digitalocean.md

haven2dotnet:
Added a one-liner to convert your access token (which isn't base64 encoded) into base64 format.


meyskens:
Thanks

/lgtm
/ok-to-test
/approve
/override dco

Issue #237, opened 2020-06-27
docs for ACMEChallengeSolverHTTP01Ingress doesn't specify what `class` values are available


jmhodges:
It'd be nice to know what variables are allowed for the `class` field on `ACMEChallengeSolverHTTP01Ingress`. There's a hint that there's some gce value allowed, but the only examples (on another page) I've seen involve `nginx`.

meyskens:
It allows all values, it will depend on what your ingress controller defines. See more at https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class

A PR to make that more clear is welcome!

PR #284, opened 2020-08-07
Clarify source of API docs

TheEnbyperor:
Signed-off-by: Q <q@as207960.net>

Arose whilst trying to fix issue #237


content/en/docs/reference/.where-are-the-docs
```diff
@@ -0,0 +1,4 @@
+# AAAAA?! Where be the docs?
+
+API Docs are autogenerated from the Golang source.
+Go edit the source comments [here](https://github.com/jetstack/cert-manager/tree/master/pkg/internal/apis) to update API docs.
```
meyskens, line 4: Can we make this the versioned API docs as we document those and not the internal one, which is only used for conversion logic.

maelvls: @TheEnbyperor Here is the link you can use:
```suggestion
Go edit the source comments [here](https://github.com/jetstack/cert-manager/tree/master/pkg/apis) to update API docs.
```

SgtCoDFish:
Wow, I'd not seen this before and I _really_ like the idea of adding this. I was caught out by this more than once when I started contributing to cert-manager!

Unfortunately we've got a new website refresh coming up and since this PR has been open for so long (sorry about that) I don't think it'll make sense in the refreshed version (and I'd actually like to always check in the generated docs just so there's _something_ in the repo).

As such I'm going to close this, but thanks again @TheEnbyperor  - really great idea and I love the thought behind it!

PR #285, opened 2020-08-12
State DigitalOcean access tokens need write access

remcohaszing:
If not, the DigitalOcean API will reply with 403 status codes.


meyskens:
Thanks for this!

/lgtm
/approve
/ok-to-test

PR #315, opened 2020-09-02
Use the latest upstream version of gen-crd-api-reference-docs

wallrj:
Installs v0.2.0, which:
 * Contains @munnerz patches
 * No longer has a released binary, so I install using `go get`
 * Requires an update to go 1.13+ so that go-get  can install the binary


.go-version
```diff
@@ -0,0 +1 @@
+1.15.1
```
wallrj, line 1: This is undocumented, but mentioned in https://github.com/netlify/build-image/issues/205

meyskens:
Nice! 

/approve
/lgtm

PR #322, opened 2020-09-04
Update for v1.0.1

meyskens:
Happy v1.0.1!


wallrj (approved):
/lgtm

We also need to update the openshift installation docs to use 1.0.1

PR #335, opened 2020-10-01
Fix invalid resource spec

Nikhilmadduri:
Signed-off-by: James Munnelly <james@munnelly.eu>


meyskens:
`cert-manager:munnerz-patch-2` should probably be deleted

/spam
/close

PR #340, opened 2020-10-02
Spellcheck and Correction

Amit-Salunke-02:


meyskens:
Thanks @JoshVanL was just about to do that

meyskens:
I wonder if /spam is the incorrect command... need to read more Prow docs...

PR #345, opened 2020-10-07
Use correct fields in HTTP01 challenge docs

linkvt:
The documentation still used probably some old field which are not part of the v1 resource definition anymore, see https://cert-manager.io/docs/reference/api-docs/#acme.cert-manager.io/v1.ACMEChallengeSolverHTTP01Ingress .

Signed-off-by: Vincent Link <mail@linkvt.de>


meyskens:
Thanks this was indeed overdue :D 

/approve
/lgtm
/ok-to-test

PR #346, opened 2020-10-08
Document how to configure Venafi TPP access-token authentication

wallrj:
Part of: https://github.com/jetstack/cert-manager/issues/3140

### DIFF
--- a/content/en/docs/configuration/venafi.md
+++ b/content/en/docs/configuration/venafi.md
@@ -109,7 +109,64 @@ the connection parameters are slightly different.
 
 In order to set up a Venafi Trust Protection Platform `Issuer`, you must first
 create a Kubernetes `Secret` resource containing your Venafi TPP API
-credentials:
+credentials.
+
+NOTE: For TPP >= 19.2 use Access Token Authentication
+and for older versions of TPP, use username / password authentication.
+
+#### Access Token Authentication
+
+Use access-token authentication if you are connecting to `TPP >= 19.2`.
+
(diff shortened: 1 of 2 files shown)


content/en/docs/configuration/venafi.md
```diff
@@ -109,7 +109,59 @@ the connection parameters are slightly different.
...
+   refresh_token:  Zd48............RJ1sAK6MTnebXlm7+Q==
+   ```
+
+5. Save the token to a Secret in the Kubernetes cluster
+
+```bash
+$ kubectl create secret generic \
+       tpp-secret \
+       --namespace=<NAMESPACE OF YOUR ISSUER RESOURCE> \
+       --from-literal=token='YOUR_TPP_ACCESS_TOKEN'
```
munnerz, line 159: Drive-by comment, but how is this token refreshed? Are users expected to manually update this token? If so, is the controller designed to gracefully detect an expired token and prevent further requests to TPP?

I haven't looked into the new auth mechanism's internal workings, but it seems like this sort of thing shouldn't be exposed to end users at all. For example in oauth, the access/refresh token is what is used to authenticate requests _after_ some other initial authentication is performed (eg username/password auth, service account based auth, or whatever else).

Either way, it'd be good to note here how a user should maintain this access token secret over time 🙂

wallrj: Thanks @munnerz . Yes, we're discussing it now. 

The same problem is described in the vault issuer docs: https://cert-manager.io/docs/configuration/vault/#authenticating-with-a-token

PR #148, opened 2020-03-10
feat(azure): add documentation of usage for Managed Identities

gitirabassi:
Signed-off-by: gitirabassi <giacomo@tirabassi.eu>

This is the documentation for https://github.com/jetstack/cert-manager/pull/2681

PR #350, opened 2020-10-20
Add Docs to support using Managed Identity

### LINKED ISSUE #328: Document use of Azure Managed Identities for AzureDNS
**Is your feature request related to a problem? Please describe.**

Support for Azure managed identities was added by this pull request and added to release [0.15.0](https://github.com/jetstack/cert-manager/releases/tag/v0.15.0). Unfortunately there is no documentation regarding use of MSI in the [official docs](https://cert-manager.io/docs/configuration/acme/dns01/azuredns/). I went looking around and I found that the original contributor @gitirabassi of the aforementioned MR wrote some [documentation](https://github.com/gitirabassi/website/blob/feat/azure-auth-msi-master/content/en/docs/configuration/acme/dns01/azuredns.md#managed-identity), but for whatever reason this was not merged back into master.

Moreover, the way the original contrib. explains the way this feature works contrasts the way pod-identity (which is instrumental for the feature to work) wants to define the identity bindings ( i.e. one says that `aadpodidbinding:<value>` should be ["cert-manager"](https://github.com/gitirabassi/website/blob/feat/azure-auth-msi-master/content/en/docs/configuration/acme/dns01/azuredns.md#managed-identity) vs. the other says it should be [$IDENTITY_NAME](https://github.com/Azure/aad-pod-identity#6-deployment-and-validation) aka. the MSI name) of course the maintainers of this repo cannot be responsible for the claim, since it is not in the official docs, still, I would ask if there is anyone that can dispel the confusion.

Is there someone who knows exactly how using managed identities works? Can we work together to document the feature?
In my case I would actually like to use multiple managed identities for different ClusterIssuers, or at the very least, use a different one rather than the omsagent MSI to modify my DNS records, do we know if this is possible to do at all?

**Describe the solution you'd like**
I see that the documentation [MR](https://github.com/cert-manager/website/pull/148) containing the aforementioned documentation for use of MSI has been opened in March and has recently received some love. The example though only refers to a terraform example, I would appreciate it helm install/static manifests are added too.

**Describe alternatives you've considered**
I've considered bypassing managed identities assignment using pod identity, by generating an MSI (using an ARM template) and then accessing the underlying service principal, and then giving the SP credentials to cert-manager via secret. Unfortunately it is not possible to access SP credentials after it has been created. The only time when those are returned is when an SP is created directly through az cli.

**Additional context**
Based from #3148 I can see that maybe there is also an issue if you have multiple MSI defined in the target resource group where AKS (and therefore pod identity is defined), can someone confirm this? The fact that you cannot define the name of the MSI you want to target seems to be a feature that could indeed be added (or is this an hint that `aadpodidbinding:???` should contain the `$IDENTITY_NAME`?)

**Environment details (if applicable)**:
- Kubernetes version (e.g. v1.10.2): 1.18.6
- Cloud-provider/provisioner (e.g. GKE, kops AWS, etc): Azure
- cert-manager version (e.g. v0.4.0): 1.0.0
- Install method (e.g. helm or static manifests): helm

/kind documentation
/triage support

JasonWhall:
Hopefully Resolves #328 

Picked up and addressed issues from PR #148 

- Adds Examples using Terraform/az cli to setup
- Example yaml to create identity and binding
- Labels to set to add identity binding to pod
- example of issuer using identity

### DIFF
--- a/content/en/docs/configuration/acme/dns01/azuredns.md
+++ b/content/en/docs/configuration/acme/dns01/azuredns.md
@@ -5,6 +5,130 @@ weight: 30
 type: "docs"
 ---
 
+To configure the AzureDNS DNS01 Challenege in a Kubernetes cluster there are 2 ways available:
+- Managed Identity ([Using AAD Pod Identities]((https://azure.github.io/aad-pod-identity)))
+- Service Principal
+
+## Managed Identity
+
+- Firstly we need to create the identity and grant this access to contribute to the DNS Zone.
+
+Example creation using `azure-cli` and `jq`:
+```bash
+# Choose a unique Identity name and existing resource group to create identity in.
+IDENTITY=$(az identity create --name $IDENTITY_NAME --resource-group $IDENTITY_GROUP )
+
+# Gets principalId to use for role assignment
+PRINCIPAL_ID=$(echo $IDENTITY | jq -r '.principalId')
+
+# Used for identity binding
+CLIENT_ID=$(echo $IDENTITY | jq -r '.clientId')
+RESOURCE_ID=$(echo $IDENTITY | jq -r '.id')
+
+# Get existing DNS Zone Id
+ZONE_ID=$(az network dns zone show --name $ZONE_NAME --resource-group $ZONE_GROUP --query "[].id" -o tsv)
+
+# Create role assignment
+az role assignment create --role "DNS Zone Contributor" --assignee $PRINCIPAL_ID --scope $ZONE_ID
+```
+
+Example creation using Terraform
+```terraform
+variable resource_group_name {}
+variable location {}
+variable dns_zone_id {}
+
+# Creates Identity
+resource "azurerm_user_assigned_identity" "dns_identity" {
+  name                = "cert-manager-dns01"
+  resource_group_name = var.resource_group_name
+  location            = var.location
+}
+
+# Creates Role Assignment
+resource "azurerm_role_assignment" "dns_contributor" {
+  scope                = var.dns_zone_id
+  role_definition_name = "DNS Zone Contributor"
+  principal_id         = azurerm_user_assigned_identity.dns_identity.principal_id
+}
+
+# Client Id Used for identity binding
+output "identity_client_id" {
+  value = azurerm_user_assigned_identity.dns_identity.client_id
+}
+
+# Resource Id Used for identity binding
+output "identity_resource_id" {
+  value = azurerm_user_assigned_identity.dns_identity.id
+}
+```
+
+- Next we need to ensure we have installed [AAD Pod Identity](https://azure.github.io/aad-pod-identity) using their walkthrough. This will install the CRDs and pods required to assign the identity.
+
+- Now we can create the identity resource and binding using the below yaml
+
+```yaml
+apiVersion: "aadpodidentity.k8s.io/v1"
+kind: AzureIdentity
+metadata:
+  annotations:
+    # recommended to use namespaced identites https://azure.github.io/aad-pod-identity/docs/configure/match_pods_in_namespace/
+    aadpodidentity.k8s.io/Behavior: namespaced 
+  name: certman-identity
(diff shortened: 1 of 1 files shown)


meyskens:
I'm not an azure expert so

/cc @riccardodl 
If you could also take a look it would be amazing!

content/en/docs/configuration/acme/dns01/azuredns.md
```diff
@@ -5,6 +5,130 @@ weight: 30
 type: "docs"
 ---
 
+To configure the AzureDNS DNS01 Challenege in a Kubernetes cluster there are 2 ways available:
+- Managed Identity ([Using AAD Pod Identities]((https://azure.github.io/aad-pod-identity)))
```
meyskens, line 9: This markdown feels a bit broken

calexandre:
@JasonWhall is it possible to adapt this to use the managed identity that the kubelet uses by default for an AKS cluster? I managed to make it work for the `external-dns` deployment, but I'm struggling to make it work for `cert-manager`...

It is kubelet identity is obtainable using this terraform output as an example:

```terraform
output "aks_default_managed_identity_client_id" {
  description = "The Client ID of the user-defined Managed Identity assigned to the Kubelets"
  value = azurerm_kubernetes_cluster.this.kubelet_identity[0].client_id
}
```

JasonWhall:
> @JasonWhall is it possible to adapt this to use the managed identity that the kubelet uses by default for an AKS cluster? I managed to make it work for the `external-dns` deployment, but I'm struggling to make it work for `cert-manager`...
> 
> The main advantage - if using an AKS cluster - is that you shouldn't need (in theory) the [AAD Pod Identity](https://azure.github.io/aad-pod-identity).
> 
> The kubelet identity is obtainable using this terraform output as an example:
> 
> ```terraform
> output "aks_default_managed_identity_client_id" {
>   description = "The Client ID of the user-defined Managed Identity assigned to the Kubelets"
>   value = azurerm_kubernetes_cluster.this.kubelet_identity[0].client_id
> }
> ```

Hey @calexandre, I was a bit torn on adding documentation on this option. While you certainly could use the Kubelet identity when your AKS cluster is deployed using Managed identity, there is some added risk here.

IIRC, by default all pods in the Kubernetes cluster will have access to call the azure metadata endpoint and assume the role of the Kubelet identity, so by extension if you assign this identity access to contribute to your DNS zone any pod also has this permission which could be dangerous.

There are means to control this however, for example using network policies to block access to the metadata IP. But this is then a Kubernetes operators choice to add this policy in or not.

That being said the actual implementation of using the Kubelet identity for cert-manager shouldn't be much different from the AAD pod identity implementation. All you will need to do is assign the Kubelet identity access to contribute to your DNS zone and then your issuer just needs to exclude the principalId and secret, similarly to the Pod identity issuer example.

I'm happy to add an additional commit for these steps if you think it's worthwhile having this explained as an option with the guidance that is the least secure?

content/en/docs/configuration/acme/dns01/azuredns.md
JasonWhall: Now unbroken!

calexandre:
> That being said the actual implementation of using the Kubelet identity for cert-manager shouldn't be much different from the AAD pod identity implementation. All you will need to do is assign the Kubelet identity access to contribute to your DNS zone and then your issuer just needs to exclude the principalId and secret, similarly to the Pod identity issuer example.

Unfortunately I haven't yet managed to make this work. I've came across some issues, but to avoid pollute this PR, we can continue this discussion on the referenced issue #328.

> I'm happy to add an additional commit for these steps if you think it's worthwhile having this explained as an option with the guidance that is the least secure?

IMHO I think it would worth to mention the possibility - supported by an example. I can provide you with mine IF we can make it work 😉

JasonWhall:
@meyskens is there anything else I can do to help get this merged in?

meyskens:
/approve
/lgtm

Thanks a lot! Looks good to me!

PR #352, opened 2020-10-21
Fix confusing part about ingress class defaults

meyskens:
Add clarification on used ingress class when not overwritten


jakexks (approved):
/approve
/lgtm

PR #353, opened 2020-10-21
Create externalloadbalancer.md

nd2014-public:
Return on Experience with ACME/Cert-manager behind a Load Balancer.

### DIFF
--- a/content/en/docs/configuration/externalloadbalancer.md
+++ b/content/en/docs/configuration/externalloadbalancer.md
@@ -33,6 +33,6 @@ When using HTTP(s) protocols for your Load Balancer, it can intercept the challe
 
 In this case, cert-manager will fail `did not get expected response when querying endpoint, expected 'xxxx' but got: yyyy (truncated)`.
 
-This kind of error can be thrown for multiple reasons. This case shows a correctly formated response, but not the expected one. In the case of OVH host, the HTTP requests replaced the reponse with OVH's Let's Encrpyt code and provoked this error.
+This kind of error can be thrown for multiple reasons. This case shows a correctly formated response, but not the expected one. In the case of OVH, the HTTP requests replaced the reponse with OVH's Let's Encrpyt code and provoked this error.
 
 The solution is to configure the Load Balancer with TCP protocol so that the HTTP request will not be intercepted by the host.
--- a/content/en/docs/configuration/externalloadbalancer.md
+++ b/content/en/docs/configuration/externalloadbalancer.md
@@ -18,7 +18,7 @@ Some Network Load Balancer have this kind of limitation for several reasons. It 
 To check if you are facing this problem :
 
 1. Check that the challenge endpoint is publically accessible : `curl <endpoint>`
-2. Check that the challenge endpoint is NOT accessible from inside behind the Load Balancer : Use SSH to open a session on a node places behind the LB ; then launch the same command than before : `curl <endpoint`
+2. Check that the challenge endpoint is NOT accessible from inside behind the Load Balancer: use SSH to open a session on a node places behind the LB; then launch the same command than before : `curl <endpoint>`
 
 The HTTP-01 challenge endpoint can be found in the logs when the `pre-check` fails. If it does not appear in the logs, you can check the challenge URL by `kubectl`command.
 
--- a/content/en/docs/configuration/externalloadbalancer.md
+++ b/content/en/docs/configuration/externalloadbalancer.md
@@ -5,7 +5,7 @@ weight: 40
 type: "docs"
 ---
 
-When you are using an external load balancer provided by any host, you can face several configuration issues to get it work with Cert Manager.
+When you are using an external load balancer provided by any host, you can face several configuration issues to get it work with cert-manager.
 
 This documentation is meant to help in configure HTTP-01 type of challenges for instances behind external load balancer.
 
--- a/content/en/docs/configuration/externalloadbalancer.md
+++ b/content/en/docs/configuration/externalloadbalancer.md
@@ -33,4 +33,4 @@ When using HTTP(s) protocols for your Load Balancer, it can intercept the challe
 
 In this case, cert-manager will fail `did not get expected response when querying endpoint, expected 'xxxx' but got: yyyy (truncated)`.
 
-This kind of error can be thrown for multiple reasons. This case shows a correctly formatted response, but not the expected one. The solution is to configure the Load Balancer with TCP protocol so that the HTTP request will not be intercepted by the host.
+This kind of error can be thrown for multiple reasons. This case shows a correctly formatted response, but not the expected one. The solution is to configure the Load Balancer with TCP protocol so that the HTTP request will not be intercepted by the host. 
--- a/content/en/docs/configuration/externalloadbalancer.md
+++ b/content/en/docs/configuration/externalloadbalancer.md
@@ -27,7 +27,7 @@ The HTTP-01 challenge endpoint can be found in the logs when the `pre-check` fai
 
 ## Load Balancer HTTP endpoints
 
-If you are using a Load Balancer, you should be able to configure the Load Balancer protocol as HTTP, HTTPS, TCP, UDP. Several Load Balancer now offer free SSL certificates with Let's Encrypt.
+If you are using a Load Balancer, you should be able to configure the Load Balancer protocol as HTTP, HTTPS, TCP, UDP. Several Load Balancer now offer free TLS certificates with Let's Encrypt.
 
 When using HTTP(s) protocols for your Load Balancer, it can intercept the challenge URL to replace the response's verification hash with their hash.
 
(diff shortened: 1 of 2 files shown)


content/en/docs/configuration/externalloadbalancer.md
```diff
@@ -0,0 +1,38 @@
+---
+title: "External Load Balancer"
+linkTitle: "External Load Balancer"
+weight: 40
+type: "docs"
+---
+
+When you are using an external load balancer provided by any host, you can face several configuration issues to get it work with Cert Manager.
```
meyskens, line 8:
```suggestion
When you are using an external load balancer provided by any host, you can face several configuration issues to get it work with cert-manager.
```

content/en/docs/configuration/externalloadbalancer.md
```diff
@@ -0,0 +1,38 @@
...
+
+This documentation is meant to help in configure HTTP-01 type of challenges for instances behind external load balancer.
+
+## NAT Loopback / Hairpin
+
+The first configuration point is NAT loopback. You can face `pre check` issues due to Load Balancer preventing instances behind him to access his external interface.
```
meyskens, line 14:
```suggestion
The first configuration point is NAT loopback. You can face pre-check issues due to Load Balancer preventing instances behind it to access its external interface.
```

content/en/docs/configuration/externalloadbalancer.md
```diff
@@ -0,0 +1,38 @@
...
+
+Some Network Load Balancer have this kind of limitation for several reasons. It can be configured through iptables rerouting configuration known as `NAT loopback`.
+
+To check if you are facing this problem :
+
+1. Check that the challenge endpoint is publically accessible : `curl <endpoint>`
+2. Check that the challenge endpoint is NOT accessible from inside behind the Load Balancer : Use SSH to open a session on a node places behind the LB ; then launch the same command than before : `curl <endpoint`
```
meyskens, line 21:
```suggestion
2. Check that the challenge endpoint is NOT accessible from inside behind the Load Balancer: use SSH to open a session on a node places behind the LB; then launch the same command than before : `curl <endpoint>`
```

meyskens: Can we also recommend using `kubectl exec`?

content/en/docs/configuration/externalloadbalancer.md
```diff
@@ -0,0 +1,38 @@
...
+
+The HTTP-01 challenge endpoint can be found in the logs when the `pre-check` fails. If it does not appear in the logs, you can check the challenge URL by `kubectl`command.
```
meyskens, line 23: I would not recommend looking in the logs for this at all.
I would replace this part by looking at either the challenge or ingress resourve

content/en/docs/configuration/externalloadbalancer.md
```diff
@@ -0,0 +1,38 @@
...
+
+`<endpoint>` is the URL used to test the HTTP-01 from the certificate `Issuer`. For Let's Encrypt for example, the URL is formed like <domain>/.wellknown/<hash>
```
meyskens, line 25:
```suggestion
`<endpoint>` is the URL used to test the HTTP-01 from the certificate `Issuer`. For Let's Encrypt for example, the URL is formed like `<domain>/.well-known/acme-challenge/<hash>`
```

content/en/docs/configuration/externalloadbalancer.md
```diff
@@ -0,0 +1,38 @@
...
+
+
+## Load Balancer HTTP endpoints
+
+If you are using a Load Balancer, you should be able to configure the Load Balancer protocol as HTTP, HTTPS, TCP, UDP. Several Load Balancer now offer free SSL certificates with Let's Encrypt.
```
meyskens, line 30:
```suggestion
If you are using a Load Balancer, you should be able to configure the Load Balancer protocol as HTTP, HTTPS, TCP, UDP. Several Load Balancer now offer free TLS certificates with Let's Encrypt.
```

content/en/docs/configuration/externalloadbalancer.md
```diff
@@ -0,0 +1,38 @@
...
+
+When using HTTP(s) protocols for your Load Balancer, it can intercept the challenge URL to replace the response's verification hash with their hash.
```
meyskens, line 32: do you have an example of a provider doing this? this is just plain wrong according to the ACME spec and should be reported to these services!

content/en/docs/configuration/externalloadbalancer.md
```diff
@@ -0,0 +1,38 @@
...
+
+In this case, Cert-Manager will fail `did not get expected response when querying endpoint, expected 'xxxx' but got: yyyy (truncated)`.
```
meyskens, line 34:
```suggestion
In this case, cert-manager will fail `did not get expected response when querying endpoint, expected 'xxxx' but got: yyyy (truncated)`.
```

content/en/docs/configuration/externalloadbalancer.md
```diff
@@ -0,0 +1,38 @@
...
+
+This kind of error can be thrown for multiple reasons. This case shows a correctly formated response, but not the expected one. In the case of OVH host, the HTTP requests replaced the reponse with OVH's Let's Encrpyt code and provoked this error.
```
meyskens, line 36: oh... thanks! answered my previous question

meyskens, line 36:
```suggestion
This kind of error can be thrown for multiple reasons. This case shows a correctly formated response, but not the expected one. In the case of OVH, the HTTP requests replaced the reponse with OVH's Let's Encrpyt code and provoked this error.
```

meyskens:
Thanks for this! I think this will help a bunch of people. 
Just some small nits here and there :) 
/ok-to-test

nd2014-public:
> Thanks for this! I think this will help a bunch of people.
> Just some small nits here and there :)
> /ok-to-test

Yes, so many typos errors... Thanks for your review.

Hope this will help.

meyskens:
@nd2014-public I see you closed the comments but the content hasn't changed, am I missing a commit?

content/en/docs/configuration/externalloadbalancer.md
meyskens: @nd2014-public would be nice, if you also have something like a form link?
Do you also have config to replicate this? looking into the OVH Kubernetes service now and can only seem to keep in in TCP mode not HTTP(S) mode

nd2014-public: This is the conversation where _az helped me about this problem : 

https://community.letsencrypt.org/t/first-certificate-behind-an-external-load-balancer/136428

I used a LB from OVH, not a managed Kubernetes cluster from OVH.

meyskens: oh okay from what I understood these are deprecated?

nd2014-public: Not sure we are talking about the same thing.

I use dedicated servers from OVH and a Load Balancer. I am using Cert-manager with Rancher on hybrid cluster from multiple hosts. The Load Balancer can be configured using TCP, UDP, HTTP, HTTPS… Firstly I used HTTP + HTTPs frontends and this provoked the bug with the response interception.

Using TCP/UDP protocols solved the problem.

nd2014-public:
> @nd2014-public I see you closed the comments but the content hasn't changed, am I missing a commit?

No I just forgot to commit the changes ... Mea Culpa ...

content/en/docs/configuration/externalloadbalancer.md
meyskens: Okay I was assuming you used the OVH Kunernetes service. Just looked at all producs... this LB service is quite hidden :D

content/en/docs/configuration/externalloadbalancer.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Load Balancer HTTP endpoints
+
+If you are using a Load Balancer, you should be able to configure the Load Balancer protocol as HTTP, HTTPS, TCP, UDP. Several Load Balancer now offer free TLS certificates with Let's Encrypt.
+
+When using HTTP(s) protocols for your Load Balancer, it can intercept the challenge URL to replace the response's verification hash with their hash.
+
+In this case, cert-manager will fail `did not get expected response when querying endpoint, expected 'xxxx' but got: yyyy (truncated)`.
+
+This kind of error can be thrown for multiple reasons. This case shows a correctly formated response, but not the expected one. In the case of OVH, the HTTP requests replaced the reponse with OVH's Let's Encrpyt code and provoked this error.
```
meyskens, line 36: I think we can leave out OVH in here given they have 2 completely different LB services which will cause confusion

nd2014-public: Ok, I'll be more general and precise the fact that it does not apply to maanged Kubernetes services.

nd2014-public (commented):
Everything seem Ok for me now.

meyskens:
/approve
/lgtm

Thanks!

Issue #360, opened 2020-11-05
Venafi Cloud "Zone" Specification


paulternate:
Earlier this year Venafi Cloud stopped supporting Zone specification by name, and instead now requires the Zone ID (GUID). 

Below is an updated working example:

```
apiVersion: cert-manager.io/v1alpha2
kind: Issuer
metadata:
  name: cloud-venafi-issuer
  namespace: <NAMESPACE YOU WANT TO ISSUE CERTIFICATES IN>
spec:
  venafi:
    zone: "801bdbd0-8587-11ea-b487-4d978b4efe3d" # Set this to the Venafi policy zone you want to use
    cloud:
      apiTokenSecretRef:
        name: cloud-secret
        key: apikey
```

PR #364, opened 2020-11-11
Use the GUID zone for venafi Cloud example

meyskens:
Fixes https://github.com/cert-manager/website/issues/360


wallrj (approved):
Thanks, I discovered this while testing against Vcloud yesterday.

/lgtm
/approve

PR #366, opened 2020-11-12
Add preferredChain also into the ACME config docs

meyskens:
Copies over the part from the release note. Also updates on the latest LE tranition update and adds a note that it works with all ACME servers


jakexks (approved):
/lgtm

PR #373, opened 2020-11-19
Make the year of the date clear on docs/configuration/acme/ page

shuuji3:
Hello🙂 When reading about the replacement of CA, I had a little confusion about whether the CA  had been replaced already because the date can be the past. 

This PR adds the year of replacement to make it clear.

Thank you


meyskens:
Good catch! However January 2020 feels like 10 years ago already....

/ok-to-test
/approve
/lgtm

shuuji3:
Thanks! Yes, there are too many things in 2020... 💫

PR #379, opened 2020-11-25
Added needed escape to second sed expression since

claycooper:
slash is being used as the separator in the expression.

Added option to base64 to prevent output string
from being line wrapped.

Signed-off-by: Clay Cooper <claycooper@users.noreply.github.com>


meyskens:
Thanks! Looks good to me

/lgtm
/approve
/pk-to-test

PR #381, opened 2020-11-26
Updating section for MatchLabels

justinas-b:
At the moment documentation states that when using `MatchLabels` selector for `Issuers` at least on label should be matched in order for solver to be chosen, however it seems that all of the labels defined in `MatchLabels` stanza *must* also be applied on certificate.

Signed-off-by: Justinas B <12399634+justinas-b@users.noreply.github.com>


meyskens:
Can you also update the example below?

justinas-b:
Hey @meyskens , updated examples also

meyskens:
/approve
/lgtm
/ok-to-test

Thanks!

PR #390, opened 2020-12-16
docs(installation/kubernetes): update kubeprod info

jai:
kubeprod seems to be available for EKS now, based on this: https://github.com/bitnami/kube-prod-runtime. Updating docs to reflect that.


meyskens:
Thanks!

/ok-to-test
/approve
/lgtm

PR #392, opened 2020-12-18
Remove Helm 2 instructions

meyskens:
As per https://helm.sh/blog/helm-2-becomes-unsupported/ Helm 2 is no longer supported, so we should remove it from our instructions.


wallrj (approved):
/lgtm

PR #407, opened 2021-01-20
Update site copyright to CNCF guidelines

meyskens:
Changes part of https://github.com/cncf/toc/issues/557


munnerz:
Seems we now have two different copyright claims in our footer after this change - is that correct? I'd have thought we'd remove the "cert-manager authors" one?

Screenshot 2021-01-21 at 10 28 01

layouts/partials/footer.html
```diff
@@ -20,10 +20,13 @@
         {{ with .Site.Params.copyright }}
         <div class="row">
           <div class="col-12">
-            <small class="text-white">&copy; {{ now.Year}} {{ .}} {{ T "footer_all_rights_reserved" }}</small>
+            <small class="text-white">&copy; {{ now.Year}} The cert-manager Authors</small>
           </div>
           <div class="col-12">
-            <small class="text-white">cert-manager was originally created by <a href="https://jetstack.io"><span id="jetstack-footer">{{ with resources.Get "icons/jetstack.svg" }}{{ ( . | minify).Content | safeHTML }}{{ end }}</span></a></small>
```
munnerz, line 26: Do we still want to keep the "created by Jetstack" text? From what I understand, this _is_ permitted by LF

layouts/partials/footer.html
```diff
@@ -20,10 +20,13 @@
...
+            <small class="text-white">&copy; {{ now.Year}} The Linux Foundation. All rights reserved. The Linux Foundation has registered trademarks and uses trademarks.</small>
+          </div>
+          <div class="col-12">
+            <small class="text-white">The Linux Foundation has registered trademarks and uses trademarks. For a list of trademarks of The Linux Foundation, please see our <a class="text-white" href="https://www.linuxfoundation.org/trademark-usage">Trademark Usage page</a></small>
```
munnerz, line 29: Is `The Linux Foundation has registered trademarks and uses trademarks. For a list of trademarks ...` correct? So many 'trademark' 😆

munnerz:
Ah, looking at projectcontour.io, this seems like it might all be correct 👀 

Screenshot 2021-01-21 at 10 30 13

layouts/partials/footer.html
meyskens: I have looked at a few CNCF sites and didn't see any, removed it till we have clarity on it

meyskens:
Yeah the coputight is weird as it applies differently to content than trademarks

layouts/partials/footer.html
meyskens: it's their core business ;)

munnerz:
This looks good to me - @mattbates can you take a look (you can also view a deploy preview by following [the status link at the bottom of the PR](https://deploy-preview-407--cert-manager.netlify.app/)) and write `/hold cancel` in a comment if you are happy, and this will merge? 😄

/approve
/lgtm
/hold
/cc @mattbates 
/assign @mattbates

layouts/partials/footer.html
mattbates: I'd like for it to remain. It's on other project sites (eg - [Linkerd](https://linkerd.io/), [Kyverno](https://kyverno.io/), [Kuma](https://kuma.io/). As @munnerz says, it's also permitted by the [website-guidelines](https://github.com/cncf/foundation/blob/master/website-guidelines.md):

>It is OK to say that, e.g., “Prometheus was originally created by Soundcloud” or “Kubernetes builds upon 15 years of experience of running production workloads at Google,” but the origin company should not otherwise be referred to on the project homepage.

How about text that is similar to the message used by Kuma? I really like the sentiment, and it's of course very true for cert-manager and it's 250+ contributors.

> Kuma was originally created by Kong, Inc. and made successful by its open community.

mattbates:
See my comments in https://github.com/cert-manager/website/pull/407#discussion_r562606796 - grateful if we can come up with something to this effect.

layouts/partials/footer.html
meyskens: Thanks for the link! the guidelines in teh TOC ticket didn't mention it, happy to add it back as it was!

meyskens:
fixed!

/assign @mattbates 
/unassign

maelvls:
I adjusted the footer a bit:

(before/after)

footer-before footer-after

content/en/_index.html
```diff
@@ -106,3 +106,16 @@ <h4>Follow us on Twitter</h4>
...
 		</div>
 </section>
+<section>
+	<div class="col">
+		<div class="row justify-content-center">
+			<div class="col-sm-12 col-md-2 col-lg-5 text-center">
+				<h4>cert-manager is a CNCF member project</h4>
+				</br>
+				<p>
+					<img src="/png/cncf.png" alt="cert-manager is a CNCF member project" width="400">
```
munnerz, line 116: Any particular reason why in the footer partial we use `{{ with resources.Get "icons/jetstack.svg" }}` and here we directly reference the png? Should they both be done in the same way?

munnerz:
One last comment, which maybe can be ignored :). Remove the `/hold` once you're ready!

/lgtm

content/en/_index.html
maelvls: Looks like `{{ with resources.Get "icons/jetstack.svg" }}` embeds the svg into the HTML page, which I guess I can't do (unless I base64 encode it I guess)

PR #413, opened 2021-01-28
Mention that Venafi TPP server is the only external dep in e2e

maelvls:
As mentioned in [#cert-manager-dev](https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1611771108029400), I was wondering about how we relied on the Venafi TPP server and who to ask in case of emergency

I added a paragraph about this, not sure it is the right place or if it makes sense though

Signed-off-by: Maël Valais <mael@vls.dev>

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -199,3 +199,4 @@ subchart
 cainjector
 Velero
 istio-csr
+Cleary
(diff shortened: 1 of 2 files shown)


wallrj (commented):
I notice that @simplyzee is in the process of setting up a dedicated TPP instance. So perhaps we should wait for that to be ready. See PR link above.

PR #417, opened 2021-02-01
add -w0 to base64 command to get a singel line

chrisdecker1201:
I think it's easier to get a valid base64 string with the -w0 argument which makes a single line instead of a multiline output.

### DIFF
--- a/content/en/docs/configuration/ca.md
+++ b/content/en/docs/configuration/ca.md
@@ -28,7 +28,7 @@ being the `cert-manager` namespace, however can be configured using the
 Below is an example of a secret resource that will be used for signing. Take
 note of the index keys used for each field as these are required in order for
 cert-manager to find the certificate and key. Also note that, like all secrets,
-data must be base64 encoded. The command `$ cat crt.pem | base64` should help
+data must be base64 encoded. The command `$ cat crt.pem | base64 -w0` should help
 you here.
 
 ```yaml
--- a/content/en/docs/configuration/ca.md
+++ b/content/en/docs/configuration/ca.md
@@ -28,8 +28,9 @@ being the `cert-manager` namespace, however can be configured using the
 Below is an example of a secret resource that will be used for signing. Take
 note of the index keys used for each field as these are required in order for
 cert-manager to find the certificate and key. Also note that, like all secrets,
-data must be base64 encoded. The command `$ cat crt.pem | base64 -w0` should help
-you here.
+data must be base64 encoded. The command `$ cat crt.pem | base64 -w0` should help you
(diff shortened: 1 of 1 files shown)


maelvls:
Good point, the line wrapping that base64 often causes issues, `-w0` would make sense in the context of creating a secret.

Note that `-w0` is specific to GNU base64. On BSD-based systems, the correct command would be

```sh
base64 -b0
``` 

Not sure what to do with this variance between platforms... would it still make sense to add `-w0` when we know that some users will get an error running it?

chrisdecker1201:
Oh. I did not know their are different CLI arguments on different systems. Maybe add the differences into the documentation.

~~~
Also note that, like all secrets, data must be base64 encoded. The command `$ cat crt.pem | base64 -w0` should help
you on GNU-based systems (Debian, Ubuntu, etc.) and `$ cat crt.pem | base64 -b0` on BSD-based systems.
~~~

maelvls:
Agreed! Adding a note about base64 on BSD-based systems like macOS makes sense to me 🙂

content/en/docs/configuration/ca.md
```diff
@@ -28,7 +28,7 @@ being the `cert-manager` namespace, however can be configured using the
 Below is an example of a secret resource that will be used for signing. Take
 note of the index keys used for each field as these are required in order for
 cert-manager to find the certificate and key. Also note that, like all secrets,
-data must be base64 encoded. The command `$ cat crt.pem | base64` should help
+data must be base64 encoded. The command `$ cat crt.pem | base64 -w0` should help
```
maelvls, line 31:
```suggestion
data must be base64 encoded. The command `$ cat crt.pem | base64 -w0` should help you
on GNU-based systems (Debian, Ubuntu, etc.) and `$ cat crt.pem | base64 -b0` on BSD-based systems (most notably macOS).
```

content/en/docs/configuration/ca.md
```diff
@@ -28,7 +28,7 @@ being the `cert-manager` namespace, however can be configured using the
...
 you here.
```
maelvls, line 32:
```suggestion
```suggestion
data must be base64 encoded. The command `$ cat crt.pem | base64 -w0` should help you
on GNU-based systems (Debian, Ubuntu, etc.) and `$ cat crt.pem | base64 -b0` on BSD-based systems (most notably macOS).
```
you here.
```

**Comment by maelvls** [comment]
Do you mind if I edit your commit to add the DCO? i.e.

```
Signed-off-by: Christian Decker <christian.decker@lookslikematrix.de>
```

maelvls:
Thank you so much for your contribution!! 🙏

/approve

PR #420, opened 2021-02-03
Move externalloadbalancer to acme/http01/

JoshVanL:


munnerz:
I think you may want an entry in the redirects file for this, if the page has been previously released? :) https://github.com/cert-manager/website/blob/master/_redirects

JoshVanL:
Course, thanks @munnerz

PR #428, opened 2021-02-07
fix(install by kubectl): provide a temporary command to install

abserari:
Signed-off-by: Abser ari <abserari@gmail.com>


meyskens:
This should not have been published yet. reverted in https://github.com/cert-manager/website/pull/430

/close

PR #442, opened 2021-02-17
Updates supported version policy to be n and n-1.

JoshVanL:

/hold

### DIFF
--- a/content/en/docs/concepts/project-maturity.md
+++ b/content/en/docs/concepts/project-maturity.md
@@ -12,6 +12,13 @@ in both production and non-production clusters.
 
 The cert-manager API is currently at `v1` and is stable.
 
+## Supported Versions
+
+cert-manager supports versions for the current and last minor release.  This
+means that if the latest release is `v1.2`, cert-manager supports the versions
+`v1.2` and `v1.1`. Support means that bug fixes will be backported to supported
+minor versions and patch releases will be made.
+
 ## Compatibility
 
 cert-manager has a hard guarantee of compatibly with the current stable upstream
--- a/content/en/docs/contributing/release-process.md
+++ b/content/en/docs/contributing/release-process.md
@@ -270,10 +270,11 @@ A patch release contains critical bug fixes for the project.  They are managed o
 an ad-hoc basis, and should only be required when critical bugs/regressions are
 found in the release.
 
-We will only perform patch release for the **current** version of cert-manager.
+We will only perform patch release for the **current** and **last** minor
+version of cert-manager.
 
 Once a new minor release has been cut, we will stop providing patches for the
-version before it.
+one before last version, [see](../../concepts/project-maturity/).
 
 ### Release schedule
 


content/en/docs/contributing/release-process.md
```diff
@@ -270,10 +270,11 @@ A patch release contains critical bug fixes for the project.  They are managed o
 an ad-hoc basis, and should only be required when critical bugs/regressions are
 found in the release.
 
-We will only perform patch release for the **current** version of cert-manager.
+We will only perform patch release for the **current** and **last** minor
```
irbekrm, line 273: Suggestion: perhaps we could use the word 'previous' instead of 'last' here or just describe them as `n`, `n-1` ? 'last' is confusing as we seem to sometimes use it to describe 'latest'.

content/en/docs/contributing/release-process.md
```diff
@@ -270,10 +270,11 @@ A patch release contains critical bug fixes for the project.  They are managed o
...
+version of cert-manager.
 
 Once a new minor release has been cut, we will stop providing patches for the
-version before it.
+one before last version, [see](../../concepts/project-maturity/).
```
irbekrm, line 277: Suggestion: perhaps 'last' -> 'latest' here would be clearer?

content/en/docs/concepts/project-maturity.md
```diff
@@ -12,6 +12,13 @@ in both production and non-production clusters.
 
 The cert-manager API is currently at `v1` and is stable.
 
+## Supported Versions
+
+cert-manager supports versions for the current and last minor release.  This
```
irbekrm, line 17: Suggestion: perhaps 'previous' instead of 'last' would be clearer?

irbekrm (commented):
Looks good, I've just got some suggestions around 'last'/'latest'/'previous' wording.

content/en/docs/concepts/project-maturity.md
```diff
@@ -12,6 +12,13 @@ in both production and non-production clusters.
...
+means that if the latest release is `v1.2`, cert-manager supports the versions
+`v1.2` and `v1.1`. Support means that bug fixes will be backported to supported
+minor versions and patch releases will be made.
```
wallrj, line 20: But why support the last minor release? 
Maybe add a note about the rationale, perhaps in terms of providing support for releases in the last X months.
But does this mean that upon releasing 1.3, we will no longer support 1.1 and no longer support people trying to run CM on K8S < 1.16?

OR

Are we making this policy change in order to continue to support people on older Kubernetes, which are now only supported by cert-manager 1.1.x, in which case, the docs should say something about that.

In https://github.com/jetstack/cert-manager/issues/3413 @meyskens proposed calling it 1.1 LTS, but if that is our policy it would be good to add a paragraph explaining how long we will support that LTS branch.

My2c: I personally don't think we should change the policy. I like the current policy of only supporting the current released version.

maelvls:
The "supported releases" paragraph looks similar to what's in #452, should I move my work to `project-maturity.md`?

SgtCoDFish:
I think this is obsoleted by https://cert-manager.io/docs/installation/supported-releases/ from #452 so I'll close this PR as a cleanup.

If I've missed something - sorry and please re-open :sweat_smile:

PR #447, opened 2021-02-18
WIP Adds repo structure description

### LINKED ISSUE #414: Explain cert-manager repo structure
For a new contributor it could maybe be useful if the `Contributing` section of the website contained some information about the repository structure (i.e what cert manager's component's functionality goes into which subdirectory)

irbekrm:
fixes #414 

/kind documentation

### DIFF
--- a/content/en/docs/contributing/_index.md
+++ b/content/en/docs/contributing/_index.md
@@ -16,6 +16,7 @@ external cert-manager components:
 - [DCO sign off](./sign-off/)
 - [Developing with kind](./kind/)
 - [Running end-to-end tests](./e2e/)
+- [Repository structure](./repo-structure/)
 - [Release process](release-process/)
 - [Feature policy](policy/)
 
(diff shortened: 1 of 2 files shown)


JoshVanL:
Cautious about adding this at this granularity as is, some things are subject to change quite rapidly so become out of date. Perhaps we can achieve the same at a more course level?

I think the page might also be malformed as I don’t see it in the deploy.

/assign @irbekrm

irbekrm:
Thanks for looking at it @JoshVanL !

> Cautious about adding this at this granularity as is, some things are subject to change quite rapidly so become out of date. 

Agreed about the risk of becoming out of date. I will probably leave out the separate certificate control loop descriptions. I was going to get rid of large parts of the rest (I generated it with `tree` originally).

> I think the page might also be malformed as I don’t see it in the deploy.

Thanks for spotting that- I had not added it to the index.

maelvls:
@irbekrm Do you think we can merge it? 😅

SgtCoDFish:
Given this has been open for so long, I'm gonna close it just to clean up open PRs!

(I quite like the idea here, but I'd probably choose to have something like this in the cert-manager repo itself rather than here? Still, I think this PR has lived long enough!)

PR #455, opened 2021-02-26
move cert-manager from jetstack to cert-manager

jakexks:
Update references to jetstack/cert-manager to cert-manager/cert-manager

Context: https://github.com/jetstack/cert-manager/issues/3671


maelvls:
Is it ever going to happen? 😞

PR #458, opened 2021-03-04
Fix incorrectly documented default duration

jakexks:


content/en/docs/faq/_index.md
```diff
@@ -51,7 +51,7 @@ spec:
 ```
 
 ### If `renewBefore` or `duration` is not defined, what will be the default value?
-cert-manager will default to a `duration` of 30 days with a `renewBefore` of 30 days. 
+cert-manager will default to a `duration` of [https://github.com/jetstack/cert-manager/blob/v1.2.0/pkg/apis/certmanager/v1/const.go#L26](90 days) with a `renewBefore` of [https://github.com/jetstack/cert-manager/blob/v1.2.0/pkg/apis/certmanager/v1/const.go#L32](30 days).
```
JoshVanL, line 54: Can we also include:

```
 If `renewBefore` is not set and the duration of the signed certificate is shorter
or equal to 30 days, the `renewBefore` time will be set to 2/3 of the signed
certificate validity duration.
```

JoshVanL (changes requested):
/assign @jakexks

PR #460, opened 2021-03-06
Fix the SVG logo file using the source .ai file

### LINKED ISSUE #369: SVG Broken
The rendering of the cert-mamager SVG logo seems to be broken on recent rendering engines like Chrome beta or some releases of Firefox.

maelvls:
Fixes #369


jakexks:
I think this should be cherry-picked all the way back, but maybe for sanity's sake just back to 0.16?

maelvls:
before vs. after:
before after

PR #473, opened 2021-03-15
Updates CR concept docs with UserInfo fields and fields being immutable

JoshVanL:

Adds description for the new UserInfo fields, and a note that all spec fields and cert-manager annotations are immutable.

We should merge this when we do the release.


maelvls:
Nit: would it be possible to add a link to the cert-manager PR where this feature was added in this PR? That helps a lot when trying to navigate 👍

PR #488, opened 2021-03-26
SelfSigned Doc Changes

SgtCoDFish:
Quite a large rework of the `SelfSigned` issuer docs, intended to highlight potential gotchas and clarify language.

### DIFF
--- a/content/en/docs/configuration/selfsigned.md
+++ b/content/en/docs/configuration/selfsigned.md
@@ -6,27 +6,31 @@ type: "docs"
 ---
 
 The `SelfSigned` issuer doesn't represent a certificate authority as such, but
-instead denotes that certificates will be signed through "self signing" using
-the given private key. This means that the provided private key of the resulting
-certificate will be used to sign its own certificate.
+instead denotes that certificates will "sign themselves" using a given private
+key. In other words, the private key of the certificate will be used to sign
+the certificate itself.
 
-This `Issuer` type is useful for bootstrapping the CA certificate key pair for
-some Private Key Infrastructure (PKI), or for otherwise creating simple
-certificates.  Clients consuming these certificates have _no way_ to trust this
-certificate since there is no CA signer apart from itself, and as such, would be
-forced to trust the certificate as is.
+This `Issuer` type is useful for bootstrapping a root certificate for a
+custom PKI (Public Key Infrastructure), or for otherwise creating simple
+ad-hoc certificates.
 
-> Note: `CertificateRequests` that reference a self signed certificate _must_
-> also contain the annotation `cert-manager.io/private-key-secret-name`. This is
-> because without access to the private key of the certificate request, the
-> `CertificateRequest` will be unable to self sign the certificate. This
-> annotation is added automatically by the `Certificate` controller.
+There are important [caveats](#caveats) - including security issues - to
+consider with `SelfSigned` issuers; in general you'd likely  want to use a
+[`CA`](../ca/) issuer rather than a `SelfSigned` issuer. That said,
+`SelfSigned` issuers are really useful for initially [bootstrapping](#bootstrapping-ca-issuers)
+a `CA` issuer.
+
+> Note: a `CertificateRequest` that references a self-signed certificate _must_
+> also contain the `cert-manager.io/private-key-secret-name` annotation since
+> the private key corresponding to the `CertificateRequest` is required to
+> sign the certificate. This annotation is added automatically by the
+> `Certificate` controller.
 
 ## Deployment
 
-Since the `SelfSigned` `Issuer` requires no dependency on any other resource to be
-configured, it is the simplest to configure. All that is required is for the
-`SelfSigned` stanza to be present in the issuers spec.
+Since the `SelfSigned` issuer has no dependency on any other resource, it is
+the simplest to configure. Only the `SelfSigned` stanza is required to be
+present in the issuer spec, with no other configuration required:
 
 ```yaml
 apiVersion: cert-manager.io/v1
@@ -36,9 +40,76 @@ metadata:
   namespace: sandbox
 spec:
   selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: selfsigned-cluster-issuer
+spec:
+  selfSigned: {}
 ```
-Optionally, you can specify [CRL](https://en.wikipedia.org/wiki/Certificate_revocation_list) Distribution Points. An array of strings each of which identifies the location of the CRL from which the revocation of this certificate can be checked:
+
+Once deployed, you should be able to see immediately that the issuer is ready
+for signing:
+
+```bash
+$ kubectl get issuers  -n sandbox -o wide selfsigned-issuer
+NAME                READY   STATUS                AGE
+selfsigned-issuer   True                          2m
+
+$ kubectl get clusterissuers -o wide selfsigned-cluster-issuer
+NAME                        READY   STATUS   AGE
+selfsigned-cluster-issuer   True             3m
 ```
+
+### Bootstrapping `CA` Issuers
+
+One of the ideal use cases for `SelfSigned` issuers is to bootstrap a custom
+root certificate for a private PKI, including with the cert-manager [`CA`](../ca/)
+issuer.
+
+The YAML below will create a `SelfSigned` issuer, issue a root certificate and
+use that root as a `CA` issuer:
+
+```yaml
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: selfsigned-issuer
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: osm-ca
+  namespace: sandbox
+spec:
+  isCA: true
+  commonName: osm-system
+  secretName: osm-ca
+  issuerRef:
+    name: selfsigned-issuer
+    kind: ClusterIssuer
+    group: cert-manager.io
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: osm-ca
+  namespace: sandbox
+spec:
+  ca:
+    secretName: osm-ca
+```
+
+### CRL Distribution Points
+
+You may also optionally specify [CRL](https://en.wikipedia.org/wiki/Certificate_revocation_list)
+Distribution Points as an array of strings, each of which identifies the location of a CRL in
+which the revocation status of issued certificates can be checked:
+
+```yaml
 ...
 spec:
   selfSigned:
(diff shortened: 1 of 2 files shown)


content/en/docs/configuration/selfsigned.md
```diff
@@ -58,3 +52,47 @@ selfsigned-issuer   True                          2m
...
+manually set a Subject, the Issuer will be empty and the certificate will
+technically be invalid.
+
+Validation of this specific area of the spec is patchy and varies between TLS
+libraries, but there's always the risk that a library will improve its
+validation - entirely within spec - in the future and break your app if you're
+using a certificate with an empty Issuer DN.
+
+To avoid this, be sure to set a Subject for `SelfSigned` certs. In any case,
+cert-manager will emit a Kubernetes warning event if it detects that an invalid
```
irbekrm, line 97: Suggestion: maybe we could specify what event or perhaps just point at the code that does that https://github.com/jetstack/cert-manager/blob/master/pkg/controller/certificaterequests/selfsigned/selfsigned.go#L140? Not sure if that's needed, but I guess I would be curious what event/log message to look for.

irbekrm (approved):
Looks great to me.

As a user I always appreciate when the docs for some software that I am using talks about general good practices/how to use it right, not just what commands to run- so it's nice to see that we are doing things that way too!

content/en/docs/configuration/selfsigned.md
```diff
@@ -6,27 +6,29 @@ type: "docs"
...
-> because without access to the private key of the certificate request, the
-> `CertificateRequest` will be unable to self sign the certificate. This
-> annotation is added automatically by the `Certificate` controller.
+instead denotes that certificates will "sign themselves" using a given private
+key. In other words, the private key of the certificate will be used to sign
+the certificate itself.
+
+This `Issuer` type is useful for bootstrapping a root certificate for a
+custom PKI (Public Key Infrastructure), or for otherwise creating simple
+ad-hoc certificates.
```
JoshVanL, line 15: I think this change removes some of the intention of the original text, which is that the self signed issuer can be used to bootstrap a CA Issuer. Can we add that back in?


```yaml
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: selfsigned-issuer
spec:
  selfSigned: {}
---
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: osm-ca
  namespace: sandbox
spec:
  isCA: true
  commonName: osm-system
  secretName: osm-ca
  issuerRef:
    name: selfsigned-issuer
    kind: ClusterIssuer
    group: cert-manager.io
---
kind: Issuer
metadata:
  name: osm-ca
  namespace: sandbox
spec:
  ca:
    secretName: osm-ca
```

JoshVanL (changes requested):

content/en/docs/configuration/selfsigned.md
```diff
@@ -6,27 +6,29 @@ type: "docs"
...
+> `Certificate` controller.
 
 ## Deployment
 
-Since the `SelfSigned` `Issuer` requires no dependency on any other resource to be
-configured, it is the simplest to configure. All that is required is for the
-`SelfSigned` stanza to be present in the issuers spec.
+Since the `SelfSigned` `Issuer` requires no dependency on any other resource to
+be configured, it is the simplest to configure. Only the `SelfSigned` stanza
+is required to be present in the issuer spec:
```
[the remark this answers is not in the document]
SgtCoDFish: you raise a good point, yeah - I think the intention of both the original page and my rework is to convey that you don't need to add _any other_ configuration to use a `SelfSigned` issuer. I'll try to make that clearer.

content/en/docs/configuration/selfsigned.md
```diff
@@ -58,3 +52,47 @@ selfsigned-issuer   True                          2m
...
+
+### Certificate Validity
+
+One side-effect of a certificate being self-signed is that its Subject DN and
+its Issuer DN are identical. The X.509[RFC 5280, section 4.1.2.4](https://tools.ietf.org/html/rfc5280#section-4.1.2.4) requires that:
+
+> The issuer field MUST contain a non-empty distinguished name (DN).
+
+However, self-signed certs don't have a subject DN set by default. Unless you
+manually set a Subject, the Issuer will be empty and the certificate will
```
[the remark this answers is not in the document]
SgtCoDFish: I'm going to mention `spec.subject` around line 96, as in your other comment. I'll leave this mention of the subject to be broader, but change it to "Subject DN"

content/en/docs/configuration/selfsigned.md
SgtCoDFish: I would personally _never_ have gotten a hint about bootstrapping `CA` issuers using `SelfSigned` from the original document. I've read it over a few times now and I just can't see any suggestion of that. It talks generally about bootstrapping PKI, but I can't see any explicit link to `CA` issuers. Am I missing something?

It definitely seems like a really worthwhile thing to _add_ and a good use-case to highlight, though, including the config you posted.

Is there a reason the CA issuer in your example uses `apiVersion: cert-manager.io/v1alpha2` or am I OK to change it to `apiVersion: cert-manager.io/v1`?

JoshVanL: Yep, you're right, it wasn't even very well implied this is an intended and recommended way to bootstrap a CA Issuer.

Yes please change it to `v1`. Copy paste from some old manifest examples I had :smile:

SgtCoDFish:
I think I've now addressed all the comments that were raised :+1:

JoshVanL:
Up to you whether you want to hold this till v1.3 when that event trigger will be true

SgtCoDFish:
The final forcepush I just made was to add this change:

```diff
diff --git a/content/en/docs/configuration/selfsigned.md b/content/en/docs/configuration/selfsigned.md
index fe6363a..fef8a1a 100644
--- a/content/en/docs/configuration/selfsigned.md
+++ b/content/en/docs/configuration/selfsigned.md
@@ -148,6 +148,6 @@ To avoid this, be sure to set a Subject for `SelfSigned` certs. This can be
 done by setting the `spec.subject` on a cert-manager `Certificate` object
 which will be issued by a `SelfSigned` issuer.
 
-In any case, cert-manager will emit a Kubernetes [warning event](https://github.com/jetstack/cert-manager/blob/45befd86966c563663d18848943a1066d9681bf8/pkg/controller/certificaterequests/selfsigned/selfsigned.go#L140)
+Starting in version 1.3, cert-manager will emit a Kubernetes [warning event](https://github.com/jetstack/cert-manager/blob/45befd86966c563663d18848943a1066d9681bf8/pkg/controller/certificaterequests/selfsigned/selfsigned.go#L140)
 of type `BadConfig` if it detects that an certificate is being created
 by a `SelfSigned` issuer which has an empty Issuer DN.
```

content/en/docs/configuration/selfsigned.md
```diff
@@ -36,25 +40,114 @@ metadata:
...
+```
+
+### Bootstrapping `CA` Issuers
+
+One of the ideal use cases for `SelfSigned` issuers is to bootstrap a custom
+root certificate for a private PKI, including with the cert-manager [`CA`](../ca/)
+issuer.
+
+The below YAML will create a `SelfSigned` issuer, issue a root certificate and
+in use that root as a `CA` issuer:
```
wallrj, line 72:
```suggestion
The YAML below will create a `SelfSigned` issuer, issue a root certificate and
use that root as a `CA` issuer:
```

content/en/docs/configuration/selfsigned.md
```diff
@@ -36,25 +40,114 @@ metadata:
...
+spec:
+  ca:
+    secretName: osm-ca
 ```
+
+### CRL Distribution Points
+
+You may also optionally specify [CRL](https://en.wikipedia.org/wiki/Certificate_revocation_list)
+Distribution Points as an array of strings, each of which identifies the location of a CRL in
+which the revocation status of issued certificates can be checked:
```
wallrj, line 110:
```suggestion
You may also optionally specify [CRL](https://en.wikipedia.org/wiki/Certificate_revocation_list)
Distribution Points as an array of URLs, each of which identifies the location of a CRL in
which the revocation status of issued certificates can be checked:
```

Do they have to be URLs? I assumed so.

content/en/docs/configuration/selfsigned.md
```diff
@@ -36,25 +40,114 @@ metadata:
...
+
+One side-effect of a certificate being self-signed is that its Subject DN and
+its Issuer DN are identical. The X.509 [RFC 5280, section 4.1.2.4](https://tools.ietf.org/html/rfc5280#section-4.1.2.4)
+requires that:
+
+> The issuer field MUST contain a non-empty distinguished name (DN).
+
+However, self-signed certs don't have a subject DN set by default. Unless you
+manually set a set a certificate's Subject DN, the Issuer DN will be empty
+and the certificate will technically be invalid.
```
wallrj, line 140:
```suggestion
However, self-signed certs don't have a subject DN set by default. Unless you
manually set a certificate's Subject DN, the Issuer DN will be empty
and the certificate will technically be invalid.
```

content/en/docs/configuration/selfsigned.md
```diff
@@ -36,25 +40,114 @@ metadata:
...
+
+To avoid this, be sure to set a Subject for `SelfSigned` certs. This can be
+done by setting the `spec.subject` on a cert-manager `Certificate` object
+which will be issued by a `SelfSigned` issuer.
 
-Certificates are now ready to be requested by using the `SelfSigned` `Issuer`
-named `selfsigned-issuer` within the `sandbox` namespace.
+Starting in version 1.3, cert-manager will emit a Kubernetes [warning event](https://github.com/jetstack/cert-manager/blob/45befd86966c563663d18848943a1066d9681bf8/pkg/controller/certificaterequests/selfsigned/selfsigned.go#L140)
+of type `BadConfig` if it detects that an certificate is being created
+by a `SelfSigned` issuer which has an empty Issuer DN.
```
wallrj, line 153:
```suggestion
Starting in version 1.3, cert-manager will emit a Kubernetes [warning event](https://github.com/jetstack/cert-manager/blob/45befd86966c563663d18848943a1066d9681bf8/pkg/controller/certificaterequests/selfsigned/selfsigned.go#L140)
of type `BadConfig` if it detects that a certificate is being created
by a `SelfSigned` issuer which has an empty Issuer DN.
```

wallrj (changes requested):
Thanks Ashley,

I spotted a couple of typos.

And optionally, given that there's a paragraph about a 1.3 specific feature, it might be worth merging this into the https://github.com/cert-manager/website/tree/release-next branch instead, along with a short entry in a new 1.3 release notes page: https://cert-manager.io/next-docs/release-notes/

content/en/docs/configuration/selfsigned.md
SgtCoDFish: You raise a good point that I'd not considered: in the spec, they're `DistributionPoint`s: https://tools.ietf.org/html/rfc5280#section-4.2.1.13.

We don't support the full array of options there (and I imagine most implementations of TLS probably don't either). In practice, I can't imagine anyone wanting to specify something here that's not a URL (or URI), so my change from "strings" to "URLs" is probably correct and helpful.

(Honestly, I think my preference would be to remove any reference in the docs to cRLDistributionPoints for SelfSigned issuers because I can't imagine any situation where it would make sense to include it...)

wallrj: Ok. So we'll stick with strings then.

wallrj (approved):
/lgtm

PR #489, opened 2021-03-27
Update ingress.md

mwilmer:
Correct the ingress example to correctly match the api version (v1), adding the mandatory spec.rules[*].http.paths[*].pathType key/value and rewriting the spec.rules[*].http.paths[*].backend to the new format.

### DIFF
--- a/content/en/docs/usage/ingress.md
+++ b/content/en/docs/usage/ingress.md
@@ -33,10 +33,13 @@ spec:
   - host: example.com
     http:
       paths:
-      - backend:
-          serviceName: myservice
-          servicePort: 80
+      - pathType: Prefix
         path: /
+        backend:
+          service:
+            name: myservice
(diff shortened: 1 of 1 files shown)


irbekrm:
Hey @mwilmer ,

Thank you for spotting this- I fixed the yamls in #476 , but forgot those `.md` files 🤦🏼 

Would you please be able to [sign the commit](https://cert-manager.io/docs/contributing/sign-off/) 😅 ?

There is another example with an incorrect Ingress [here](https://github.com/cert-manager/website/blob/master/content/en/docs/tutorials/acme/ingress.md) in case you were interested to add that to the PR, else happy to merge this 😄

irbekrm:
Since I just got rights to override the need for dco sign-off and we should fix this, I am just going to merge 😄

PR #490, opened 2021-03-27
Removes a deprecated note

irbekrm:
Removes a note that seems to be a leftover from when `cert-manager`'s manifest had to be applied with `--validate=false`.

/kind cleanup


SgtCoDFish:
/lgtm

(I've no experience with Openshift  but this looks totally reasonable in context for the file)

irbekrm:
Thanks!

PR #493, opened 2021-03-29
Add linux foundation footer

maelvls:
One of the requirements for the CNCF TOC (https://github.com/cncf/toc/issues/557) is to add a LF footer:

> * [ ]  Website: ensure LF footer is there and [website guidelines](https://github.com/cncf/foundation/blob/master/copyright-notices.md#copyright-notices) followed

In this PR, I add a footer on the landing page. I also changed the license headers that said "the Jetstack cert-manager authors" to "the cert-manager authors" in one single place (the README) instead of in every single file as proposed by the [CNCF guideline](https://github.com/cncf/foundation/blob/cb9c7633/copyright-notices.md#copyright-notices).


content/en/_index.html
```diff
@@ -106,3 +106,17 @@ <h4>Follow us on Twitter</h4>
 			</div>
 		</div>
 </section>
+<section>
+	<div class="col">
+		<div class="row justify-content-center">
+			<div class="col-sm-12 col-md-2 col-lg-5 text-center">
+				<h4>Linkerd is a CNCF member project</h4>
```
meyskens, line 113: I know linkerd is but thanks for noticing

meyskens (changes requested):

scripts/bin/hugo
```diff
@@ -1,19 +1,5 @@
 #!/usr/bin/env bash
 
-# Copyright 2019 The Jetstack cert-manager contributors.
```
munnerz, line 3: Why has all this been removed?

munnerz: I'm not certain, but it seems you may have misinterpreted the linked CNCF guideline:

```
CNCF does not require or recommend that every contributor include their copyright notice in contributed files. See below for more details on why not.
```

From my reading, this implies that _every_ contributor does not need to list the *names* of copyright holders, in favour of having a general statement of "the cert-manager contributors". (rather than saying that we don't need it in every _file_).

I think it's good to keep here, as automated license checkers can also pick this stuff up (which others could be making use of)

maelvls: You are right, I definitely misinterpreted the guidelines 😞 I'll re-add the license notice

Shouldn't we remove the mention of Jetstack though? e.g.:

```diff
-# Copyright 2019 The Jetstack cert-manager contributors.
+# Copyright 2019 The cert-manager contributors.
```

maelvls: Out of curiosity, why do we have to add the license notice to every single file; is it legally necessary?

maelvls:
Looks like my PR is a duplicate of https://github.com/cert-manager/website/pull/407. It was created 3 months ago by @meyskens, and for some reason is still not merged 🤨

I will close my PR and move the work to #407 instead if that's OK 👍

scripts/bin/hugo
munnerz: Yes we should remove the Jetstack part 😄

> Out of curiosity, why do we have to add the license notice to every single file; is it legally necessary?

If someone else copies our files into their own project, they'll need to add this header themselves (and often may not/may forget). Also, if there are multiple licenses in the repository (e.g. our logo is licensed differently to code), it is explicit and clear what license is in use for that content.

That's my thinking anyway! :)

PR #498, opened 2021-04-05
Add tip using Bazel with corporate/custom CA

erikgb:
Suggesting to share some of the headache I had setting up my local cert-manager dev environment. English is not my native language, so feel free to suggest wording improvements... 😄


erikgb:
/unassign @munnerz 

It seems like James is not very active atm.

maelvls:
Great tip, thanks for sharing!

If you like you can use an info block:

```markdown
{{% alert title="Tip" color="info" %}}
...
{{% /alert %}}
```

Docs: https://www.docsy.dev/docs/adding-content/shortcodes/#shortcode-helpers

erikgb:
> Great tip, thanks for sharing!
> 
> If you like you can use an info block:

@maelvls I just copy-modified the formatting of the tip just above. Should I fix that one as well then - to make them conformant?

maelvls:
I just noticed that there was the same TIP above; you are right, let’s go with TIP! 👍

/lgtm
/approve

PR #499, opened 2021-04-05
Add note using CA issuer as intermediate

erikgb:
Add documentation (note) on how to use CA issuer as intermediate. This was already missing from the docs, but the note is now reflecting the behavior suggested in https://github.com/jetstack/cert-manager/pull/3865.

English is not my native language, so feel free to suggest wording improvements... 😄 And used PKI terms may also be scrutinized.

Blocked by https://github.com/jetstack/cert-manager/pull/3865

/hold


erikgb:
@uncc @munnerz 
@cc @SgtCoDFish 

It seems like @munnerz is not very active atm, but feel free to take a look. 😄 Hoping to get some feedback on the PKI wording from @SgtCoDFish.

content/en/docs/configuration/ca.md
```diff
@@ -42,6 +42,12 @@ data:
...
 ```
+> Note: In case your CA issuer represents an intermediate, ensure that `tls.crt`
+> contains the issuing CA certificate _bundle_, in the correct order of signing:
+> `issuing -> intermediate(s) -> <root>`. The root (self-signed) CA certificate is
+> optional, but adding it will ensure that the correct CA certificate is stored in 
+> the `Certificate` secret with key `ca.crt`. When using CA issuer, cert-manager will
+> put the last certificate in the certificate chain configured into client `ca.crt`.
```
SgtCoDFish, line 50:
```suggestion

> Note: If your issuer represents an intermediate, ensure that `tls.crt` contains
> the issuer's full chain in the correct order: `issuer -> intermediate(s) -> root`.
> The root (self-signed) CA certificate is optional, but adding it will ensure that
> the correct CA certificate is stored in the secrets for issued `Certificate`s under
> the `ca.crt` key.
```

Does this wording work? (I also added an extra line break)

SgtCoDFish (commented):
I tweaked the wording slightly; what do you think to it? Anything missing?

content/en/docs/configuration/ca.md
erikgb: Works for me, but I wonder if anyone would like to know what happens if they don't include the root? Strictly speaking, the only mandatory in issuer `tls.crt` is the issuing CA certificate?

SgtCoDFish:
```suggestion

> Note: If your issuer represents an intermediate, ensure that `tls.crt` contains
> the issuer's full chain in the correct order: `issuer -> intermediate(s) -> root`.
> The root (self-signed) CA certificate is optional, but adding it will ensure that
> the correct CA certificate is stored in the secrets for issued `Certificate`s under
> the `ca.crt` key. If you fail to provide a complete chain, it might not be possible
> for consumers of issued `Certificate`s to verify whether they're trusted.
```

fair point; how about this?

erikgb: Better! Thanks!

PR #505, opened 2021-04-07
add SECURITY.md pointing at main cert-manager repo

SgtCoDFish:
the intent is to preserve a single source of truth for the reporting
process, and as such the file in this repo just points to the main repo

also adds a page to the "contributing" section of the site, with
similar information.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,6 @@
+# Vulnerability Reporting Process
+
+Security is the number one priority for cert-manager. If you think you've
+found a vulnerability in the cert-manager website, or in any cert-manager
+project, please follow the [vulnerability reporting process](https://github.com/jetstack/cert-manager/blob/master/SECURITY.md)
+documented in the main cert-manager repository.
--- a/content/en/docs/contributing/_index.md
+++ b/content/en/docs/contributing/_index.md
@@ -16,8 +16,9 @@ external cert-manager components:
 - [DCO sign off](./sign-off/)
 - [Developing with kind](./kind/)
 - [Running end-to-end tests](./e2e/)
-- [Release process](release-process/)
-- [Feature policy](policy/)
+- [Release process](./release-process/)
+- [Feature policy](./policy/)
+- [Reporting security vulnerabilities](./security/)
 
 ## Welcome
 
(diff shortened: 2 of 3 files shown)


irbekrm (commented):
Looks good to me!

Only thing, should we add it to the index here https://deploy-preview-505--cert-manager.netlify.app/docs/contributing/ in the main page?

SgtCoDFish:
> Only thing, should we add it to the index here https://deploy-preview-505--cert-manager.netlify.app/docs/contributing/ in the main page?

yep, that's a good point. I (foolishly) assumed all the indexes were auto-generated, like the links at the bottom of that page are. I'll add it in.

irbekrm:
> I (foolishly) assumed all the indexes were auto-generated, like the links at the bottom of that page are

I only knew to check for this because I had made the same assumption with a previous PR of my own 😅 

Wonder if that can be automated somehow.

SgtCoDFish:
I'm guessing something's complaining because the `release-next` notes are being worked on; I'll come back to this PR tomorrow or Friday

micahhausler:
What do you think about a note in GitHub Issue templates? Something to the effect of "If you think you've found a security vulnerability, please see X"?

SgtCoDFish:
> What do you think about a note in GitHub Issue templates? Something to the effect of "If you think you've found a security vulnerability, please see X"?

I was actually pretty surprised to see GitHub do that automatically based on `SECURITY.md`

E.g. in https://github.com/jetstack/cert-manager/issues when I hit "New Issue" I'm given the option to "Report a security vulnerability", which links to a page displaying `SECURITY.md`

What would adding a note to the templates catch? I guess if someone went directly to https://github.com/jetstack/cert-manager/issues/new ?

SgtCoDFish:
@irbekrm the last forcepush was a manual rebase and shouldn't have changed anything.

could I get another LGTM?

(I can add a note to github issue templates in a separate PR if we want them)

irbekrm:
Looks good to me!

> I was actually pretty surprised to see GitHub do that automatically based on SECURITY.md

This also looked good. I always create issues using the 'new issue' button- not sure whether there is another flow that people follow that we need to consider? As you say, perhaps that could go into another PR if needed.

/lgtm

PR #508, opened 2021-04-07
bug: Updated Hugo version

edeediong:
Signed-off-by: Edidiong Etuk <edidiongmichael08@gmail.com>

Linked to issue #507


maelvls:
**To the attention of the reviewer:** we want to make sure that we are not breaking any anchor id in the upgrade process. As an example, the anchor `#but-i-like-makefiles-more` should still work:

https://cert-manager.io/docs/contributing/building/#but-i-like-makefiles-more

maelvls:
Done in #628! Closing now.

/close

PR #437, opened 2021-02-16
Updates installation guide to use 1.2

JoshVanL:
Updates supported Kubernetes version to be 1.16
Removes all reference to the legacy manifests

Signed-off-by: joshvanl <vleeuwenjoshua@gmail.com>

PR #509, opened 2021-04-08
Bump versions

irbekrm:
Bumps versions in installation instructions.

Not sure if I am PR-ing against the right branch?


wallrj (approved):
/lgtm

Seems to cover all the cases that we changed during the last release in https://github.com/cert-manager/website/pull/437/files

PR #515, opened 2021-04-08
Adds note to external issuer contributions to wait for approval

JoshVanL:

/assign @wallrj

### DIFF
--- a/content/en/docs/contributing/external-issuers.md
+++ b/content/en/docs/contributing/external-issuers.md
@@ -36,6 +36,11 @@ corresponding issuer resource exists in Kubernetes. If these are both true, it
 will then use the information inside that issuer resource to attempt to create a
 signed certificate, based upon the certificate request.
 
+Issuers must also ensure that before signing, the `CertificateRequest` is
+[`Approved`](../../concepts/certificaterequest/#approval). If the
+`CertificateRequest` is not `Approved`, the issuer should not process it.
+Issuers are not responsible for approving `CertificateRequests`.
+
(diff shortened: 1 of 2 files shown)


wallrj (changes requested):
A couple of things:

1. What is our advice to external issuer authors who want to support cert-manager < 1.3?
2. Shall we add an extra note to the 1.3 release notes > breaking changes, to say that cert-manager is likely to be incompatible with external issuers until they are updated to support the approval conditions....or actually, is that true....because I suppose existing external issuers will simply ignore the approved condition and sign anyway.....but will that break the approval controller in cert-manager?

content/en/docs/contributing/external-issuers.md
```diff
@@ -36,6 +36,15 @@ corresponding issuer resource exists in Kubernetes. If these are both true, it
...
 signed certificate, based upon the certificate request.
 
+Issuers must also ensure that before signing, the `CertificateRequest` is
+[`Approved`](../../concepts/certificaterequest/#approval). If the
+`CertificateRequest` is not `Approved`, the issuer should not process it.
+Issuers are not responsible for approving `CertificateRequests`. Issuer
+implementations may choose to support older versions of cert-manager (pre
+`v1.3`) by signing `CertificateRequests` that will never have an approval
+condition set.  Disabling this gate check should be an optional flag, and the
+gate **enabled by default**.
```
wallrj, line 46: So perhaps we should update the sample-external-issuer to demonstrate how we'd implement this gate.

wallrj (approved):
/lgtm

PR #520, opened 2021-04-12
Release process: RELEASE_NAME is polluting the shell, Helm might be picking it up

maelvls:
Helm also uses RELEASE_NAME, which means that you might accidentally create a Helm release with the value of this variable if you run Helm right after doing the release process. @wallrj encountered [this exact issue](https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1617902959336100).


wallrj (approved):
/lgtm
/approve

PR #521, opened 2021-04-12
Talk about key rotation downtime and "rotationPolicy: Always"

maelvls:
Questions around possible downtime while rotating the private key secret were asked multiple times on the Kubernetes Slack: [1](https://kubernetes.slack.com/archives/C4NV3DWUC/p1618148597265800), [2](https://kubernetes.slack.com/archives/C4NV3DWUC/p1581109605050100). This PR is about mentioning that there is no expected downtime.

| Preview:  https://deploy-preview-521--cert-manager.netlify.app/docs/usage/certificate/#rotation-private-key |
|-|

Related:
- https://github.com/jetstack/cert-manager/issues/2402 (PR introducing private key auto-rotation when cert renewed)

### DIFF
--- a/content/en/docs/usage/certificate.md
+++ b/content/en/docs/usage/certificate.md
@@ -108,9 +108,8 @@ cert-manager supports requesting certificates that have a number of [custom key
 usages](https://tools.ietf.org/html/rfc5280#section-4.2.1.3) and [extended key
 usages](https://tools.ietf.org/html/rfc5280#section-4.2.1.12). Although
 cert-manager will attempt to honor this request, some issuers will remove, add
-defaults, or otherwise completely ignore the request and is determined on an
-issuer by issuer basis. The `CA` and `SelfSigned` `Issuer` will always return
-certificates matching the usages you have requested.
+defaults, or otherwise completely ignore the request.
+The `CA` and `SelfSigned` `Issuer` will always return certificates matching the usages you have requested.
 
 Unless any number of usages has been set, cert-manager will set the default
 requested usages of `digital signature`, `key encipherment`, and `server auth`.
@@ -120,7 +119,7 @@ certificate does not match the current key usage set.
 An exhaustive list of supported key usages can be found in the [API reference
 documentation](../../reference/api-docs/#cert-manager.io/v1alpha2.KeyUsage).
 
-## Temporary Certificates while Issuing
+## Temporary Certificates while Issuing {#temporary-certificates-whilst-issuing}
 
 On old GKE versions (`1.10.7-gke.1` and below), when requesting certificates
 [using the ingress-shim](../ingress/) alongside the
@@ -139,27 +138,44 @@ to your Ingress objects:
 cert-manager.io/issue-temporary-certificate": "true"
 ```
 
-That made sure that a temporary self-signed certificate is present in the
-`Secret`. The self-signed certificate is replaced with the properly signed
-certificate later on.
+That made sure that a temporary self-signed certificate was present in the
+`Secret`. The self-signed certificate was replaced with the signed certificate
+later on.
 
 ## Rotation of the private key {#rotation-private-key}
 
 By default, the private key won't be rotated automatically. Using the setting
-`rotationPolicy: Always`, the private key secret associated with a certificate
+`rotationPolicy: Always`, the private key Secret associated with a Certificate
 object can be configured to be rotated as soon as an action triggers the
-reissuance of the certificate object (see
-[actions](#actions-triggering-private-key-rotation) below).
+reissuance of the Certificate object (see
+[Actions that will trigger a rotation of the private key](#actions-triggering-private-key-rotation) below).
 
-With this setting, cert-manager waits until the certificate object is correctly
-signed before overwriting the private key secret, which means you can expect
-**no downtime** with regards to the private key secret.
+With `rotationPolicy: Always`, cert-manager waits until the Certificate
+object is correctly signed before overwriting the `tls.key` file in the
+Secret.
 
-Some Issuer types may disallow re-using private keys. If this is the case, you
-must explicitly configure the `rotationPolicy: Always` setting for each of your
-certificates objects accordingly.
+With this setting, you can expect **no downtime** if your application can
+detect changes to the mounted `tls.crt` and `tls.key` using and reload them
+gracefully or automatically restart.
 
-In the following example, the certificate has been set with `rotationPolicy: Always`:
+If your application only loads the private key and signed certificate once
+at start up, the new certificate won't immediately be served by your
+application, and you will want to either manually restart your pod with
+`kubectl rollout restart`, or automate the action by running
+[wave](https://github.com/wave-k8s/wave). Wave is a Secret controller that
+makes sure deployments get restarted whenever a mounted Secret changes.
+
+{{% alert title="Re-use of private keys" color="primary" %}}
+
+Some issuers, like the built-in [Venafi
+issuer](/docs/configuration/venafi/), may disallow re-using private keys.
+If this is the case, you must explicitly configure the `rotationPolicy:
+Always` setting for each of your Certificate objects accordingly.
+
+{{% /alert %}}
+
+In the following example, the certificate has been set with
+`rotationPolicy: Always`:
 
 ```yaml
 apiVersion: cert-manager.io/v1
@@ -172,12 +188,12 @@ spec:
 
 ### Actions that will trigger a rotation of the private key {#actions-triggering-private-key-rotation}
 
-Setting the `rotationPolicy: Always` won't do anything by itself. In order to
-get the private key secret rotated, the certificate objects must be reissued. A
-certificate object is reissued with either:
+Setting the `rotationPolicy: Always` or change the Certificate's spec won't
+rotate the private key immediately. In order to get the private key secret
+rotated, the certificate objects must be reissued. A certificate object is
+reissued with either:
 
 - when the X.509 certificate is nearing expiry,
-- when a change to the certificate object's spec is made,
 - when a reissuance is manually triggered with the following:
   ```sh
   kubectl cert-manager renew cert-1
@@ -185,14 +201,13 @@ certificate object is reissued with either:
   Note that the above command requires the [kubectl
   cert-manager](/docs/usage/kubectl-plugin/#renew) plugin.
 
-
 {{% pageinfo color="warning" %}}
 
-**🛑** Deleting the Secret resource associated a the Certificate resource is
+**❌** Deleting the Secret resource associated with a Certificate resource is
 **not a recommended solution** for manually rotating the private key. The
-recommended way to manually rotate the private key is to trigger a reissuance of
-the Certificate resource with the following command (requires the [kubectl
-cert-manager](/docs/usage/kubectl-plugin/#renew) plugin):
+recommended way to manually rotate the private key is to trigger the reissuance
+of the Certificate resource with the following command (requires the [`kubectl
+cert-manager`](/docs/usage/kubectl-plugin/#renew) plugin):
 
 ```sh
 kubectl cert-manager renew cert-1
@@ -211,7 +226,7 @@ The possible values for `rotationPolicy` are:
 
 With `rotationPolicy: Never`, a private key is only generated if one does not
 already exist in the target Secret resource (using the `tls.key` key). All
-further issuance's will re-use this private key. This is the default in order to
+further issuances will re-use this private key. This is the default in order to
 maintain compatibility with previous releases.
 
 With `rotationPolicy: Always`, a new private key will be generated each time an
--- a/content/en/docs/usage/certificate.md
+++ b/content/en/docs/usage/certificate.md
@@ -126,15 +126,13 @@ On old GKE versions (`1.10.7-gke.1` and below), when requesting certificates
 [`ingress-gce`](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress)
 ingress controller, `ingress-gce`
 [required](https://github.com/kubernetes/ingress-gce/pull/388) a temporary
-certificate is to be present while waiting for the issuance of a signed
+certificate must be present while waiting for the issuance of a signed
 certificate. Note that this issue was
 [solved](https://github.com/jetstack/cert-manager/issues/606#issuecomment-424397233)
 in `1.10.7-gke.2`.
 
-To work around this, you had to add the following annotation
-to your Ingress objects:
-
 ```yaml
+# Required for GKE 1.10.7-gke.1 and below.
 cert-manager.io/issue-temporary-certificate": "true"
 ```
 
@@ -154,9 +152,9 @@ With `rotationPolicy: Always`, cert-manager waits until the Certificate
 object is correctly signed before overwriting the `tls.key` file in the
 Secret.
 
-With this setting, you can expect **no downtime** if your application can
-detect changes to the mounted `tls.crt` and `tls.key` using and reload them
-gracefully or automatically restart.
+With this setting, you can expect **no downtime** if your application can detect
+changes to the mounted `tls.crt` and `tls.key` and reload them gracefully or
+automatically restart.
 
 If your application only loads the private key and signed certificate once
 at start up, the new certificate won't immediately be served by your
@@ -188,10 +186,10 @@ spec:
 
 ### Actions that will trigger a rotation of the private key {#actions-triggering-private-key-rotation}
 
-Setting the `rotationPolicy: Always` or change the Certificate's spec won't
-rotate the private key immediately. In order to get the private key secret
-rotated, the certificate objects must be reissued. A certificate object is
-reissued with either:
+Setting the `rotationPolicy: Always` or changing other fields in the Certificate
+spec won't rotate the private key immediately. In order to rotate the private
+key, the certificate objects must be reissued. A certificate object is reissued
+under the following circumstances:
 
 - when the X.509 certificate is nearing expiry,
 - when a reissuance is manually triggered with the following:
@@ -230,15 +228,15 @@ further issuances will re-use this private key. This is the default in order to
 maintain compatibility with previous releases.
 
 With `rotationPolicy: Always`, a new private key will be generated each time an
-action triggers the reissuance of the certificate object (see
-[actions](#actions-triggering-private-key-rotation) above). Note that if the
-private key secret already exists when creating the certificate object, the
-existing private key will not be used, since the rotation mechanism also
-includes the initial issuance.
+action triggers the reissuance of the certificate object (see [Actions that will
+trigger a rotation of the private key](#actions-triggering-private-key-rotation)
+above). Note that if the private key secret already exists when creating the
+certificate object, the existing private key will not be used, since the
+rotation mechanism also includes the initial issuance.
 
 {{% pageinfo color="info" %}}
 
-👉 It is recommended you configure `rotationPolicy: Always` on your certificate
+👉 We recommend that you configure `rotationPolicy: Always` on your certificate
 resources. It is considered to be a good practice to rotate private keys when a
 certificate is renewed.
 
--- a/content/en/docs/usage/certificate.md
+++ b/content/en/docs/usage/certificate.md
@@ -13,7 +13,6 @@ you will interact with cert-manager to request signed certificates.
 In order to issue any certificates, you'll need to configure an
 [`Issuer`](../../configuration/) resource first.
 
-
 ## Creating Certificate Resources
 
 A `Certificate` resource specifies fields that are used to generate certificate
@@ -40,7 +39,7 @@ spec:
   renewBefore: 360h # 15d
   subject:
     organizations:
-    - jetstack
+      - jetstack
   # The use of the common name field has been deprecated since 2000 and is
   # discouraged from being used.
   commonName: example.com
@@ -54,12 +53,12 @@ spec:
     - client auth
   # At least one of a DNS Name, URI, or IP address is required.
   dnsNames:
-  - example.com
-  - www.example.com
+    - example.com
+    - www.example.com
   uris:
-  - spiffe://cluster.local/ns/sandbox/sa/example
+    - spiffe://cluster.local/ns/sandbox/sa/example
   ipAddresses:
-  - 192.168.0.5
+    - 192.168.0.5
   # Issuer references are always required.
   issuerRef:
     name: ca-issuer
@@ -79,7 +78,7 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
 `sandbox` namespace (the same namespace as the `Certificate` resource).
 
 > Note: If you want to create an `Issuer` that can be referenced by
-> `Certificate` resources in *all* namespaces, you should create a
+> `Certificate` resources in _all_ namespaces, you should create a
 > [`ClusterIssuer`](../../concepts/issuer/#namespaces) resource and set the
 > `certificate.spec.issuerRef.kind` field to `ClusterIssuer`.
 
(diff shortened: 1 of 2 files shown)


maelvls:
I needed to use `{id="key-usages"}` to avoid breaking people's links anchored to the "Key Usages" section. But our old version of `hugo` does not support setting custom `id` anchors 😞 

old-hugo

As discussed in #507, upgrading to the latest Hugo version isn't trivial either 😞 

I'll think of a solution.

content/en/docs/usage/certificate.md
```diff
@@ -97,79 +96,130 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
+  cert-manager](/docs/usage/kubectl-plugin/#renew) plugin.
+
+
+{{% pageinfo color="warning" %}}
+
+**🛑** Deleting the Secret resource associated a the Certificate resource is
+**not a recommended solution** for manually rotating the private key. The
+recommended way to manually rotate the private key is to trigger a reissuance of
+the Certificate resource with the following command (requires the [kubectl
+cert-manager](/docs/usage/kubectl-plugin/#renew) plugin):
```
wallrj, line 187:
```suggestion
**🛑** Deleting the Secret resource associated a Certificate resource is
**not a recommended solution** for manually rotating the private key. The
recommended way to manually rotate the private key is to trigger a reissuance of
the Certificate resource with the following command (requires the [kubectl
cert-manager](/docs/usage/kubectl-plugin/#renew) plugin):
```

content/en/docs/usage/certificate.md
```diff
@@ -97,79 +96,130 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
+
+| Value                  | Description                                                   |
+| ---------------------- | ------------------------------------------------------------- |
+| `Never` (default)      | cert-manager reuses the existing private key on each issuance |
+| `Always` (recommended) | cert-manager regenerates a new private key on each issuance   |
+
+With `rotationPolicy: Never`, a private key is only generated if one does not
+already exist in the target Secret resource (using the `tls.key` key). All
+further issuance's will re-use this private key. This is the default in order to
+maintain compatibility with previous releases.
```
wallrj, line 207:
```suggestion
With `rotationPolicy: Never`, a private key is only generated if one does not
already exist in the target Secret resource (using the `tls.key` key). All
further issuances will re-use this private key. This is the default in order to
maintain compatibility with previous releases.
```

wallrj (changes requested):
Thanks @maelvls 

I spotted a few typos and made a few suggestions.
Also suggest going through all the changes in the PR and ensuring that references to cert-manager or Kubernetes resource types are capitalized. E.g. Certificate, Secret etc....I didn't highlight them all below.

maelvls:
TODO: add a note that says "changing Never to Always does not immediately trigger a rotation of the private key.

content/en/docs/usage/certificate.md
```diff
@@ -97,79 +96,130 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
 An exhaustive list of supported key usages can be found in the [API reference
 documentation](../../reference/api-docs/#cert-manager.io/v1alpha2.KeyUsage).
 
-## Temporary Certificates whilst Issuing
+## Temporary Certificates while Issuing
+
+When requesting certificates [using the ingress-shim](../ingress/), the
+component `ingress-gce`, if used, requires that a temporary certificate is
+present while waiting for the issuance of a signed certificate when serving. To
+facilitate this, if the following annotation:
```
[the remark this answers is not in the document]
maelvls: I found two places where this "trick" is mentioned:

- the actual PR where this annotation was added: https://github.com/jetstack/cert-manager/pull/2089,
- In the [release notes for 0.11](https://github.com/jetstack/cert-manager/blob/dffbf391dbb0fc6c1cfea62e561a9c6f54362ab0/design/release-notes/release-0.11/draft-release-notes.md#no-longer-generating-temporary-certificates-by-default):

  > **New CertificateRequest resource type**
  >
  > The activation of `CertificateRequest` controllers are no longer behind a feature flag and are now instead enabled by default.
  >
  > This means that when requesting certificates using the `Certificate` resource the `CertificateRequest` resource will be used as the default and only way to honour the request.
  >
  > The addition of this resource introduces the ability for much greater extension points to cert-manager, notably out-of-tree issuers, istio integrations, and experimental tooling such as a CSI driver. You can read more about the motivation and design of this resource in the [enhancement document](https://github.com/jetstack/cert-manager/blob/master/design/20190708.certificate-request-crd.md).
  >
  > This change should cause no disruption to how end users interact with cert-manager, with the exception of debugging now requiring this resource to be inspected also.

- But by far the best explanation has been written by James Munnelly in https://github.com/kubernetes/ingress-gce/pull/388:

  > This switches ingress-gce back to the previous behaviour of 'soft failing' if the specified TLS secret exists. Without this change, we've had many users complain that auto-TLS using cert-manager (and also kube-lego) is broken.

I'll go ahead and add a link to this ingress-gce PR.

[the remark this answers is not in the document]
maelvls: I now use the past tense since this seems to be a solved issue:

> On old GKE versions (1.10.7-gke.1 and below), when requesting certificates [using the ingress-shim](../ingress/) alongside the [`ingress-gce`](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress) ingress controller, `ingress-gce` [required](https://github.com/kubernetes/ingress-gce/pull/388) a temporary certificate is to be present while waiting for the issuance of a signed certificate. Note that this issue was solved [in 1.10.7-gke.2](https://github.com/jetstack/cert-manager/issues/606#issuecomment-424397233).
> 
> To work around this, you had to add the following annotation to your Ingress objects:
>
> ```yaml
> cert-manager.io/issue-temporary-certificate": "true"
> ```
>
> That made sure that a temporary self-signed certificate is present in the `Secret`. The self-signed certificate is replaced with the properly signed certificate later on.

@munnerz I am correct in saying that this annotation is not needed anymore?

content/en/docs/usage/certificate.md
```diff
@@ -97,79 +96,130 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
+reissuance of the certificate object (see
+[actions](#actions-triggering-private-key-rotation) below).
 
-> **WARNING**: This feature requires enabling the `ExperimentalCertificateControllers`
-> feature gate by passing the `--feature-gates=ExperimentalCertificateControllers=true`
-> flag to the controller component, or adding `--set featureGates=ExperimentalCertificateControllers=true`
-> when deploying using the Helm chart.
+With this setting, cert-manager waits until the certificate object is correctly
+signed before overwriting the private key secret, which means you can expect
+**no downtime** with regards to the private key secret.
```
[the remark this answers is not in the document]
maelvls: Good point, I totally forgot about the fact that "no downtime" also implies that any application using the secret immediately uses the new certificate...

I added your suggestion in 62a8bf15108e7f21c0dd675f268c5575cd70b738. Thank you!!

content/en/docs/usage/certificate.md
```diff
@@ -97,79 +96,130 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
 
-When a certificate is re-issued for any reason, including because it is nearing
-expiry, when a change to the spec is made or a re-issuance is manually
-triggered, cert-manager supports configuring the 'private key rotation policy'
-to either always re-use the existing private key (the default behavior) or to
-regenerate a new private key on each issuance (the recommended behavior).
+Some Issuer types may disallow re-using private keys. If this is the case, you
+must explicitly configure the `rotationPolicy: Always` setting for each of your
+certificates objects accordingly.
```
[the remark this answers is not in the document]
maelvls: I have been looking around (searched "reuse private key" on GitHub Search) but I wasn't able to find anything.

I found out about [ReusePrivateKey](https://github.com/jetstack/cert-manager/blob/2aa625edcb1a179600d4c557216c7bfeb44ae269/test/e2e/framework/helper/featureset/featureset.go#L106):

```go
	// ReusePrivateKey denotes whether the target issuer is able to sign multiple
	// certificates for the same private key. This is useful for some issuers
	// that have trouble being configured to support this feature.
	ReusePrivateKeyFeature Feature = "ReusePrivateKey"
```

After some git wrangling, I found the commit https://github.com/jetstack/cert-manager/commit/a96f93e0c9bc9b38fa1804f557707dbe9f3b3418 that introduced the `ReusePrivateKey` flag. The commit message points towards the Venafi issuer, I'll add the venafi issuer as an example.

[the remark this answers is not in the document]
maelvls: Done in 9371683!

content/en/docs/usage/certificate.md
```diff
@@ -97,79 +96,130 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
+```
+
+### Actions that will trigger a rotation of the private key {#actions-triggering-private-key-rotation}
+
+Setting the `rotationPolicy: Always` won't do anything by itself. In order to
+get the private key secret rotated, the certificate objects must be reissued. A
+certificate object is reissued with either:
+
+- when the X.509 certificate is nearing expiry,
+- when a change to the certificate object's spec is made,
```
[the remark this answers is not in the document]
maelvls: Good point, I made it clearer that a change to the certificate does not trigger a renew of the private key in ee43f0c.

content/en/docs/usage/certificate.md
```diff
@@ -97,79 +96,152 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
+requested usages of `digital signature`, `key encipherment`, and `server auth`.
 cert-manager will not attempt to request a new certificate if the current
-certificate does not match the current key usages set.
+certificate does not match the current key usage set.
 
 An exhaustive list of supported key usages can be found in the [API reference
 documentation](../../reference/api-docs/#cert-manager.io/v1alpha2.KeyUsage).
 
-## Temporary Certificates whilst Issuing
+## Temporary Certificates while Issuing
```
maelvls, line 122: I just realized that this change is a breaking change for anyone who use the anchor ID, e.g.

> https://cert-manager.io/docs/usage/certificate/#temporary-certificates-whilst-issuing

Should I use instead:

```md
## Temporary Certificates while Issuing {#temporary-certificates-whilst-issuing}
```

to avoid breaking the anchor ID?

maelvls:
@irbekrm I am stuck with a missing link (more specifically, the anchor ID `#renewal`) that only exists in `/docs-next`:

```
docs/release-notes/release-notes-1.4/index.html
  hash does not exist --- docs/release-notes/release-notes-1.4/index.html --> ../../usage/certificate/#renewal
```

Should I merge my change to `release-next` instead of `master`?

content/en/docs/usage/certificate.md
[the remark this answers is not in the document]
maelvls: I will resolve this conversation in order to move this PR forward. Please re-open it if you think I missed something.

content/en/docs/usage/certificate.md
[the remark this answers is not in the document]
maelvls: I will resolve this conversation in order to move this PR forward. Please re-open it if you think I missed something.

maelvls:
Important note: I removed an anchor id in this PR: `#configuring-private-key-rotation` 

(https://deploy-preview-521--cert-manager.netlify.app/docs/usage/certificate/#configuring-private-key-rotation)


<details>
<summary>How to find breaking changes in anchors: "diff-anchor"</summary>


Running `diff-anchors $(git merge-base origin/master @^) @`:

```diff
--- a/public/before/docs/usage/certificate/index.html
+++ b/public/after/docs/usage/certificate/index.html
@@ -1,6 +1,8 @@
 <h2 id="cleaning-up-secrets-when-certificates-are-deleted"
-<h2 id="configuring-private-key-rotation"
 <h2 id="creating-certificate-resources"
 <h2 id="key-usages"
 <h2 id="renewal"
+<h2 id="rotation-private-key"
 <h2 id="temporary-certificates-whilst-issuing"
+<h3 id="actions-triggering-private-key-rotation"
+<h3 id="the-rotationpolicy-setting"
```


To install `diff-anchors`:

```sh
cat <<'EOF' | tee ~/bin/diff-anchors && chmod +x ~/bin/diff-anchors
#! /bin/bash

#
# We want to avoid breaking the anchor ids. An anchor id is, for example:
#  <h3 id="configuring-cert-manager">Configuring cert-manager</h3>
#          <---------------------->
#                anchor id
#
# To check if you have removed an anchor id in your PR, you can run:
#
#   diff-anchors $(git merge-base origin/master @^) @

set -xeuo pipefail

# https://stackoverflow.com/a/2173421/3808537
trap "git checkout $(git branch --show-current)" SIGINT SIGTERM EXIT

BEFORE=$(git describe --always --tags "$1")
AFTER=$(git describe --always --tags "$2")

git -c advice.detachedHead=false checkout "$BEFORE"
git submodule update --init --recursive
hugo --destination=public/before --quiet

git -c advice.detachedHead=false checkout "$AFTER"
git submodule update --init --recursive
hugo --destination=public/after --quiet

set +x

find public/before -type f -regex '.*/index.html' \
  | sed s@^public/before@@ \
  | grep -Ev '^/(v|next)' \
  | while read -r p; do
    grep --only-matching '<h[0-9] id="[^"]*"' public/before"$p" | sort >public/before"$p"_ && mv public/before"$p"{_,}
    grep --only-matching '<h[0-9] id="[^"]*"' public/after"$p" | sort >public/after"$p"_ && mv public/after"$p"{_,}
    git diff --no-index public/before"$p" public/after"$p"
  done
EOF
```

</details>

JoshVanL:
This good to me, reads really well :)

/lgtm

content/en/docs/usage/certificate.md
```diff
@@ -91,71 +90,159 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
+
+On old GKE versions (`1.10.7-gke.1` and below), when requesting certificates
+[using the ingress-shim](../ingress/) alongside the
+[`ingress-gce`](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress)
+ingress controller, `ingress-gce`
+[required](https://github.com/kubernetes/ingress-gce/pull/388) a temporary
+certificate is to be present while waiting for the issuance of a signed
+certificate. Note that this issue was
+[solved](https://github.com/jetstack/cert-manager/issues/606#issuecomment-424397233)
+in `1.10.7-gke.2`.
```
wallrj, line 132:
```suggestion
On old GKE versions (`1.10.7-gke.1` and below), when requesting certificates
[using the ingress-shim](../ingress/) alongside the
[`ingress-gce`](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress)
ingress controller, `ingress-gce`
[required](https://github.com/kubernetes/ingress-gce/pull/388) a temporary
certificate must be present while waiting for the issuance of a signed
certificate. Note that this issue was
[solved](https://github.com/jetstack/cert-manager/issues/606#issuecomment-424397233)
in `1.10.7-gke.2`.
```

content/en/docs/usage/certificate.md
```diff
@@ -91,71 +90,159 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
-When [requesting certificates using ingress-shim](../ingress/), the component
-`ingress-gce`, if used, requires that a temporary certificate is present while
-waiting for issuance of a signed certificate when serving. To facilitate this,
-if the annotation `"cert-manager.io/issue-temporary-certificate": "true"` is
-present on the certificate, a self signed temporary certificate will be present
-on the `Secret` until it is overwritten once the signed certificate has been
-issued.
+With this setting, you can expect **no downtime** if your application can
+detect changes to the mounted `tls.crt` and `tls.key` using and reload them
+gracefully or automatically restart.
```
wallrj, line 159:
```suggestion
With this setting, you can expect **no downtime** if your application can
detect changes to the mounted `tls.crt` and `tls.key` and reload them
gracefully or automatically restart.
```

content/en/docs/usage/certificate.md
```diff
@@ -91,71 +90,159 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
-    rotationPolicy: Always
+    rotationPolicy: Always # 🔰 Here.
+```
+
+### Actions that will trigger a rotation of the private key {#actions-triggering-private-key-rotation}
+
+Setting the `rotationPolicy: Always` or change the Certificate's spec won't
+rotate the private key immediately. In order to get the private key secret
+rotated, the certificate objects must be reissued. A certificate object is
+reissued with either:
```
wallrj, line 194:
```suggestion
Setting the `rotationPolicy: Always` or changing other fields in the Certificate spec won't
rotate the private key immediately. In order to rotate the private key, the certificate objects must be reissued. A certificate object is
reissued under the following circumstances:
```

content/en/docs/usage/certificate.md
```diff
@@ -91,71 +90,159 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
+already exist in the target Secret resource (using the `tls.key` key). All
+further issuances will re-use this private key. This is the default in order to
+maintain compatibility with previous releases.
+
+With `rotationPolicy: Always`, a new private key will be generated each time an
+action triggers the reissuance of the certificate object (see
+[actions](#actions-triggering-private-key-rotation) above). Note that if the
+private key secret already exists when creating the certificate object, the
+existing private key will not be used, since the rotation mechanism also
+includes the initial issuance.
```
wallrj, line 237:
```suggestion
With `rotationPolicy: Always`, a new private key will be generated each time an
action triggers the reissuance of the certificate object (see
[Actions that will trigger a rotation of the private key](#actions-triggering-private-key-rotation) above). Note that if the
private key secret already exists when creating the certificate object, the
existing private key will not be used, since the rotation mechanism also
includes the initial issuance.
```

content/en/docs/usage/certificate.md
```diff
@@ -91,71 +90,159 @@ The `Certificate` will be issued using the issuer named `ca-issuer` in the
...
 
-* **Never** (default): a private key is only generated if one does not already exist in
-  the target Secret resource (using the `tls.key` key). All further issuance's will re-use
-  this private key. This is the default in order to maintain compatibility with previous releases.
-* **Always**: a new private key will be generated each time a new certificate is issued.
-  It is recommended you configure this `rotationPolicy` on your Certificate resources as it
-  is good practice to rotate private keys when a certificate is renewed.
+👉 It is recommended you configure `rotationPolicy: Always` on your certificate
+resources. It is considered to be a good practice to rotate private keys when a
+certificate is renewed.
```
wallrj, line 243:
```suggestion
👉 We recommend that you configure `rotationPolicy: Always` on your certificate
resources. It is considered to be a good practice to rotate private keys when a
certificate is renewed.
```

Can we link to a reference backing up this claim good practice?

wallrj (changes requested):
I spotted a few more typos.

JoshVanL:
/hold
Since there are more review comments

content/en/docs/usage/certificate.md
maelvls: I cannot find any good reference to support this claim. By reading some discussions:

- https://community.letsencrypt.org/t/should-i-always-generate-a-new-csr-private-key/100985
- https://security.stackexchange.com/questions/27810/should-i-change-the-private-key-when-renewing-a-certificate

People seem to suggest that:
1. RSA/ECDSA private keys can't (mostly) be guessed, which means there is less of a need for renewing them (unlike passwords that are way shorter),
2. But at the same time, an RSA/ECDSA private keys may be leaked through some old backup, leading to a possible man-in-the-middle situation.

Looking around, I found that:

- certbot renews the private key while renewing the cert by default, but there is nothing explaining why;
- NIST's [Guidelines for the Selection, Configuration, and Use of Transport Layer Security (TLS) Implementations](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-52r2.pdf) does not indicate when to renew the private key; the only bit that talks about what to do with a server private key is this:
  > The security of the server’s private key is critical to the security of TLS. If the server’s private key is weak or can be obtained by a third party, the third party can masquerade as the server to
all clients.
- NIST's [Recommendation for Key Management](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf) has a nice table but no claim on when the private key should be renewweed:
  image

- f5-bigip suggests to renew the private key (https://support.f5.com/csp/article/K14620):

  > Note: When renewing an SSL certificate from a CA, F5 recommends that you generate a new certificate signing request (CSR) and private key. Although some CAs allow you to renew a certificate by using the existing CSR on file, this method is less secure as it retains the existing private key. To generate a new CSR, follow the procedure for Creating an SSL CSR and private key.


@SgtCoDFish Do you know a place where that would be written? Or should I rather remove this claim and say that "the cert-manager team suggests users renewing the private key along with the certificate"?

munnerz: I think one big reason why you _would_ want to, is that if you were to renew a certificate using the same private key, and then after the initial certificates expiry learn that the private key had actually been leaked, the new certificate would be valid to use with the old private key, thus the 'rotation' didn't really provide any more protection and really just extended the lifetime of the existing certificate.

content/en/docs/usage/certificate.md
munnerz: Also:

> Changing the Certificate spec does not trigger the renewal process,

This statement in isolation is not true, and may cause confusion. The *majority* of fields *will* cause a re-issuance (e.g. changing dnsNames, commonName, usages etc). Changing the `rotationPolicy` (or any other fields that control how a renewal happens) *won't* cause a re-issuance immediately.

content/en/docs/usage/certificate.md
SgtCoDFish: James is spot on. I don't have like an authoritative source to hand for why you'd want to rotate as much as possible, but certainly I think it's best practice. 

The longer a key is used, the more likely it is to be exposed. Keys should be renewed "as often as possible" for me; not only for security reasons but also operational reasons to show that it can be done, so that if it has to be done in an emergency the org has confidence that they can carry it out.

maelvls: I tried my best to formulate a paragraph that "justifies" our point of view. Please forgive me if any of this is untrue. 🔥

👉 We recommend that you configure `rotationPolicy: Always` on your Certificate resources. 

Rotating both the certificate and the private key simultaneously prevents the risk of issuing a certificate associated with an exposed private key.

Another benefit to renewing the private key regularly is to let you be confident that the private key rotation can be done in case of an emergency.

More generally, it is a good practice to be rotating the keys as often as possible, reducing the risk associated with compromised keys.

wallrj (approved):
/lgtm

content/en/docs/usage/certificate.md
maelvls: In https://github.com/cert-manager/website/pull/521#discussion_r627557993, Richard proposed to remove the line:

> The certificate object is reissued when a change to the certificate object's spec is made.

I think I misunderstood Richard's suggestion back in May. I should have understood Richard's comment as:

> The certificate object is reissued when a change to one of the certificate object's fields `commonName`, `dnsNames`, `ipAddresses`, `uris`, `emailAddresses`, `subject`, `isCA`, `usages`, `duration` or `issuerRef` is made.

(as per [`RequestMatchesSpec`](https://github.com/jetstack/cert-manager/blob/8ae179b8f5aec708f8cb320a40ca8b495c9d889e/pkg/controller/certificates/util.go#L118))

Apologies for my misunderstanding.

I re-added the line about reissuing on spec change in ea1f25b.

PR #523, opened 2021-04-13
Add style guide

SgtCoDFish:
Related to https://github.com/jetstack/cert-manager/issues/3868

Note that I also have similar changes ready to go for the README in jetstack/cert-manager, but they'll link to this style guide on the website and so I'm waiting for this to be merged before I raise the other PR

### DIFF
--- a/content/en/docs/contributing/_index.md
+++ b/content/en/docs/contributing/_index.md
@@ -18,6 +18,7 @@ external cert-manager components:
 - [Running end-to-end tests](./e2e/)
 - [Release process](./release-process/)
 - [Feature policy](./policy/)
+- [Coding Conventions](./coding-conventions/)
 - [Reporting security vulnerabilities](./security/)
 
 ## Welcome
@@ -35,14 +36,12 @@ To get invites you can subscribe to [our mailing list](https://groups.google.com
 
 We have 2 regular repeating meetings:
 
-* our daily stand-up meetings [on Google Meet](
-https://meet.google.com/eum-fyvt-xpa) at 10:30 London time every weekday 
-* our bi-weekly developer meetings [are on Google Meet](https://meet.google.com/abp-bwhk-wxc) 17:00 London time every 2 weeks (check calendar invites or [meeting notes](https://docs.google.com/document/d/1Tc5t6ylY9dhXAan1OjOoldeaoys1Yh4Ir710ATfBa5U)). 
-
-You can check what time this is where you are using [`everytimezone.com`](https://everytimezone.com/).
+* daily stand-up meetings [on Google Meet](https://meet.google.com/eum-fyvt-xpa) at [10:30 London time](http://www.thetimezoneconverter.com/?t=10:30&tz=Europe/London) every weekday
+* bi-weekly developer meetings [on Google Meet](https://meet.google.com/abp-bwhk-wxc) at [17:00 London time](http://www.thetimezoneconverter.com/?t=17:00&tz=Europe/London) (for dates, check calendar invites or [meeting notes](https://docs.google.com/document/d/1Tc5t6ylY9dhXAan1OjOoldeaoys1Yh4Ir710ATfBa5U))
 
 ## Slack
+
 We have two cert-manager channels on [the Kubernetes Slack](https://slack.k8s.io) which we use to communicate:
 
-* `cert-manager` a general users channel, use this one for any usage related questions.
-* `cert-manager-dev` this is what the developers use to collaborate, please only use this for code related questions
+* `cert-manager`: a general channel for all users of cert-manager; use this one for any usage related questions.
+* `cert-manager-dev`: a channel for collaboration between cert-manager contributors and maintainers; please only use this for code related questions
--- a/content/en/docs/contributing/coding-conventions.md
+++ b/content/en/docs/contributing/coding-conventions.md
@@ -1,0 +1,52 @@
+---
+title: "Coding Conventions"
+linkTitle: "Coding Conventions"
+weight: 80
+type: "docs"
+---
+
+cert-manager, like most Go projects, delegates almost all stylistic choices to `gofmt`,
+with `goimports` on top for organizing imports. Broadly speaking, if you set your editor to run
+`goimports` when you save a file, your code will be stylistically correct.
+
+cert-manager generally also follows the Kubernetes
+[coding conventions](https://www.kubernetes.dev/docs/guide/coding-convention/) and the Google
+[Go code review comments](https://github.com/golang/go/wiki/CodeReviewComments).
+
+## Organizing Imports
+
+Imports should be organized into 3 blocks, with each block separated by two newlines:
+
+```go
+import (
+	"stdlib"
+
+	"external"
+
+	"internal"
+)
+```
+
+An example might be the following, taken from
+[`pkg/acme/accounts/client.go`](https://github.com/jetstack/cert-manager/blob/0c71fe7795858b96cabcddabf706d997cd2fba3f/pkg/acme/accounts/client.go):
+
+```go
+import (
+	"crypto/rsa"
+	"crypto/tls"
+	"net"
+	"net/http"
+	"time"
+
+	acmeapi "golang.org/x/crypto/acme"
+
+	acmecl "github.com/jetstack/cert-manager/pkg/acme/client"
+	acmeutil "github.com/jetstack/cert-manager/pkg/acme/util"
+	cmacme "github.com/jetstack/cert-manager/pkg/apis/acme/v1"
+	"github.com/jetstack/cert-manager/pkg/metrics"
+	"github.com/jetstack/cert-manager/pkg/util"
+)
+```
+
+Once this manual split of standard library, external and internal imports has been made, it will be
+enforced automatically by `goimports` when executed in the future.


content/en/docs/contributing/style-guide.md
```diff
@@ -0,0 +1,66 @@
...
+It's important to note that cert-manager `v1` doesn't generally provide any guarantees that it can be
+imported and used as a module by other programs, even for publicly exported functions, variables,
+constants, interfaces and types under `pkg/`.
+
+Only the following modules can be explicitly relied upon to maintain the go module compatibility
+guarantee in cert-manager `v1`:
+
+- `pkg/apis`
+- `pkg/api`
+- `pkg/dns/util`
```
SgtCoDFish, line 66: this list may need expanding; any suggestions welcome

content/en/docs/contributing/style-guide.md
```diff
@@ -0,0 +1,66 @@
...
+with `goimports` on top for organizing imports. Broadly speaking, if you set your editor to run
+`goimports` when you save a file, your code will be stylistically correct.
+
+cert-manager generally also follows the Kubernetes
+[coding conventions](https://www.kubernetes.dev/docs/guide/coding-convention/) and the Google
+[Go code review comments](https://github.com/golang/go/wiki/CodeReviewComments).
+
+## Organizing Imports
+
+Imports should be organized into 3 blocks, with each block separated by two newlines:
```
maelvls, line 18: Thanks for turning this "knowledge in the air" into a written sentence ❤️

content/en/docs/contributing/style-guide.md
```diff
@@ -0,0 +1,66 @@
...
+	cmacme "github.com/jetstack/cert-manager/pkg/apis/acme/v1"
+	"github.com/jetstack/cert-manager/pkg/metrics"
+	"github.com/jetstack/cert-manager/pkg/util"
+)
+```
+
+Once this manual split of standard library, external and internal imports has been made, it will be
+enforced automatically by `goimports` when executed in the future.
+
+## Go Module Compatibility
```
maelvls, line 55: I think we should also add that to the [README.md](https://github.com/jetstack/cert-manager/blob/master/README.md) (taking inspiration for our beloved [client-go](https://github.com/kubernetes/client-go#compatibility-your-code---client-go) of course 😆)

content/en/docs/contributing/style-guide.md
```diff
@@ -0,0 +1,66 @@
...
+
+It's important to note that cert-manager `v1` doesn't generally provide any guarantees that it can be
+imported and used as a module by other programs, even for publicly exported functions, variables,
+constants, interfaces and types under `pkg/`.
```
maelvls, line 59: I think we should also add this as a big red notice banner to our [README.md](https://github.com/jetstack/cert-manager/blob/master/README.md) (taking inspiration for our beloved [client-go](https://github.com/kubernetes/client-go#compatibility-your-code---client-go) of course 😆)

SgtCoDFish: I should have mentioned this in the PR notes, and I'll add it now.

I have a a few changes ready to go for cert-manager's README, but it'll link to these docs so it depends on this PR being merged first.

content/en/docs/contributing/style-guide.md
irbekrm: I do see folks importing [/pkg/issuer/acme/dns/util](https://pkg.go.dev/github.com/jetstack/cert-manager/pkg/issuer/acme/dns/util?tab=importedby) (which does not mean that we promise compatibility as such, but perhaps that we should)? Same with [pkg/util/pki](https://pkg.go.dev/github.com/jetstack/cert-manager/pkg/util/pki)

irbekrm (approved):
Looks good to me and great that we will have this written down.

I think it is great that we will now have this written down.
I am also wondering if we might want to (eventually) make this into a general good coding practices doc with more smaller detail of what coding practices we want to follow (i.e in regards to flaky tests, always use fake clock etc)- or if that's better put elsewhere.

irbekrm:
Adding an lgtm with hold- this looks good to me as such, I don't know if you want extra 👀 
/hold
/lgtm

jakexks (approved):
👍 turning tribal knowledge into documentation
/lgtm

content/en/docs/contributing/style-guide.md
jakexks: Our example [dns provider webhook](https://github.com/cert-manager/webhook-example/blob/1f895be0fe2643c03838b043ce1d622f4024f791/example/example.go#L10) imports `pkg/acme/webhook`. Not sure what we should do with that as we've actively encouraged importing it.

https://github.com/cert-manager/webhook-example/blob/1f895be0fe2643c03838b043ce1d622f4024f791/example/example.go#L10

SgtCoDFish: I'd be inclined to add both to the list, but then I'd be inclined to add everything under `pkg/` to the list so I would say that :grin: 

More seriously, I think anything which we know is currently being imported by open source projects should be on this list unless there's an exceptional reason. I don't like breaking people unless it's absolutely 100% necessary.

SgtCoDFish:
I've added a few paths to the list based on currently imported modules in:

- https://github.com/cert-manager/sample-external-issuer
- https://github.com/cert-manager/istio-csr/blob/master/pkg/server/server.go
- https://github.com/cert-manager/webhook-example

I also found [this](https://github.com/nuxeo-sandbox/k8s-policy-controller/blob/67cc36678cff7e7cc9f2c3d3de7c10e138a3a623/pkg/server/server.go#L16) import of code under `cmd/` which I didn't add, as a judgement call.

Again, the list isn't exhaustive and this has been a manual process.

content/en/docs/contributing/style-guide.md
```diff
@@ -0,0 +1,73 @@
...
+
+Only the following modules can be explicitly relied upon to maintain the go module compatibility
+guarantee in cert-manager `v1`:
+
+- `pkg/api`
+- `pkg/apis`
+- `pkg/acme/webhook`
+- `pkg/client`
```
munnerz, line 67: This should not be in this list - consumers of our repository should generate their own typed clients rather than depend on ours, as they can (and have) changed between k8s releases that we depend upon

content/en/docs/contributing/style-guide.md
```diff
@@ -0,0 +1,73 @@
...
+- `pkg/dns/util`
+- `pkg/issuer/acme/dns/util`
+- `pkg/util/pki`
```
munnerz, line 70: I don't think this should be on this list either - if there are specific functions we expect to be stable, can we separate them out? It's typically been quite messy in here 😬

content/en/docs/contributing/style-guide.md
munnerz: I take the opposite stance to be honest. Whilst I do agree we should aim towards everything being here, we've historically not performed reviews of code that I'd expect if we are aiming to make the project stable as a library.

I'd be more in favour of having an _extremely_ minimal publicly supported Go API surface to begin with, and gradually moving functions out and marking them as 'stable' on a case by case basis alongside whatever refactoring is required

content/en/docs/contributing/style-guide.md
SgtCoDFish: Do we count as consumers of cert-manager? :grin:  https://github.com/cert-manager/istio-csr/blob/master/pkg/server/server.go#L29

munnerz: Yes we do - however given we control all these things, it's far more manageable for us to manage these updates in tandem. It's a well known issue when depending on externally generated k8s clients whereby you can get yourself into impossible dependency situations 😬

content/en/docs/contributing/style-guide.md
SgtCoDFish: I added this because it's used in a bunch of existing open source stuff already: https://pkg.go.dev/github.com/jetstack/cert-manager@v1.3.1/pkg/util/pki?tab=importedby - as originally pointed out by @irbekrm

munnerz:
I really think given that none of the code admitted into the codebase so far has taken long term maintainability of the Go interfaces into consideration, if we do want to start doing so, we should start with nothing and work our way up, rather than just blanket declaring code we already have as 'stable' else we'll be creating a v2 very soon..

content/en/docs/contributing/style-guide.md
SgtCoDFish: I'd appreciate a link with details about the well known issue because it wasn't well known to me as a relative k8s beginner :grin:

In any case, I'll remove `pkg/client`

content/en/docs/contributing/style-guide.md
munnerz: I don't think that should be our benchmark for what should be stable though - I think that indicates we need to review the code in these packages to ensure it's all in a state where we _are_ comfortable supporting it long term, before doing so.

content/en/docs/contributing/style-guide.md
munnerz: I can try and find a link, but it typically manifests itself because:

1) code-generator does not guarantee Go API compatible output between minor releases of k8s
2) the things that are generated by code-generator can also be sensitive to the version of apimachinery being used between minor versions
3) this means between two minor versions of our k8s.io dependencies, the generated clients/packages here can change in non Go API compatible ways.

I know we have run into this in the past (hence why we have also been careful to not depend on too many external projects that also depend on k8s things - notable ones that we do depend on are controller-runtime, and for this one I have had to make PRs to bump their k8s dependency versions so that we are able to upgrade _numerous times_)

SgtCoDFish:
> I really think given that none of the code admitted into the codebase so far has taken long term maintainability of the Go interfaces into consideration, if we do want to start doing so, we should start with nothing and work our way up, rather than just blanket declaring code we already have as 'stable' else we'll be creating a v2 very soon..

Yeah I'd love to create a v2 with a minimal publicly exported surface area for this reason; it would be a huge shame to break the go module compatibility guarantee in v1 unless absolutely required, e.g. for a security issue.

We clearly agree on the approach generally; the list is _exactly_ me trying to "work our way up" after starting with nothing :smile:  Most of what's on this list currently is stuff that's currently being used by open source projects, or else stuff we actually tell people to import.

(starting to think it would be valuable to split the module compat list off into a separate PR because the style guide is several orders of magnitude more valuable and presumably less controversial given it's had no comments)

content/en/docs/contributing/style-guide.md
SgtCoDFish: That's super valuable, thanks very much for taking the time to explain, it's much appreciated :+1:

SgtCoDFish:
Commenting here because it'll be easier to follow in the future:

> I don't think that should be our benchmark for what should be stable though - I think that indicates we need to review the code in these packages to ensure it's all in a state where we are comfortable supporting it long term, before doing so.

Having a review of code that's going to be added to a possible "go module compatibility list" sounds like a really good step. I'm gonna split this PR into two - one for the style guide, and one for the module stuff, under the assumption that the style guide should sail through and the module stuff might need further review and discussion.

content/en/docs/contributing/_index.md
```diff
@@ -36,13 +37,14 @@ To get invites you can subscribe to [our mailing list](https://groups.google.com
...
 You can check what time this is where you are using [`everytimezone.com`](https://everytimezone.com/).
 
 ## Slack
+
 We have two cert-manager channels on [the Kubernetes Slack](https://slack.k8s.io) which we use to communicate:
 
-* `cert-manager` a general users channel, use this one for any usage related questions.
-* `cert-manager-dev` this is what the developers use to collaborate, please only use this for code related questions
+* `cert-manager`: a general channel for all users of cert-manager; use this one for any usage related questions.
+* `cert-manager-dev`: a channel for collaboration between cert-manager developers and maintainers; please only use this for code related questions
```
maelvls, line 50: some people come to help on the documentation and might not be developers per-se 😅

```suggestion
* `cert-manager-dev`: a channel for collaboration between cert-manager contributors and maintainers; please only use this for code related questions
```

SgtCoDFish: ah yeah, excellent point, I'll make the change!

content/en/docs/contributing/_index.md
```diff
@@ -36,13 +37,14 @@ To get invites you can subscribe to [our mailing list](https://groups.google.com
 We have 2 regular repeating meetings:
 
 * our daily stand-up meetings [on Google Meet](
-https://meet.google.com/eum-fyvt-xpa) at 10:30 London time every weekday 
-* our bi-weekly developer meetings [are on Google Meet](https://meet.google.com/abp-bwhk-wxc) 17:00 London time every 2 weeks (check calendar invites or [meeting notes](https://docs.google.com/document/d/1Tc5t6ylY9dhXAan1OjOoldeaoys1Yh4Ir710ATfBa5U)). 
+https://meet.google.com/eum-fyvt-xpa) at 10:30 London time every weekday
+* our bi-weekly developer meetings [are on Google Meet](https://meet.google.com/abp-bwhk-wxc) 17:00 London time every 2 weeks (check calendar invites or [meeting notes](https://docs.google.com/document/d/1Tc5t6ylY9dhXAan1OjOoldeaoys1Yh4Ir710ATfBa5U)).
```
irbekrm, line 41: Nit that I couldn't resist: 'bi-weekly...every 2 weeks' <- some redundancy of information

irbekrm (approved):
Looks good to me!

General question: I was thinking about what other things could go into this- if something like information about how to avoid test flakes fit into this doc, would it make sense to call it something like 'coding conventions' rather than style guide?

content/en/docs/contributing/_index.md
SgtCoDFish: > Nit that I couldn't resist: 'bi-weekly...every 2 weeks' <- some redundancy of information

Yep agreed, no need for redundancy there. I'll remove.

> would it make sense to call it something like 'coding conventions' rather than style guide

I think that's a much better name for the document, yeah. I'll change it.

content/en/docs/contributing/_index.md
```diff
@@ -35,14 +36,12 @@ To get invites you can subscribe to [our mailing list](https://groups.google.com
 
 We have 2 regular repeating meetings:
 
-* our daily stand-up meetings [on Google Meet](
-https://meet.google.com/eum-fyvt-xpa) at 10:30 London time every weekday 
-* our bi-weekly developer meetings [are on Google Meet](https://meet.google.com/abp-bwhk-wxc) 17:00 London time every 2 weeks (check calendar invites or [meeting notes](https://docs.google.com/document/d/1Tc5t6ylY9dhXAan1OjOoldeaoys1Yh4Ir710ATfBa5U)). 
-
-You can check what time this is where you are using [`everytimezone.com`](https://everytimezone.com/).
+* daily stand-up meetings [on Google Meet](https://meet.google.com/eum-fyvt-xpa) at [10:30 London time](http://www.thetimezoneconverter.com/?t=10:30&tz=Europe/London) every weekday
+* bi-weekly developer meetings [on Google Meet](https://meet.google.com/abp-bwhk-wxc) at [17:00 London time](http://www.thetimezoneconverter.com/?t=17:00&tz=Europe/London) (for dates, check calendar invites or [meeting notes](https://docs.google.com/document/d/1Tc5t6ylY9dhXAan1OjOoldeaoys1Yh4Ir710ATfBa5U))
```
SgtCoDFish, line 40: I've also added clickable direct timezone conversions, inspired by what the Kubernetes project does

irbekrm:
🥳 
I added the hold earlier- I guess everyone's questions were answered, so perhaps I can unhold

/hold cancel
/lgtm

maelvls:
@SgtCoDFish would it be worth adding a line such as

> We settled on the US spelling for both cert-manager.io and the cert-manager codebase.

PR #526, opened 2021-04-16
Coding-conventions: let's settle on the US spelling

maelvls:
I remember we talked about this once with @jsoref, but we did not actually write it down.

### DIFF
--- a/content/en/docs/contributing/coding-conventions.md
+++ b/content/en/docs/contributing/coding-conventions.md
@@ -53,9 +53,8 @@ enforced automatically by `goimports` when executed in the future.
 
 ## UK vs. US spelling
 
-For the sake of consistency, we settled on using the US spelling for the
-documentation in https://cert-manager.io as well as within the cert-manager
-codebase.
+For the sake of consistency, cert-manager uses US spelling for the documentation
+in https://cert-manager.io as well as within the cert-manager codebase.
 
 For example:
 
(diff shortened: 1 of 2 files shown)


content/en/docs/contributing/coding-conventions.md
```diff
@@ -50,3 +50,15 @@ import (
 
 Once this manual split of standard library, external and internal imports has been made, it will be
 enforced automatically by `goimports` when executed in the future.
+
+## UK vs. US spelling
+
+For the sake of consistency, we settled on using the US spelling for the
```
SgtCoDFish, line 56: nitpick: I think the first "the" isn't required, and I'd maybe phrase it very slightly differently:

```suggestion
For the sake of consistency, cert-manager uses US spelling for the
```

content/en/docs/contributing/coding-conventions.md
```diff
@@ -50,3 +50,15 @@ import (
...
+documentation in https://cert-manager.io as well as within the cert-manager
+codebase.
+
+For example:
+
+- "authorise" should be spelled "authorize",
+- "honour" should be spelled "honor",
+- "standardise" should be spelled "standardize".
```
SgtCoDFish, line 64: Rather than looking to give examples here, would it be helpful instead link to a more comprehensive list, e.g. https://wiki.ubuntu.com/EnglishTranslation/WordSubstitution ?

That list is from "translating" en_US -> en_GB, but I think it would work to illustrate the idea.

SgtCoDFish (commented):
I think this is good to document and it's good you raised it; I have a couple of fairly minor comments.

content/en/docs/contributing/coding-conventions.md
jsoref: I like this suggestion, since it means the spell-checker won't have to know to ignore the examples.

PR #527, opened 2021-04-20
Fix typo on the cert flow explanation

elpablete:
Just a simple typo fix


irbekrm (approved):
Thanks for spotting this and PR-ing in the fix! 🎉

Issue #397, opened 2020-12-30
Restore Procedure: Prevent Cert Re-order


chainlink:
When restoring, if i install cert manager first before applying the backed up resources, it tries to re-issue certs for ingresses present on the cluster.

Backup:
Add step to backup certificate secrets - _Is it better to backup these secrets, or let LE re-issue?_

I think the restore procedure should be:
1. Restore CRDs
2. Restore (cleaned) backup (see #234)
3. Install Cert manager

Thoughts? Happy to PR the change to the docs if this makes sense

meyskens:
> Is it better to backup these secrets, or let LE re-issue?

Probably better to backup the secrets too unless you think the storage medium of your backups is compromised/easy to infiltrate. 
It will allow to:
- restore certificates before everything us online to solve the ACME challenges
- not hit the LE rate limits

If you are sharing the backup on something like a Git repo, Slack, shared computer,... I would avoid it as it will leak your private key.

wallrj:
Thanks for the bug report @chainlink 

I'm going to close this issue in favour of another more general issue about improving the Backup and restore instructions 
 * https://github.com/cert-manager/website/issues/234

If you have time to contribute some improved documentation  as a PR we'd be most grateful and I'll help review and test it.

/close

Issue #234, opened 2020-06-15
Backup and Restore Resources


ailox:
This issue is related to https://github.com/jetstack/cert-manager/issues/2112.

I'm sorry, but the backup/restoration process does simply not work as advertised, and conveys a false sense of security. I'd argue it should be removed or, at the very least, augmented with a warning of some kind.

rpherrera:
This procedure is really misleading and does not work like that...  Luckily I did it on staging environment first...

rpherrera:
I could manage to restore my backup using the following (suboptimal/error prone/bizarre) procedure:

* Take a 2nd backup from the created resources that we wanna overwrite:

```
kubectl get -o yaml \
     --all-namespaces \
     issuer,clusterissuer,certificates,certificaterequests > cert-manager-backup-2.yaml
```

* Just to play safe: please don't forget to version your 1st backup YAML file before continuing...

* Put your 1st backup (let's say: `cert-manager-backup-1.yaml`) side by side with the 2nd backup (`cert-manager-backup-2.yaml`) on the text editor...

* On the 1st backup file, make sure to replace the following fields using the values from the 2nd backup on their equivalent code regions:
    * `creationTimestamp`
    *  `resourceVersion`
    * `uid`
    * `lastTransitionTime`

* Apply the changes you made on top of your 1st backup file:

```
kubectl apply -f cert-manager-backup-staging.yaml
```

* It worked for me and I hope that helps while we don't have a "clean" backup/restore procedure that we can follow...

wallrj:
Hey @ailox 

Thanks for the bug report. There are a bunch of similar issues in the cert-manager repo too. 
We need to document (and test) an improved process for restoring cert-manager backups:

wallrj:
In https://github.com/jetstack/cert-manager/issues/2112#issuecomment-619076005 @munnerz  wrote:
As noted above, I don't think it's trivial for us to create an entirely 'restorable' resource that can be applied during the restore phase.

I think we can take this issue and turn it into one of two things:

    Documenting how to perform the backup/restore process with something like Velero
    Extending our own CLI tool to have a dedicated backup and restore subcomannd.

I think (2) would be really interesting/useful for users, although in a lot of cases (1) may be more appropriate if you're running backups across a large number of clusters/regularly.

PR #531, opened 2021-04-22
Update backup instructions

### LINKED ISSUE #479: Update backup instructions
Backup instructions [here](https://cert-manager.io/docs/tutorials/backup/) could be updated.

We could add instructions as to how to update with Velero as well as a note about the order of restore (i.e `Secrets` should be created before `cert-manager` `CR`s and any ingress resources using `ingress-shim` )

irbekrm:
I did some manual backup and restore testing (mostly with Velero on GKE) and added a few more notes to the backup doc.

The fact that the `Certificate`s for `Ingress` lose the correct owner reference during a restore and `cert-manager` cannot recreate it, seems problematic from backup and restore perspective, as there is no straightforward way to exclude the `Ingress` `Certificate`s only from a Velero backup. 

I added a recommendation to not restore `CertificateRequest`s.
I don't think that restoring them would necessarily cause failures as such, but with the new identity fields it seems like we should not restore them since the restorer's identity is likely going to be different than creator's and it might just cause confusion.
Also `cert-manager` cannot recreate their status after a restore correctly, so they end up having Ready condition set to False, which I think would also be confusing for users.
In all cases I could think of there should be no need to restore completed `CertificateRequest`s, but perhaps there are some edge cases where they are needed?

Some issues related to `cert-manager` backup and restore:

- [#2318](https://github.com/jetstack/cert-manager/issues/2318)
- [#2112](https://github.com/jetstack/cert-manager/issues/2112)
- #234
- #397 
- [#3612](https://github.com/jetstack/cert-manager/issues/3612)

Some of the issues above refer to the currently documented backup strategy via `kubectl get.. kubectl apply`- I have not looked into that much as a part of this PR.
I have not yet properly tested restoring from an etcd snapshot which I was planning- that might be a separate PR.

Fixes #479 

/kind cleanup

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -139,6 +139,7 @@ propagations
 publicised
 RBAC
 rebase
+reissuance
 remediate
 repo
 reStructuredText
--- a/content/en/docs/tutorials/backup.md
+++ b/content/en/docs/tutorials/backup.md
@@ -13,8 +13,8 @@ re-install.
 
 The following commands will back up the configuration of `cert-manager`
 resources. Doing that might be useful before upgrading `cert-manager`. As
-this backup does not include the Secrets containing the X.509 certificates,
-restoring to a cluster that does not already have those Secrets will result in
+this backup does not include the `Secrets` containing the X.509 certificates,
+restoring to a cluster that does not already have those `Secret`s will result in
 the certificates being reissued.
 
 ### Backup
@@ -28,24 +28,24 @@ $ kubectl get -o yaml \
 ```
 
 If you are transferring data to a new cluster, you may also need to copy across
-additional Secret resources that are referenced by your configured Issuers, such
+additional `Secret` resources that are referenced by your configured Issuers, such
 as:
 
 #### CA Issuers
 
-- The root CA Secret referenced by `issuer.spec.ca.secretName`
+- The root CA `Secret` referenced by `issuer.spec.ca.secretName`
 
 #### Vault Issuers
 
-- The token authentication Secret referenced by
+- The token authentication `Secret` referenced by
   `issuer.spec.vault.auth.tokenSecretRef`
-- The AppRole configuration Secret referenced by
+- The AppRole configuration `Secret` referenced by
   `issuer.spec.vault.auth.appRole.secretRef`
 
 #### ACME Issuers
 
-- The ACME account private key Secret referenced by `issuer.acme.privateKeySecretRef`
-- Any Secrets referenced by DNS providers configured under the
+- The ACME account private key `Secret` referenced by `issuer.acme.privateKeySecretRef`
+- Any `Secret`s referenced by DNS providers configured under the
   `issuer.acme.dns01.providers` and `issuer.acme.solvers.dns01` fields.
 
 ### Restore
@@ -77,12 +77,12 @@ are using [`ingress-shim`](../../usage/ingress/).
 
 #### Excluding some cert-manager resources from backup
 
-`cert-manager` has a number of resources- `Order`s, `Challenge`s and
-`CertificateRequest`s -that are designed to represent a point-in-time operation
-such as a request for a certificate. As such, their status often depends on other
-ephemeral resources (i.e a temporary Secret holding a private key), so
-`cert-manager` cannot always correctly re-create the status of these resources
-by just looking at the cluster state at some later time.
+`cert-manager` has a number of custom resources that are designed to represent a
+point-in-time operation. An example would be a `CertificateRequest` that
+represents a one-time request for an X.509 certificate. The status of these
+resources can depend on other ephemeral resources (such as a temporary `Secret`
+holding a private key) so `cert-manager` might not be able to correctly recreate
+the state of these resources at a later point.
 
 In most cases backup and restore tools will not restore the statuses of custom resources,
 so including such one-time resources in a backup can result in an unnecessary reissuance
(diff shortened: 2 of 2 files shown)


content/en/docs/tutorials/backup.md
```diff
@@ -49,5 +57,114 @@ created above after installing cert-manager.
...
+This section refers to backing up and restoring 'all' Kubernetes resources in a
+cluster- including some `cert-manager` ones- for scenarios such as disaster
+recovery, cluster migration etc.
+
+Note that we have not done any extensive testing, only a few common scenarios-
+it is recommended to test the backup and restore strategy before putting it in place
+as there are likely to be some other edge cases.
+We would like to ensure that users can reliably backup and restore `cert-manager`
+resources using common tools- if you encounter any errors, please do open a
+GitHub issue or PR an update to this document.
```
wallrj, line 71:
```suggestion
*Note*: We have tested this process on simple Kubernetes test clusters with a limited set of Kubernetes releases. To avoid data loss, please test both the backup and the restore strategy on your own cluster before depending upon it in production. If you encounter any errors, please open a GitHub issue or a PR to document variations on this process for different Kubernetes environments. 
```

I suggest something like that ^^  instead, but happy to go with the original version if you prefer.

content/en/docs/tutorials/backup.md
```diff
@@ -49,5 +57,114 @@ created above after installing cert-manager.
...
+
+### Avoiding unnecessary certificate reissuance
+
+#### Order of restore
+
+If `cert-manager` does not find a Kubernetes Secret with an X.509 certificate
+for a `Certificate`, reissuance will be triggered. To avoid unnecessary
+reissuance after a restore, ensure that Secrets are restored before
+`Certificate`s. Similarly, Secrets should be restored before `Ingress`es if you
+are using [`ingress-shim`](../../usage/ingress/).
```
wallrj, line 81:
```suggestion
If `cert-manager` does not find a Kubernetes `Secret` with an X.509 certificate
for a `Certificate`, reissuance will be triggered. To avoid unnecessary
reissuance after a restore, ensure that `Secret`s are restored before
`Certificate`s. Similarly, `Secret`s should be restored before `Ingress`es if you
are using [`ingress-shim`](../../usage/ingress/).
```

Maybe quote `Secret` for consistency with the quoting of other Kubernetes resource types.

content/en/docs/tutorials/backup.md
```diff
@@ -49,5 +57,114 @@ created above after installing cert-manager.
...
+ephemeral resources (i.e a temporary Secret holding a private key), so
+`cert-manager` cannot always correctly re-create the status of these resources
+by just looking at the cluster state at some later time.
+
+In most cases backup and restore tools will not restore the statuses of custom resources,
+so including such one-time resources in a backup can result in an unnecessary reissuance
+after a restore as without the status fields `cert-manager` will not be able to tell that,
+for example, an `Order` has already been fulfilled.
+To avoid unnecessary reissuance, we recommend that `Order`s and `Challenge`s are excluded
+from the backup. We also don't recommend backing up `CertificateRequest`s, see [Backing up CertificateRequests](#backing-up-certificaterequests)
```
wallrj, line 97: That's a nice explanation.

wallrj (changes requested):
Thanks for adding all that extra detail.

* I added a few suggestions
* I'd prefer the use of `(...)` or `—` instead of `- `, and in some of those cases, separate sentences might be easier to read.

.spelling
```diff
@@ -139,6 +139,7 @@ propagations
 publicised
 RBAC
 rebase
+reissuance
```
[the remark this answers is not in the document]
irbekrm: Thanks- I've changed `re-issue` -> `reissue` for consistency.

irbekrm:
Thanks for the review @wallrj - I've added updated the PR. Thanks for looking at the sentences, I did feel like they were rambling a bit, so was hoping someone could suggest some improvements.

content/en/docs/tutorials/backup.md
```diff
@@ -49,5 +57,109 @@ created above after installing cert-manager.
...
+are using [`ingress-shim`](../../usage/ingress/).
+
+#### Excluding some cert-manager resources from backup
+
+`cert-manager` has a number of resources- `Order`s, `Challenge`s and
+`CertificateRequest`s -that are designed to represent a point-in-time operation
+such as a request for a certificate. As such, their status often depends on other
+ephemeral resources (i.e a temporary Secret holding a private key), so
+`cert-manager` cannot always correctly re-create the status of these resources
+by just looking at the cluster state at some later time.
```
wallrj, line 85:
```suggestion
`cert-manager` has a number of resources — `Order`s, `Challenge`s and
`CertificateRequest`s  — that are designed to represent a point-in-time operation
such as a request for a certificate. As such, their status often depends on other
ephemeral resources (i.e a temporary Secret holding a private key), so
`cert-manager` cannot always correctly re-create the status of these resources
by just looking at the cluster state at some later time.
```

wallrj (changes requested):
I should have highlighted it in the previous review, but there's a lingering dash which should be an emdash, or bracketed.

content/en/docs/tutorials/backup.md
irbekrm: Thanks @wallrj ! I have eliminated those dashes by re-writing this using shorter separate sentences here and in a few more places 😄

wallrj (approved):
/lgtm
/approve

PR #548, opened 2021-05-03
Move the "Approval API" documentation to /concepts/certificaterequest

maelvls:
| Preview: https://deploy-preview-548--cert-manager.netlify.app/docs/concepts/certificaterequest/#approval |
|-|

The is part of the effort of making the website more user-friendly. The `certificaterequest` page did not feel very user-friendly. When looking at that page, I am confused by some terms:

- "signer" vs. "issuer"
- "managed CertificateRequest",
- "UserInfo",
- "Requestor".

I went on and tried defining these terms and have a diagram that explains what is the relation between an issuer and an approver.

Related:
- comment 1: https://github.com/jetstack/cert-manager/pull/3797#discussion_r614803873
- comment 2: https://github.com/jetstack/cert-manager/pull/3797#discussion_r614830317

/kind cleanup

### DIFF
--- a/content/en/docs/concepts/certificaterequest.md
+++ b/content/en/docs/concepts/certificaterequest.md
@@ -222,14 +222,7 @@ CertificateRequests:
 
 - Or you can write your own custom approver.
 
-### The `Approved` and `Denied` conditions and approvers
-
-The approval API is made of two mutually exclusive conditions that can be set on
-the CertificateRequests status: `Approved` and `Denied`. When the approval API
-is enabled, a CertificateRequest must have the `Approved` condition before a
-signer (such as the ACME issuer) can issue the certificate.
-
-{{% pageinfo color="info" %}}
+### Signer vs. Issuer
 
 As part of the approval API, we decided to make use of the term "signer"
 instead of "issuer" as a way to come closer to the terms used in the
@@ -247,7 +240,12 @@ signer. By "issuer" we refer to both external issuers (such as
 [aws-pca-issuer](https://github.com/jniebuhr/aws-pca-issuer)) and internal
 issuers (such as the ACME issuer embedded into cert-manager).
 
-{{% /pageinfo %}}
+### The `Approved` and `Denied` conditions and approvers
+
+The approval API is made of two mutually exclusive conditions that can be set on
+the CertificateRequests status: `Approved` and `Denied`. When the approval API
+is enabled, a CertificateRequest must have the `Approved` condition before a
+signer (such as the ACME issuer) can issue the certificate.
 
 The approval API kicks in right before the signer (e.g., the ACME issuer) signs
 the X.509 CSR contained in a CertificateRequest. We call "approver" the
@@ -259,7 +257,7 @@ The following diagrams shows the two possible scenarios.
 #### ✅ The `Approved` scenario {#approved-scenario}
 
 Note that the issuer is responsible for setting the `Ready=True` condition. The
-approver does not set `Ready`, nor does cert-manager.
+approver (including the built-in cert-manager approver) does not set `Ready`.
 
 ```diagram
 kind: CertificateRequest              kind: CertificateRequest              kind: CertificateRequest
@@ -587,10 +585,10 @@ two rules:
    `Approved`. If the CertificateRequest is marked as `Denied` or has no
    approval condition altogether, the signer is expected not to sign it.
 
-\*Some issuers [do not
-support](/docs/configuration/external/#issuers-that-honour-approval) the
-approval API yet. Such approvers break the rule (2). cert-manager detects
-CertificateRequests that have been signed by an issuer that ignored the approval
-API. When it detects one, cert-manager pretends that the CertificateRequests has
-been approved. To avoid this issue, make sure to run only issuers that support
-the approval API.
+\*Some issuers may not support the approval API. The approver that support the
+Approval API are listed in the [external issuers
+page](/docs/configuration/external/). Approvers that do support the Approval API
+will be breaking the rule (2). cert-manager detects CertificateRequests that
+have been signed by an issuer that ignored the approval API. When it detects
+one, cert-manager pretends that the CertificateRequests has been approved. To
+avoid this issue, make sure to run only issuers that support the approval API.
--- a/content/en/docs/concepts/certificaterequest.md
+++ b/content/en/docs/concepts/certificaterequest.md
@@ -185,7 +185,7 @@ extraArgs:
   - --controllers=*,-certificaterequests-approver
 ```
 
-{{% alert title="Disable approve-all" color="info" %}}
+{{% alert title="Disable approve-all" color="primary" %}}
 
 The flag `--controllers=*,-certificaterequests-approver` prevents cert-manager
 from starting its allow-all built-in approval controller.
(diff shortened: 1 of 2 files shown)


maelvls:
I am now finished with this change. I am hoping to get feedback from @JoshVanL and whoever would like to help 😅

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
+
+When a user or controller attempts to approve or deny a CertificateRequest, the
+cert-manager webhook evaluates whether it has sufficient permissions to do so.
+These permissions are based upon the request itself, and more specifically, the
+CertificateRequest's `issuerRef` field.
+
+{{% pageinfo color="warning" %}}
+
+The verb `approve` on the resource `signer` only works with `ClusterRole`, and
+the `apiGroup` must be set to `cert-manager.io`.
```
JoshVanL, line 442: This seems quite random and these two things don't feel like they should be in the same sentence. Let's keep the first part and change it to: The approver needs the approve permissions at the cluster level.

The group is obvious since we are interacting with cert-manager.

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
+  name: private-ca
+spec:
+  arn: arn-c1f4f80cd072
+  region: eu-west-1
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: letsencrypt
+  namespace: namespace-with-letsencrypt
```
JoshVanL, line 461: Seems like a strange namespace name :sweat_smile:

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
 rules:
   - apiGroups: ["cert-manager.io"]
     resources: ["signers"]
     verbs: ["approve"]
-    resourceNames: ["myissuers.my-example.io/myapp"]
+    resourceNames:
+      # Allow this specific AWSPCAClusterIssuer issuer:
+      - "awspcaissuers.awspca.cert-manager.io/private-ca"
+      #  <-----------> <--------------------> <------------------> <-------->
+      #  resource type     resource group          signer name     issuer name
```
JoshVanL, line 479:
```
<-------->
issuer name
````

Don't think this should be here?

File: content/en/docs/concepts/certificaterequest.md [hunk 4]
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
-resources: ["signers"]
-verbs: ["approve"]
-resourceNames:
- # namesapced signers
- - "<signer-resource-name>.<signer-group>/<signer-namespace>.<signer-name>"
- # cluster scoped signers
- - "<signer-resource-name>.<signer-group>/<signer-name>"
- # all signers of this resource name
- - "<signer-resource-name>.<signer-group>/*"
-```
```
JoshVanL, line 199: We have deleted this yaml snippet, however I find this much easier to read and understand as a kube user.

Is there a way we can include this fist, and then follow on with a worked example please?

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
-- The verb must *always* be `approve`, which grants the approver the permissions to set *both* Approved and Denied conditions
+```sh
+kubectl cert-manager approve cert-request-1
+```
 
-An example of signing all `myissuer` signers in all namespaces, and
-`clustermyissuers` with the name `myapp`, in the `my-example.io` group:
+Here is what the approved CertificateRequest looks like:
 
 ```yaml
```
JoshVanL, line 542: We are we deleting this list? I think we should have a concise list, before walking into a worked example.

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
+kind: CertificateRequest
+status:
+  conditions:
+    - type: Approved
+      status: "True"
+      reason: opa-approver-controller
+      message: Certificate request was approved by opa-approver-controller
+```
+
+> The `opa-approver-controller` is hypothetical and does not exist yet.
```
JoshVanL, line 521: We don't need to include this.

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
+```
 
-The Approved and Denied conditions are two distinct condition types on the
-CertificateRequest. These conditions must only have the status of True, and
-are mutually exclusive (i.e. a CertificateRequest cannot have an Approved and
-Denied condition simultaneously). This behavior is enforced in the cert-manager
-validating admission webhook.
+Or set it in `values.yaml`:
 
-An "approver" is an entity that is responsible for setting the Approved/Denied
```
JoshVanL, line 115: This definition seems to have been lost

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
+### Reason and message of the Approval API
+
+The `reason` and `message` fields on the `Denied` and `Approved` condition
+are approver-specific, meaning that each approver will set their own. Here
+is the set of guidelines:
+
+| Field     | Guideline                                               |
+| --------- | ------------------------------------------------------- |
+| `reason`  | Machine-readable string identifying the approver        |
+| `message` | Human-readable explanation of why the condition was set |
```
JoshVanL, line 507: This is moving away from the original intention. Please add back the original definitions.

```
The Reason field of the Approved/Denied condition should be set to *who* set the
condition. Who can be interpreted however makes sense to the approver
implementation. For example, it may include the API group of an approving policy
controller, or the client agent of a manual request.
````


````
The Message field of the Approved/Denied condition should be set to *why* the
condition is set. Again, why can be interpreted however makes sense to the
implementation of the approver. For example, the name of the resource that
approves this request, the violations which caused the request to be denied, or
the team to who manually approved the request.
```

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
+kind: CertificateRequest
+status:
+  conditions:
+    - type: Approved
+      status: "True"
+      reason: cert-manager.io
+      message: Certificate request has been approved by cert-manager.io
 ```
 
-If the approver does not have sufficient permissions defined above to set the
```
JoshVanL, line 217: Lost?

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
 
+If you are using Helm, you can add the following flag to your `helm install`
+incantation:
 
-#### Behavior
+```sh
+--set extraArgs="{--controllers=*,-certificaterequests-approver}"
+```
 
-The Approved and Denied conditions are two distinct condition types on the
```
JoshVanL, line 109: Lost?

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
+2. The signer is only\* able to sign a CertificateRequest that is marked as
+   `Approved`. If the CertificateRequest is marked as `Denied` or has no
+   approval condition altogether, the signer is expected not to sign it.
+
+\*Some issuers may not support the approval API. The approver that support the
+Approval API are listed in the [external issuers
+page](/docs/configuration/external/). Approvers that do support the Approval API
+will be breaking the rule (2). cert-manager detects CertificateRequests that
+have been signed by an issuer that ignored the approval API. When it detects
+one, cert-manager pretends that the CertificateRequests has been approved. To
```
JoshVanL, line 593: cert-manager does not pretend anything :)

content/en/docs/concepts/certificaterequest.md
```diff
@@ -72,170 +97,498 @@ Each ready condition consists of the pair `Ready` - a boolean value, and
...
+1. The approver can only set `Approved` or `Denied` on CertificateRequests that
+   reference an `Issuer` or a `ClusterIssuer` (or any other external issuer) for
+   which it has the permission to approve.
+2. The signer is only\* able to sign a CertificateRequest that is marked as
+   `Approved`. If the CertificateRequest is marked as `Denied` or has no
+   approval condition altogether, the signer is expected not to sign it.
+
+\*Some issuers may not support the approval API. The approver that support the
+Approval API are listed in the [external issuers
+page](/docs/configuration/external/). Approvers that do support the Approval API
```
JoshVanL, line 590: This paragraph needs changing- all known issuers now support the approval API.

JoshVanL (changes requested):
Thanks @maelvls, I have made a pass and left some comments.

While I think the worked examples are a great addition, it seems like we have _lost_ information in this doc, and in some cases saying the wrong thing.

I would really like to keep the concise "list style" points that were here previously, and _then_ follow into a worked example. This saves a lot of time for users who know about this topic already, and just need a quick glance at the key points.

To make sure nothing gets lost, perhaps this can be split into multiple PRs for each section? I've found it difficult to see what has been removed, changed/moved or both.

/assign @maelvls

inteon:
This PR seems to be out-of-date.
I think most of the info added in this PR is available on the website now.

PR #552, opened 2021-05-06
Do not specify namespace for DO DNS01 secret

jgaskins:
The examples for `Issuer` do not specify a namespace, so when we copy/paste the examples they go into the current namespace. Specifying that this secret goes in the `cert-manager` namespace potentially puts it into a _different_ namespace than the `Issuer`, creating confusion when it doesn't work.

This PR suggests implying placement of the secret in the same namespace as other resources being created.

### DIFF
--- a/content/en/docs/configuration/acme/dns01/digitalocean.md
+++ b/content/en/docs/configuration/acme/dns01/digitalocean.md
@@ -14,7 +14,6 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: digitalocean-dns
-  namespace: cert-manager
 data:
   # insert your DO access token here
   access-token: "base64 encoded access-token here"


SgtCoDFish:
@jgaskins thanks for raising the PR, it's much appreciated!

I can see that the reference to the digital ocean API token will only search in the same namespace as the resource referencing the secret: 

`ACMEIssuerDNS01ProviderDigitalOcean`[1] -> `SecretKeySelector`[2] -> `LocalObjectReference`[3]

> [LocalObjectReference is a] reference to an object in the same namespace as the referent.

So I can see why the example would be frustrating. I agree with the change you've proposed.

/ok-to-test

[1] https://github.com/jetstack/cert-manager/blob/2aa625edcb1a179600d4c557216c7bfeb44ae269/pkg/apis/acme/v1/types_issuer.go#L418-L422
[2] https://github.com/jetstack/cert-manager/blob/2aa625edcb1a179600d4c557216c7bfeb44ae269/pkg/apis/meta/v1/types.go#L63-L74
[3] https://github.com/jetstack/cert-manager/blob/2aa625edcb1a179600d4c557216c7bfeb44ae269/pkg/apis/meta/v1/types.go#L39-L49

SgtCoDFish:
/lgtm

Thanks again! :grin:

PR #558, opened 2021-05-11
While restoring a backup, the uid and resourceVersion should be ignored

maelvls:
While reading about https://github.com/jetstack/cert-manager/issues/3512, I wondered: why are we letting people apply a `backup.yaml` manifest that contains `uid` and `resourceVersion` fields?

In this PR, I propose to change the "restore" step to:

```sh
kubectl apply -f <(awk '!/^ *(resourceVersion|uid): [^ ]+$/' backup.yaml)
```

so that both the UIDs and resourceVersions are stripped.

cc @irbekrm

### DIFF
--- a/content/en/docs/tutorials/backup.md
+++ b/content/en/docs/tutorials/backup.md
@@ -13,18 +13,16 @@ re-install.
 
 The following commands will back up the configuration of `cert-manager`
 resources. Doing that might be useful before upgrading `cert-manager`. As
-this backup does not include the `Secrets` containing the X.509 certificates,
-restoring to a cluster that does not already have those `Secret`s will result in
-the certificates being reissued.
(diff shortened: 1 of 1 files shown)


maelvls:
@irbekrm Do you think this PR looks OK?

irbekrm:
Thanks for this change, looks good to me!

I've been having some doubts about that specific part in the backup doc, there seems to be a whole bunch of edge cases, i.e in case of `ingress-shim` `Certificate`s.
But I guess, in any case, it does make sense to exclude resource version and uid.
I've ran the command and verified that it works.

/lgtm
/approve

irbekrm:
I guess it could have also made sense to just provide some command how to strip the uid and resource version from backup file- that would help with https://github.com/jetstack/cert-manager/issues/3512 where they are deploying with flux.

PR #560, opened 2021-05-14
ExperimentalCertificateControllers=true no longer needed for key rotation

sfc-gh-cxu:
In fact, if `--feature-gates=ExperimentalCertificateControllers=true` is provided, cert-manager would abort with errors

### DIFF
--- a/content/en/docs/usage/certificate.md
+++ b/content/en/docs/usage/certificate.md
@@ -132,11 +132,6 @@ issued.
 
 ## Configuring private key rotation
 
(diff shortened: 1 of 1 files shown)


sfc-gh-cxu:
@munnerz @JoshVanL would you plase take a look? Thank you!

JoshVanL:
This shouldn't be in the docs any more as is no longer a feature flag. Thanks for catching and fixing this @sfc-gh-cxu 

/ok-to-test
/approve
/lgtm

PR #575, opened 2021-05-25
Make warning about clashing ports on AWS more clear

SgtCoDFish:
Discussed briefly in slack; this is apparently a gotcha that keeps popping up

Signed-off-by: Ashley Davis <ashley.davis@jetstack.io>

/kind documentation

### DIFF
--- a/content/en/docs/installation/compatibility.md
+++ b/content/en/docs/installation/compatibility.md
@@ -30,12 +30,13 @@ for more details.
 
 When Google configure the control plane for private clusters, they automatically
 configure VPC peering between your Kubernetes cluster's network and a separate
-Google managed project.
+Google-managed project.
 
 In order to restrict what Google are able to access within your cluster, the
-firewall rules configured restrict access to your Kubernetes pods. This will
-mean that you will experience the webhook to not work and experience errors such
-as `Internal error occurred: failed calling admission webhook ... the server is currently unable to handle the request`.
+firewall rules configured restrict access to your Kubernetes pods. This means
+that the webhook won't work, and you'll see errors such as
+`Internal error occurred: failed calling admission webhook ... the server is
+currently unable to handle the request`.
 
 In order to use the webhook component with a GKE private cluster, you must
 configure an additional firewall rule to allow the GKE control plane access to
--- a/content/en/docs/installation/compatibility.md
+++ b/content/en/docs/installation/compatibility.md
@@ -6,7 +6,25 @@ type: "docs"
 ---
 
 Below you will find details on various compatibility issues and quirks that you
-may be effected by in your environment.
+may be affected by when deploying cert-manager. If you believe we've missed something
+please feel free to raise an issue or a pull request with the details!
+
+{{% alert title="Important" color="warning" %}}
+If you're using AWS Fargate or else if you've specifically configured
+cert-manager to run the host's network, be aware that kubelet listens on port
+`10250` by default which clashes with the default port for the cert-manager
+webhook.
+
+As such, you'll need to change the webhook's port when setting up cert-manager.
+
+For installations using Helm, you can set the `webhook.securePort` parameter
+when installing cert-manager either using a command line flag or an entry in
+your `values.yaml` file.
+
+If you have a port clash, you could see confusing error messages regarding
+untrusted certs. See [#3237](https://github.com/jetstack/cert-manager/issues/3237)
+for more details.
+{{% /alert %}}
 
 ## GKE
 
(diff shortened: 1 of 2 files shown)


content/en/docs/installation/compatibility.md
```diff
@@ -40,9 +40,13 @@ To address this, the webhook can be run in the host network so it can be reached
 by cert-manager, by setting the `webhook.hostNetwork` key to true on your
 deployment, or, if using Helm, configuring it in your `values.yaml` file.
 
-Note that since kubelet uses port `10250` by default on the host network, the
-`webhook.securePort` value must be changed to a different, free port.
+**IMPORTANT**: kubelet on AWS uses port `10250` by default on the host network, which
```
irbekrm, line 43: Suggestion/nit: 'AWS' -> 'AWS EKS'

content/en/docs/installation/compatibility.md
```diff
@@ -40,9 +40,13 @@ To address this, the webhook can be run in the host network so it can be reached
...
+clashes with the default port for the cert-manager webhook.
 
+As such, the `webhook.securePort` value must be changed to a different, free port when
```
irbekrm, line 46: Suggestion: specify that `webhook.securePort` is a Helm chart parameter and suggest how to do the same with standard manifests (seems to be passing a flag and setting a port by looking at [the Helm template](https://github.com/jetstack/cert-manager/blob/master/deploy/charts/cert-manager/templates/webhook-deployment.yaml))

I guess it might be nice to do the same with `webhook.hostNetwork` above.

irbekrm (changes requested):
Thanks for doing this- it's a lot clearer now and I think this section gets used quite a lot too!

I've left two nits, otherwise happy to lgtm.

content/en/docs/installation/compatibility.md
irbekrm: Arrgh need more coffee 😆 
Ignore the above nit- kubelet always uses port 10250 (by default) https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
I suggest to remove the 'on AWS' bit 😄

content/en/docs/installation/compatibility.md
```diff
@@ -40,9 +40,13 @@ To address this, the webhook can be run in the host network so it can be reached
 by cert-manager, by setting the `webhook.hostNetwork` key to true on your
 deployment, or, if using Helm, configuring it in your `values.yaml` file.
 
-Note that since kubelet uses port `10250` by default on the host network, the
-`webhook.securePort` value must be changed to a different, free port.
+**IMPORTANT**: kubelet on AWS uses port `10250` by default on the host network, which
+clashes with the default port for the cert-manager webhook.
```
maelvls, line 44: suggestion: use Docsy's "warning" shortcode:

```suggestion
{{% alert title="Important" color="warning" %}}

kubelet on AWS uses port `10250` by default on the host network, which
clashes with the default port for the cert-manager webhook.

{{% /alert %}}
```

It looks like this:

warning-docsy

content/en/docs/installation/compatibility.md
```diff
@@ -6,7 +6,22 @@ type: "docs"
 ---
 
 Below you will find details on various compatibility issues and quirks that you
-may be effected by in your environment.
+may be affected by when deploying cert-manager. If you believe we've missed something
+please feel free to raise an issue or a pull request with the details!
+
+{{% alert title="Important" color="warning" %}}
+If you run cert-manager using the host's network, be aware that kubelet listens on port `10250`
```
irbekrm, line 13: Suggestion: perhaps we should also add that by default `cert-manager` is not using host's network? I can see how it's good to make this warning clear, but am a little bit worried that we may be confusing someone who has just opened to page to see if they can run `cert-manager` on GKE etc. Alternatively I guess we could also move the warning to the AWS section or make a separate subsection for running on host's network and link to that from AWS EKS/AWS Fargate sections (probably overcomplicated).

irbekrm (approved):
Looks good to me, glad we'll have this bit clearer.

I've added an optional suggestion- ignore if you think the warning should stay as is.

irbekrm:
Adding lgtm with a hold- as I added an optional suggestion. Do unhold if you'd rather merge as is!

/hold
/lgtm

content/en/docs/installation/compatibility.md
SgtCoDFish: yeah I think that could be clearer, good suggestion. I'll take a swing at it.

irbekrm:
Thanks! This looks good to me 🥳 

/hold cancel
/lgtm

PR #577, opened 2021-05-27
improve language describing ingress hosts

### LINKED ISSUE #475: Docs imply hosts is optional but it isn't
> tls: # < placing a host in the TLS config will indicate a certificate should be created
https://github.com/cert-manager/website/blame/59df6949822a40319fb0521662f897b2260007a9/content/en/docs/usage/ingress.md#L40

This implies that you can just specify a TLS config without a host, perhaps for an pre-generated cert. However when trying this you get the following error:

```
> kubectl describe ingress web
[...]
Warning  BadConfig  3m11s (x2 over 3m11s)  cert-manager              TLS entry 0 is invalid: secret "web-crt" for ingress TLS has no hosts specified
```

This is expected if you read the [ingress docs](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls), but this comment in the cert-manager docs is quite misleading.

SgtCoDFish:
fixes #475

(also removes a single trailing whitespace)

### DIFF
--- a/content/en/docs/usage/ingress.md
+++ b/content/en/docs/usage/ingress.md
@@ -38,9 +38,9 @@ spec:
         backend:
           service:
             name: myservice
-            port: 
+            port:
               number: 80
-  tls: # < placing a host in the TLS config will indicate a certificate should be created
+  tls: # < placing a host in the TLS config will determine what ends up in the cert's subjectAltNames
   - hosts:
     - example.com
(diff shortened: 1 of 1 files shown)


irbekrm (approved):
Looks good to me- definitely a lot clearer!

SgtCoDFish:
haha I was about to post this and #575 on #cert-manager-dev asking for reviews and you've already done it! :1st_place_medal: thanks very much :grin:

PR #578, opened 2021-05-28
A note of caution about the use of ca.crt

wallrj:
| Preview | https://deploy-preview-578--cert-manager.netlify.app/docs/concepts/certificate/ |
|-|-|

### DIFF
--- a/content/en/docs/concepts/certificate.md
+++ b/content/en/docs/concepts/certificate.md
@@ -62,6 +62,31 @@ which contains the following explanation:
 > assumption that the remote end must already possess it in order to
 > validate it in any case.
 
+{{% alert title="Use ca.crt with caution" color="warning" %}}
+
+When configuring a client to connect to a TLS server with a serving certificate that is signed by a private CA,
+you will need to provide the client with the CA certificate in order for it to verify the server.
+`ca.crt` will likely contain the certificate you need to trust,
+but __do not mount the same `Secret` as the server__ to access `ca.crt`.
+This is because:
+
+1. That `Secret` also contains the private key of the server, which should only be accessible to the server.
+   You should use RBAC to ensure that the `Secret` containing the serving certificate and private key are only accessible to Pods that need it.
+2. Rotating CA certificates safely relies on being able to have both the old and new CA certificates trusted at the same time.
+   By consuming the CA directly from the source, this isn't possible;
+   you'll be _forced_ to have some down-time in order to rotate certificates.
+
+When configuring the client you should independently choose and fetch the CA certificates that you want to trust.
+Download the CA out of band and store it in a `Secret` or `ConfigMap` separate from the `Secret` containing the server's private key and certificate.
+
+This ensures that if the material in the `Secret` containing the server key and certificate is tampered with,
+the client will fail to connect to the compromised server.
+
+The same concept also applies when configuring a server for mutually-authenticated TLS;
+don't give the server access to Secret containing the client certificate and private key.
+
+{{% /alert %}}
+
 The `dnsNames` field specifies a list of [`Subject Alternative
 Names`](https://en.wikipedia.org/wiki/Subject_Alternative_Name) to be associated
 with the certificate.


content/en/docs/concepts/certificate.md
```diff
@@ -54,11 +54,44 @@ it is stored in `ca.crt`.
...
+   and store it in a Secret that it is independent of the Secret populated by cert-manager.
+
+This ensures that if the material in Secret containing the server certificates is modified,
+the client will fail to connect to the compromised server.
+
+
+**When configuring a server** to trust to a TLS client with a client certificate that is signed by a private CA,
+you will need access to the private CA certificate in order to verify the client.
+But if the server is running in a Kubernetes Pod,
+it should not mount the Secret containing the client key and certificate, because:
```
maelvls, line 82: "To trust to" seemed off but I might be wrong 😅

```suggestion
**When configuring a server** to trust a TLS client with a client certificate that is signed by a private CA,
you will need access to the private CA certificate in order to verify the client.
But if the server is running in a Kubernetes Pod,
it should not mount the Secret containing the client key and certificate, because:
```

content/en/docs/concepts/certificate.md
```diff
@@ -54,11 +54,44 @@ it is stored in `ca.crt`.
...
+
+1. That Secret also contains the private key, which should only be accessible to the client.
+   You should use RBAC to ensure that the Secret containing the client certificate and key are only accessible to Pods that need it.
+2. When configuring the server you should independently choose and fetch certificate of the CA that you want to trust,
+   download the  `ca.crt` out of band,
+   and store it in a Secret that it is independent of the client Secret populated by cert-manager.
```
maelvls, line 88:
```suggestion
   and store it in a Secret that is independent of the client Secret populated by cert-manager.
```

content/en/docs/concepts/certificate.md
```diff
@@ -54,11 +54,44 @@ it is stored in `ca.crt`.
...
+
+This ensures that if the material in Secret containing the client certificates is modified,
+the server will stop accepting connections from the compromised client.
```
maelvls, line 91: Would it be worth phrasing it as "tampered with" instead of "modified"?

maelvls (changes requested):
Thank you for this clarification! I now better understand the issues around "trust distribution" when using cert-manager Secrets for mTLS authentication.

maelvls:
@wallrj I'd love to see this merged, I was looking for this information and had to look for this PR today 😅

content/en/docs/concepts/certificate.md
```diff
@@ -62,6 +62,39 @@ which contains the following explanation:
 > assumption that the remote end must already possess it in order to
 > validate it in any case.
 
+{{% alert title="Use ca.crt with caution" color="warning" %}}
```
SgtCoDFish, line 65: suggestion: I like the general idea of this, but I think the two sections are so similar that they could be condensed down into one shorter block. I can see the reason for having two - mutual-TLS isn't symmetric - but I think in this case the concepts are simple enough to have an example for configuring a client to trust a server's cert, and then a line underneath with something like:

"The same concept also applies when configuring a server for mutually-authenticated TLS; don't use an actual Secret containing a client certificate."

content/en/docs/concepts/certificate.md
```diff
@@ -62,6 +62,39 @@ which contains the following explanation:
...
+**When configuring a client** to connect to a TLS server with a serving certificate that is signed by a private CA,
+you will need to provide the client with the CA certificate in order for it to verify the server.
+If the client is running in a Kubernetes Pod _DO NOT MOUNT THE SAME SECRET AS THE SERVER_,
+to access the `ca.crt` file, because:
+
+1. That Secret also contains the private key of the server, which should only be accessible to the server.
+   You should use RBAC to ensure that the Secret containing the serving certificate and private key are only accessible to Pods that need it.
+2. When configuring the client you should independently choose and fetch certificate of the CA that you want to trust.
+   Download the  `ca.crt` out of band,
+   and store it in a Secret that is independent of the Secret populated by cert-manager.
```
SgtCoDFish, line 76: suggestion: The paragraph before the list that this appears in ends:

> DO NOT MOUNT THE SAME SECRET AS THE SERVER, to access the `ca.crt` file, because:

so this list should contain _reasons_ not to do that, I think. As phrased here, I don't think this is really a reason not to use `ca.crt`; it's just describing the best practice. 

How about this, adding a second reason not to use `ca.crt` and moving what's currently in here 2 down below with some minor rewording:

```suggestion
2. Rotating CA certificates safely relies on being able to have both the old and new CA certificates trusted at the same time. By consuming the CA directly from the source, this isn't possible; you'll be _forced_ to have some down-time in order to to rotate certificates.

When configuring the client you should independently choose and fetch certificates that you want to trust. Download `ca.crt` out of band and store it in a Secret or ConfigMap, totally independent of the Secret containing the server's certificate.
```

content/en/docs/concepts/certificate.md
```diff
@@ -62,6 +62,39 @@ which contains the following explanation:
 > assumption that the remote end must already possess it in order to
 > validate it in any case.
 
+{{% alert title="Use ca.crt with caution" color="warning" %}}
+
+**When configuring a client** to connect to a TLS server with a serving certificate that is signed by a private CA,
+you will need to provide the client with the CA certificate in order for it to verify the server.
+If the client is running in a Kubernetes Pod _DO NOT MOUNT THE SAME SECRET AS THE SERVER_,
+to access the `ca.crt` file, because:
```
SgtCoDFish, line 70: nitpick: the commas don't look quite right to me for this on the preview. Also a minor rewording:

```suggestion
While `ca.crt` will likely contain the certificate you need to trust, _DO NOT MOUNT THE SAME SECRET AS THE SERVER_ so you can access `ca.crt`. This is because:
```

SgtCoDFish (changes requested):
Love the idea here, it's an important thing to highlight. Would it be worth also adding an FAQ entry for this and linking to this section from the FAQ?

I have some suggestions on this; what do you think?

SgtCoDFish (approved):
/lgtm

SgtCoDFish (approved):
/lgtm

(again :grin: )

Issue #426, opened 2021-02-05
Create a sequence diagram that shows how a certificate gets issued with let's encrypt


maelvls:
During the "cert-manager: 2020 retrospective" on 6th January 2021 (see [retro dashboard](https://action.parabol.co/team/hshPyuEuf)), we discussed the need for a diagram about the "internals" of cert-manager.

The goal is to represent a well-known flow (the creation of a Certificate using Let's Encrypt) and draw all the cert-manager CRDs and Secrets related to this flow.

I had drawn a quick diagram [a long time ago](https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1604673834216600):

<img alt="" src="https://user-images.githubusercontent.com/2195781/113146120-96706a80-922f-11eb-8348-bc9102656dc0.jpg" width="300">

I then started converting the drawing to a [textik diagram](https://textik.com/#7dc27a1faa9d56de0), but I soon realized it wouldn't be as practical as I intended it to be:

<img alt="" src="https://user-images.githubusercontent.com/2195781/107038034-3d9acd80-67bc-11eb-84f4-a208cd67a6b7.png" width="300">

As @coderanger suggested a month ago, there might be a better way with https://mermaid-js.github.io/mermaid/#/sequenceDiagram.

Irbe's high-level diagram: https://miro.com/app/board/o9J_lUdPqQI=/

Irbe's great [diagram](https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1617730470255500?thread_ts=1617703491.239900&cid=CDEQJ0Q8M):

<img alt="" src="https://user-images.githubusercontent.com/2195781/114370211-820d6580-9b7f-11eb-944e-37a2d3add419.jpg" width="300">

irbekrm:
> During the "cert-manager: 2020 retrospective" on 6th January 2021 (see retro dashboard), we discussed the need for a diagram about the "internals" of cert-manager.
I started drawing a diagram using the ASCII editor textik.com, but I soon realized it wouldn't be as practical as I intended it to be

Maybe this could be a cool talk (for example for cert-manager dev days) with a few diagrams, walk through the codebase etc.

maelvls:
Progress update: I abandonned the idea of using a sequence diagram, it just makes things even harder to understand. Here is what I have now:

![IMG_6515](https://user-images.githubusercontent.com/2195781/115728505-d89b5080-a384-11eb-94e7-1dc1f359a527.jpeg)

[...]

PR #582, opened 2021-06-03
Add the "life of a certificate" diagram

maelvls:
This PR is part of the work in https://github.com/cert-manager/website/issues/426. This diagram is not done, but I propose to release an initial version instead of waiting further.

| Preview | https://deploy-preview-582--cert-manager.netlify.app/docs/concepts/certificate/ |
|-|-|

cc @irbekrm


irbekrm:
Looks great to me already 🥳 🥳 

(I think the correct preview link might be https://deploy-preview-582--cert-manager.netlify.app/docs/concepts/certificate/)


A small nit- I think the `Order` is not removed immediately, but just by the time the next `Order` needs to be created?

Suggestion: 'requestmanager creates CertificateRequest with revision = "1" since its revision is nil' <- perhaps it would make sense to drop the 'since its revision is nil', I imagine it would be either obvious or confusing?

Another suggestion: I think the `Challenge`s are created from `Order.status.authorizations` fields, perhaps worth to move the arrows slightly to indicate this?

Agree that it might make sense to merge this as is though and I think the diagram is great- it should be super helpful to users already.
Please unhold to merge as is if you'd rather do that!

/hold
/approve
/lgtm

maelvls:
poke @irbekrm

irbekrm:
Thanks for the poke 😄 

I think we should merge this, it would give users a lot of value imo.

Thank you for working on this and making it so nice!

/lgtm

/hold cancel

PR #585, opened 2021-06-04
release-process: mention the code freeze, do not merge master after the first beta

maelvls:
Signed-off-by: Maël Valais <mael@vls.dev>

PR #586, opened 2021-06-04
Add a section about Certificate's renewal time calculation

maelvls:
Fixes a bug that is preventing me from merging https://github.com/cert-manager/website/pull/585#issuecomment-854832231. When this page was merged to release-next, it broke the build on master since these new links do not exist yet on `/docs`. They only exist in `/next-docs`.

To remediate that, we use a relative path.

cc @irbekrm


irbekrm:
Thank you very much for fixing this 🙏🏼 🙏🏼 

/approve
/lgtm

PR #588, opened 2021-06-07
Follow-up to #570 (2): Add a section about Certificate's renewal time calculation

maelvls:
This is a follow-up to #570. 

I propose to add a bit of air to the "Renewal" section. I also propose to disambiguate certificate vs. Certificate by pretending X.509 to every lower-case certificate.

| Preview | https://deploy-preview-588--cert-manager.netlify.app/docs/concepts/certificate |
|-|-|

cc @irbekrm

PR #590, opened 2021-06-08
Remove the spelling exceptions, they are not needed anymore since #588

maelvls:
The reason why I had to remove .spelling exceptions is that `./scripts/verify` clones from the branch release-next into the `/docs-next` folder, but the `.spelling` additions that were made to release-next branch aren't know by alex. That means any word added `.spelling` and merged to release-next will break `./script/verify` for any other branch.

In #588, I removed the need for the two .spelling additions. In this PR, I remove the words entirely since they are not needed anymore.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -246,8 +246,6 @@ prepended
 retweets
 JetstackHQ
 acme-dns
-andreas-p
-renewBefore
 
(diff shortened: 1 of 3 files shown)


maelvls:
Not needed anymore after #591 

/close

PR #599, opened 2021-06-10
Multiple selector types per block all need to match

### LINKED ISSUE #584: Improve the documentation for issuer.spec.acme.solvers.selector
See the conversation on Slack here https://kubernetes.slack.com/archives/C4NV3DWUC/p1622720926204400

Current docs [here](https://cert-manager.io/docs/configuration/acme/#all-together) seem to suggest that you can issue a cert whose DNS name does not match any of the DNS names from ` issuer.spec.acme.solvers[n].selector.dnsNames` but does match a zone from ` issuer.spec.acme.solvers[n].selector.dnsZones`.
Looking at the code [here](https://github.com/jetstack/cert-manager/blob/master/pkg/controller/acmeorders/util.go#L139) and having tried it myself that doesn't seem to be the case. 
We should make it clear that `issuer.spec.acme.solvers[n].selector.dnsNames` , ` issuer.spec.acme.solvers[n].selector.dnsZones` and ` issuer.spec.acme.solvers[n].selector.matchLabels` will all need to have match for a solver `n` to be selected for a challenge.

irbekrm:
This PR updates the doc for ACME issuer's solver selectors.

See this Slack conversation for context https://kubernetes.slack.com/archives/C4NV3DWUC/p1622720926204400

Current documentation suggest that if multiple selectors (dns names, zones, labels) are specified for a single selector block, the  solver will be chosen to solve a challenge if _either_ of the selectors from the block matches https://cert-manager.io/docs/configuration/acme/#all-together.

This doesn't seem to be the case- they would _all_ need to match, see the code [here](https://github.com/jetstack/cert-manager/blob/master/pkg/controller/acmeorders/util.go#L139).

Fixes #584 

### DIFF
--- a/content/en/docs/configuration/acme/_index.md
+++ b/content/en/docs/configuration/acme/_index.md
@@ -313,10 +313,11 @@ spec:
 #### All Together
 
 Each solver is able to have any number of the three selector types defined. In
-the following example, the `DNS01` solver will be used to solve challenges for
-domains for `Certificates` that contain the DNS names `a.example.com` and
-`b.example.com`, or for `test.example.com` and all of its subdomains
-(e.g. `foo.test.example.com`).
+the following example, the `DNS01` solver for CloudFlare will be used to solve
+challenges for domains for `Certificates` that contain the DNS names
+`a.example.com` and `b.example.com`. The `DNS01` solver for Google CloudDNS will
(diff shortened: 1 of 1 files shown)


jakexks (approved):
/lgtm

PR #619, opened 2021-06-21
Allow custom bind addresses for development servers

SgtCoDFish:
also remove incorrect comment and update copyright notice

Signed-off-by: Ashley Davis <ashley.davis@jetstack.io>

### DIFF
--- a/scripts/server
+++ b/scripts/server
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2019 The Jetstack cert-manager contributors.
+# Copyright 2019 The cert-manager contributors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


irbekrm (approved):
Looks good to me, I tried running the command with and without the `--bind` flag.

In practice, do you intend to pass a custom value via the flag if you want to listen on an external address?

SgtCoDFish:
> Looks good to me, I tried running the command with and without the `--bind` flag.
> 
> In practice, do you intend to pass a custom value via the flag if you want to listen on an external address?

Yeah, I was using this yesterday because my phone and work computer are both on the same network (my home network, via a VPN). So I did `export HUGO_BIND_ADDRESS="192.168.xxx.yyy"` and then could access that address on my phone for testing purposes.

PR #639, opened 2021-07-09
Fix kuard example WSOD

iambryancs:
When using `Exact` as `pathType`, static resources (and any other resource outside of the `/` path) will fail to load causing the page to display only a white screen.
The `Prefix` `pathType` has to be used.

Signed-off-by: Bryan CS <yanser25@gmail.com>

### DIFF
--- a/content/en/docs/tutorials/acme/example/ingress-tls-final.yaml
+++ b/content/en/docs/tutorials/acme/example/ingress-tls-final.yaml
@@ -16,7 +16,7 @@ spec:
     http:
       paths:
       - path: /
-        pathType: Exact
+        pathType: Prefix
         backend:
           service:
             name: kuard
--- a/content/en/docs/tutorials/acme/example/ingress-tls.yaml
+++ b/content/en/docs/tutorials/acme/example/ingress-tls.yaml
@@ -16,7 +16,7 @@ spec:
     http:
       paths:
       - path: /
-        pathType: Exact
+        pathType: Prefix
         backend:
           service:
             name: kuard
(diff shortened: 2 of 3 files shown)


irbekrm:
Thanks for spotting that and most importantly for PR-ing the fix 🙏🏼 !

There are 2 more ingress examples in that tutorial- would you like to apply the same fix for those? ([here](https://github.com/cert-manager/website/blob/07b89ebd3ff6e88d913a56ca3cfb0394da3ee812/content/en/docs/tutorials/acme/example/ingress-tls-final.yaml#L19) and [here](https://github.com/cert-manager/website/blob/07b89ebd3ff6e88d913a56ca3cfb0394da3ee812/content/en/docs/tutorials/acme/example/ingress-tls.yaml#L19)) 😅

iambryancs:
> Thanks for spotting that and most importantly for PR-ing the fix 🙏🏼 !
> 
> There are 2 more ingress examples in that tutorial- would you like to apply the same fix for those? ([here](https://github.com/cert-manager/website/blob/07b89ebd3ff6e88d913a56ca3cfb0394da3ee812/content/en/docs/tutorials/acme/example/ingress-tls-final.yaml#L19) and [here](https://github.com/cert-manager/website/blob/07b89ebd3ff6e88d913a56ca3cfb0394da3ee812/content/en/docs/tutorials/acme/example/ingress-tls.yaml#L19)) 😅

Sure no problem @irbekrm .

irbekrm (approved):
Thanks for the contribution @iambryancs !

/lgtm

PR #625, opened 2021-06-22
Document the use of the HTTP01 Gateway API solver

maelvls:
| Preview: https://deploy-preview-625--cert-manager-website.netlify.app/docs/configuration/acme/http01/#configuring-the-http01-gateway-api-solver |
|-|

This is the documentation for https://github.com/jetstack/cert-manager/pull/4276.

PR #640, opened 2021-07-09
Document the certificate shim support for Gateways

maelvls:
I started writing some of the details of how we support the Gateway API in this document:

| Preview | https://deploy-preview-640--cert-manager-website.netlify.app/docs/usage/gateway-api/ |
|-|-|


irbekrm:
Suggestion: I found the second diagram [here](https://gateway-api.sigs.k8s.io/) very useful in understanding what is sig-network's recommended way to configure TLS for infra that uses Gateways. Perhaps we could copy paste that into this doc (with a reference)? Not sure if the image is somewhere in GitHub.

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,248 @@
...
+    kind: Issuer
+    group: cert-manager.io
+  dnsNames:
+    - example.com
+    - www.example.com
+  secretName: example-com-tls
+```
+
+The only kind and group supported in `certificateRef` is `Secret` and `core`,
+and both fields are required for the creation of a Certificate shim:
```
irbekrm, line 154: Suggestion: add somewhere that the 'hostname' field too is required (and maybe also that the DNS name on `Certificate` will be the hostname from the listener- or maybe that's obvious).

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,248 @@
...
+title: "Securing Gateway Resources"
+linkTitle: "Securing Gateway Resources"
+weight: 90
+type: "docs"
+---
+
+Since 1.5, cert-manager supports requesting TLS certificates using annotations
+on Gateway resources. The Gateway resource is part of the [Gateway API][gwapi],
+a set of CRDs that can be installed to your Kubernetes cluster and that aim at
+offering an richer alternative to the Ingress abstraction.
```
irbekrm, line 11: Nit:
```suggestion
offering a richer alternative to the Ingress abstraction.
```

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,248 @@
...
+Gateway resource when the Gateway is configured to terminate the TLS connection.
+There is still uncertainty around how the HTTPRoute TLS configuration should be
+used (see
+[gateway-api#577](https://github.com/kubernetes-sigs/gateway-api/issues/577)),
+and no existing implementation supports the TLS passthrough mode in which the
+HTTPRoute contains the TLS configuration and terminates the connection (see
+[istio#31747](https://github.com/istio/istio/issues/31747)).
+
+The Gateway support for Certificate shims must be enabled manually using a flag.
+If you are using Helm:
```
irbekrm, line 30: Suggestion: perhaps worth here also linking how to install Gateway itself or would that be too much?
i.e
1. Install Gateway CRDs and controller
2. Install cert-manager with a flag ..

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,248 @@
...
+If you are using the raw cert-manager manifests, you can add the following like
+to the cert-manager controller Deployment `args`:
+
+```sh
+--controllers=*,gateway-shim
+```
+
+Like with an Ingress, the annotations `cert-manager.io/issuer` or
+`cert-manager.io/cluster-issuer` tells cert-manager that this Gateway should be
+looked at to create Certificate shims. For example, the following Gateway should
```
irbekrm, line 46: Suggestion: here and elsewhere, I think arguably you could call them just `Certificate`s instead of 'Certificate shims' to not confuse folks since they will then work in the same way as `Certificate`s.

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,248 @@
...
+    name: my-issuer
+    kind: Issuer
+    group: cert-manager.io
+  dnsNames:
+    - example.com
+  secretName: example-com-tls
+```
+
+Contrary to Ingresses, you can set the same secret name multiple times. Let us
+imagine that you have these two listeners:
```
irbekrm, line 93: I think these use case based descriptions are great! 
Perhaps worth to add a separate section 'Use cases' to this doc?
A few more I could think of 1) A Gateway with one 'valid' one 'invalid' listener block- a `Certificate` still gets created for the valid one. 2) A Gateway with two listeners each for different hostname and different secret name- two separate `Certificate`s get created for each. I guess eventually (but not for this PR?) it might make sense to describe this more from a user's perspective (i.e scenario based).

irbekrm (changes requested):
This doc looks great! I've added a couple nits, otherwise would be happy to approve unless you'd like to have more of a discussion on this?

content/en/docs/usage/gateway-api.md
maelvls: Agreed, a "Use cases" section makes sense. I added it and had two subsections under it:

- "Skipping of invalid listener blocks"
- "Two listeners with the same Secret name"

> A Gateway with two listeners each for different hostname and different secret name- two separate Certificates get created for each.

I realized that you can also choose to re-use both the same hostname and secret name, I added an example with that.

> I guess eventually (but not for this PR?) it might make sense to describe this more from a user's perspective (i.e scenario based).

I struggle finding "actual" user-facing scenarios due to my lack of "actual" use of the Gateway API. As soon as I have more knowledge, I'd be happy to open a new PR with actual user scenarios. 👍

maelvls:
I made the changes! You can now take another look.

/unassign
/assign @irbekrm

content/en/docs/usage/gateway-api.md
maelvls: Fixed in b8ba185.

content/en/docs/usage/gateway-api.md
maelvls: Good idea, "shim" is confusing.

content/en/docs/usage/gateway-api.md
maelvls: Good idea. Adding the instruction about installing the CRD also makes it clear that the Gateway API feature requires the installation of an external CRD, since people may think the Gateway API comes by default with Kubernetes.

content/en/docs/usage/gateway-api.md
maelvls: Good point.

- I added a table that shows all the requirements for a "valid" listener,
- I added a sentence + a comment in the YAML example to clearly show that the hostname field is used for setting the DNS name.

maelvls:
@irbekrm This PR is good to go. I'm not sure about the annotations yet though, e.g. the `gateway-class`, but I'll figure that out in the next PR.

irbekrm (changes requested):
Thanks @maelvls I have read through this doc and I think it's really good.
Thank you for linking https://github.com/kubernetes-sigs/gateway-api/issues/577 too, really good discussion there that I wasn't aware of.

I have added some _optional_ suggestions, mostly for all the `Ingress` references. I think for a reader of this it may not be strictly necessary to have a comparison with how the ingress-shim works for each use case- I imagine that their Gateway setup will differ from their Ingress setup + in the future there may be readers who have never even used the ingress-shim + folks who do use ingress-shim don't necessarily remember by heart everything about how it works- so it seems to add extra complexity. lmk if you agree.
Otherwise happy to lgtm this, it really looks good!

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,248 @@
+---
+title: "Securing Gateway Resources"
+linkTitle: "Securing Gateway Resources"
+weight: 90
+type: "docs"
+---
+
+Since 1.5, cert-manager supports requesting TLS certificates using annotations
+on Gateway resources. The Gateway resource is part of the [Gateway API][gwapi],
```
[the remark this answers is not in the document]
maelvls: Done in c890f89.

content/en/docs/usage/gateway-api.md
```diff
@@ -85,38 +101,49 @@ spec:
...
 
-Contrary to Ingresses, you can set the same secret name multiple times. Let us
-imagine that you have these two listeners:
+## Use cases
+
+### Skipping of invalid listener blocks
+
+Like with an Ingress, the invalid listeners are skipped. For example, no
+Certificate will be created for the following Gateway since both listener that
+does not have a `tls` block won't be used to create the Certificates.
```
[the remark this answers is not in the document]
maelvls: You are right, I shouln't even talk about "invalid" or "valid", it doesn't really help understanding. The only thing I should say is that you can expect cert-manager to create a Certificate whenever all the above requirements (in the table) are met. Otherwise, cert-manager will just skip this listener block.

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,248 @@
...
+  🔥 **TODO:** add an annotation `acme.cert-manager.io/http01-gateway-class` to
+  override the `gatewayClass` defined on the Issuer.
+
+- `acme.cert-manager.io/http01-edit-in-place` since this is specific to some
+  ingress controllers like ingress-gce.
+- `kubernetes.io/tls-acme`??
+
+  🔥 **TODO:** should we keep this annotation for the Gateway support?
+
+The following annotations are similar to the ones supported on the Ingress
```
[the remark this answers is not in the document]
maelvls: Very good idea, way clearer. Done in aa1ecfc18.

content/en/docs/usage/gateway-api.md
```diff
@@ -303,28 +306,14 @@ spec:
...
-not for this PR?) it might make sense to describe this more from a user's
-perspective (i.e scenario based).
-
 ## Supported Annotations
 
-Two Ingress annotations are not available on Gateways:
-
-- `acme.cert-manager.io/http01-ingress-class` since the Gateway resource does
-  not have a `class` field.
+Four Ingress annotations are not available on Gateways:
```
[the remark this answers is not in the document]
maelvls: Done in aa1ecfc18.

content/en/docs/usage/gateway-api.md
```diff
@@ -355,3 +344,9 @@ resource:
 - `cert-manager.io/usages`: (optional) this annotation allows you to configure
   `spec.usages` field for the `Certificate` to be generated. Pass a string with
   comma-separated values i.e "key agreement,digital signature, server auth"
+
+One annotation is specific to the Gateway resource:
```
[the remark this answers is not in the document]
maelvls: Done in aa1ecfc18.

content/en/docs/usage/gateway-api.md
```diff
@@ -145,62 +173,143 @@ spec:
...
+|----------------------------|-------------------------------------------------------------|
+| `tls.hostname`             | Must not be empty.                                          |
+| `tls.mode`                 | Must be set to `Terminate`. `Passthrough` is not supported. |
+| `tls.certificateRef.name`  | Cannot be left empty.                                       |
+| `tls.certificateRef.kind`  | Must be set to `Secret`.                                    |
+| `tls.certificateRef.group` | Must be set to `core`.                                      |
+
+### Two listeners with the same Secret name
+
+Contrary to with an Ingress, you can set the same Secret name multiple times
```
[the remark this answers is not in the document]
maelvls: Done in aa1ecfc18.

content/en/docs/usage/gateway-api.md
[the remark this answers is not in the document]
maelvls: Done in aa1ecfc18.

content/en/docs/usage/gateway-api.md
```diff
@@ -85,38 +101,49 @@ spec:
...
-    - example.com
+    - example.com # ✅ Copied from the `hostname` field.
   secretName: example-com-tls
 ```
 
-Contrary to Ingresses, you can set the same secret name multiple times. Let us
-imagine that you have these two listeners:
+## Use cases
+
+### Skipping of invalid listener blocks
```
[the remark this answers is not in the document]
maelvls: Done in aa1ecfc18.

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+If you are using the raw cert-manager manifests, you can add the following like
+to the cert-manager controller Deployment `args`:
+
+```sh
+--controllers=*,gateway-shim
+```
+
+Like with an Ingress, the annotations `cert-manager.io/issuer` or
```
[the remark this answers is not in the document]
maelvls: Done in aa1ecfc18.

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,347 @@
+---
+title: "Securing Gateway Resources"
+linkTitle: "Securing Gateway Resources"
+weight: 90
```
JoshVanL, line 4: Should we change the weight, so that it sits next to the other ingress page in the list?

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,347 @@
...
+terminates the connection (see the [feature request for supporting `spec.tls` in
+HTTPRoute for Istio][istio#31747]).
+
+[istio#31747]: https://github.com/istio/istio/issues/31747
+[gateway-api#577]: https://github.com/kubernetes-sigs/gateway-api/issues/577
+
+Before enabling the Gateway support, you will have to install the Gateway API
+CRDs:
+
+```sh
```
JoshVanL, line 39:
```suggestion
```terminal
```

So we can have a copy button.

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,347 @@
...
+kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.3.0" | kubectl apply -f -
+```
+
+You then need to manually enable the Gateway support for Certificate using a
+flag:
+
+- If you are using Helm:
+
+  ```sh
+  helm upgrade --install cert-manager jetstack/cert-manager
```
JoshVanL, line 49: This is might be better on a single line for copy and pasting, I had issues myself.

Need to include the namespace flag.

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,347 @@
...
+  argument to the cert-manager controller Deployment's `args`:
+
+  ```yaml
+  args:
+    # ...
+    - --controllers=*,gateway-shim
+  ```
+
+You can also install the CRDs after installing cert-manager. cert-manager will
+pick up the Gateway CRD as soon as it is installed.
```
JoshVanL, line 63: Is this true? I would have thought the controller needs to be restarted. How are we doing this?

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,347 @@
...
+
+The annotations `cert-manager.io/issuer` or `cert-manager.io/cluster-issuer`
+tells cert-manager that this Gateway should be looked at to create Certificate.
```
JoshVanL, line 66:
```suggestion
tells cert-manager that this Gateway should be looked at to create a Certificate.
```

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,347 @@
...
+        kind: HTTPRoute
+        selector:
+          matchLabels:
+            app: foo
+      tls:
+        mode: Terminate
+        certificateRef:
+          name: example-com-tls
+          kind: Secret
+          group: core
```
JoshVanL, line 93: Will I be able to reference a Certificate at some point?

```suggestion
        certificateRef:
          name: example-com-tls
          kind: Certificate 
          group: cert-manager.io
```

JoshVanL: I guess this is up to the implementations of the Gatway API projects :)

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,347 @@
...
+- `cert-manager.io/issuer`: the name of an `Issuer` to acquire the certificate
+  required for this Gateway. The Issuer _must_ be in the same namespace as the
+  Gateway resource.
+
+- `cert-manager.io/cluster-issuer`: the name of a `ClusterIssuer` to acquire the
+  certificate required for this Gateway. It does not matter which namespace
+  your Gateway resides, as `ClusterIssuers` are non-namespaced resources.
+
+- `cert-manager.io/issuer-kind`: the name of an external `Issuer`
+  controller's `CustomResourceDefinition` (only necessary for out-of-tree `Issuers`)
```
JoshVanL, line 331: This is wrong. It is the _Kind_ of the issuer resource type. `issuer-name` is the name of the resource.

`cert-manager.io/issuer=foo` is a short hand for `cert-manager.io/issuer-name=foo` `cert-manager.io/issuer-kind=Issuer`


Please add some wording around this distinction and that one is short hand for the other. This trips some users up when using external issuers.

JoshVanL (changes requested):
Some comments from me.

/assign @maelvls

content/en/docs/usage/gateway-api.md
maelvls: Do you mean the copy button that appears on GitHub-rendered pages such as READMEs?

Screenshot from 2021-07-30 21-02-34

I don't think there is such a feature using Docsy.

content/en/docs/usage/gateway-api.md
maelvls: Thanks! I changed the weight to 105 and now it shows right after the "Securing Ingress resources":

content/en/docs/usage/gateway-api.md
maelvls: Not sure anymore. At the moment, we are discovering whether the CRDs are present or not, but the watchers still keep watching. It goes like this (see [controller.go](https://github.com/jetstack/cert-manager/pull/4276/files)):

```go
	// check if the gateway API CRDs are available
	var gatewayAvailable bool
	d := cl.Discovery()
	resources, err := d.ServerResourcesForGroupVersion(gwapi.GroupVersion.String())
	switch {
	case apierrors.IsNotFound(err):
		gatewayAvailable = false
		log.Info("the Gateway API CRDs do not seem to be present, cert-manager will keep retrying watching for them")
	case err != nil:
		return nil, nil, fmt.Errorf("while checking if the Gateway API CRD is installed: %s", err.Error())
	case len(resources.APIResources) == 0:
		gatewayAvailable = false
		log.Info("the Gateway API CRDs do not seem to be present, cert-manager will keep retrying watching for them")
	default:
		gatewayAvailable = true
	}
```

I'll remove this line since it feels quite tricky to get this "crd-hot-reloading" mechanism right, and it might not be worth it.

cc @irbekrm @jakexks

content/en/docs/usage/gateway-api.md
maelvls: Yes. At the moment, `core/Secret` is the only resource that is supported across all Gateway API implementations. I read somewhere that [GKE Gateway Controller](https://cloud.google.com/kubernetes-engine/docs/concepts/gateway-api) plans on supporting another resource but I can't find where I read that.

content/en/docs/usage/gateway-api.md
maelvls: The description seems correct, albeit confusing. `issuer-kind` _is_ the name of an external Issuer's CRD.

What do you think about this:

```suggestion
- `cert-manager.io/issuer-kind`: the kind of the external issuer resource, for example `AWSPCACIssuer`. This is only necessary for out-of-tree issuers.
```

content/en/docs/usage/gateway-api.md
JoshVanL: I see, taking a look at the code it looks like “ cert-manager will keep retrying watching for them” is incorrect. Unless I’ve missed some mechanism, we would need to setup some kind of loop to keep trying to see if they’re available.

IMO, we should _not_ do this. If a user wants to use this API, it is expected behaviour that we would error exit the program if they are not available. “Silently” continuing is bad behaviour I think.

This is inline with how all other operators+CRDs behave, especially in cert-manager.

maelvls: From these past conversations:

1. https://github.com/jetstack/cert-manager/pull/4158#discussion_r669380637
2. https://github.com/jetstack/cert-manager/pull/4276#discussion_r679364495
3. https://github.com/jetstack/cert-manager/pull/4276#discussion_r680174007

There are two separate concerns that we have been trying to figure out:

### Concern 1: how do we turn the feature on?

Should we discover the Gateway API, or should we use gate the feature behind flags such as `--controllers=*,gateway-shim` and `--controllers=*,gateway-http01`?

→ At first, we went with a flag (`--controllers=*,gateway-shim`). @jakexks then went instead with discovering after the discussion in https://github.com/jetstack/cert-manager/pull/4276#discussion_r679364495.


### Concern 2: how do we handle the fact that Gateway API is not a core API?

When cert-manager is installed before the CRDs are installed, should we retry?

→ In https://github.com/jetstack/cert-manager/pull/4158#discussion_r669380637, @irbekrm made the point that since Gateway API is not part of core Kubernetes we shouldn't crash: 

> Can we avoid crashing here? Because Gateway CRDs are not part of the core API, I imagine they could not yet be installed at this point, depending on users' workflow and it may be a better user experience if the pod didn't crash upfront if they forget to install the gateway controller first. Or, would this just cause issues down the line? Alternatively, if we keep this crash, perhaps the error message could be like gateway-shim controller is enabled, but couldn't discover ... to make debugging easier etc.

JoshVanL: This all makes sense to me!

My concern, as @irbekrm mentioned, is how can we avoid the case where a user installs things in the wrong order, and then is confused as to why nothing is working. I agree, this _will_ become GitHub issues and slack messages down the road.

At least a pod crashing is a clear indicator that something is wrong and a user won’t miss anything.

maelvls: Here is how it works currently:

**Gateway shim:**

- To be able to use the gateway shim feature, users have to pass `--controllers=*,gateway-shim`.
- When cert-manager is started with the gateway shim feature turned on but the Gateway API CRDs are not yet installed, the user will be warned that the CRD should be installed.
- The user can install the CRD after starting cert-manager, and cert-manager will pick it up and the gateway shim will start working.

**Gateway solver:**

- To be able to use the gateway solver, users need to install the Gateway API CRDs,
- When cert-manager is started without the CRDs installed, cert-manager says that the Gateway HTTP-01 solver support is disabled due to the CRDs no being installed,
- In order to install the CRDs after cert-manager is running, you have to restart cert-manager to make the gateway solver feature start working.

maelvls: @wallrj made a good point during this morning's standup: why not turn on/off both features simultaneously with a feature gate, e.g.:

```
--feature-gates=GatewayAPI=true
```

Users would still have to restart the controllers if the CRDs are installed when cert-manager is already running to start the "Gateway solver" feature (the "Gateway shim" feature would pick up the CRDs when they are installed).

That said, we still decided to go with the current implementation (described [above](https://github.com/cert-manager/website/pull/640#discussion_r682481309)) for releasing 1.5.0-beta.0.

maelvls:
@JoshVanL Do you think you can take another look?

jakexks:
/assign

I'll give it a read

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,332 @@
...
+title: "Securing Gateway Resources"
+linkTitle: "Securing Gateway Resources"
+weight: 105
+type: "docs"
+---
+
+Since 1.5, cert-manager supports requesting TLS certificates using annotations
+on Gateway resources. This works similarly as to what you can do with
+annotations on the Ingress resource, as described on the page [Securing Ingress
+Resources](/docs/usage/ingress/).
```
wallrj, line 11:
```suggestion
 **FEATURE STATE**: cert-manager v1.5 [stable] 

cert-manager can generate TLS certificates for Gateway resources.
This is configured by adding annotations to a Gateway and is similar to the process for [Securing Ingress
Resources](/docs/usage/ingress/).
```

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,332 @@
...
+
+The Gateway resource is part of the [Gateway API][gwapi], a set of CRDs that can
+be installed to your Kubernetes cluster and that aim at offering a richer
+alternative to the Ingress abstraction.
```
wallrj, line 15:
```suggestion
The Gateway resource is part of the [Gateway API][gwapi], a set of CRDs that you install on your Kubernetes cluster and which provide various improvements over the Ingress API.
```

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,332 @@
...
+Before enabling the Gateway support, you will have to install the Gateway API
+CRDs:
+
+```sh
+kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.3.0" | kubectl apply -f -
+```
+
+cert-manager will detect that the Gateway API CRD is installed and will start
+watching Gateway resources. The Gateway API CRDs must be installed before
+cert-manager starts.
```
wallrj, line 45: I still find this paragraph confusing, but accurate.
```suggestion
The Gateway API CRDs should either be installed before
cert-manager starts or the cert-manager Deployment should be restarted after installing the Gateway API CRDs.
This is important because some of the cert-manager components only perform the Gateway API check on startup.
```

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,332 @@
...
+
+
+The annotations `cert-manager.io/issuer` or `cert-manager.io/cluster-issuer`
+tells cert-manager that this Gateway should be looked at to create a
+Certificate. For example, the following Gateway should trigger the creation of
+the `example-com-tls` Certificate:
```
wallrj, line 51:
```suggestion
The annotations `cert-manager.io/issuer` or `cert-manager.io/cluster-issuer`
tell cert-manager  to create a
Certificate for a Gateway. For example, the following Gateway will trigger the creation of
a Certificate with the name `example-com-tls`:
```

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,332 @@
...
+        mode: Terminate
+        certificateRef:
+          name: example-com-tls
+          kind: Secret
+          group: core
+```
+
+A few moments later, cert-manager should create a Certificate. The Certificate
+is named after the Secret name `example-com-tls`. The `dnsNames` field is set
+with the above `hostname` field.
```
wallrj, line 81:
```suggestion
A few moments later, cert-manager will create a Certificate. The Certificate
is named after the Secret name `example-com-tls`. The `dnsNames` field is set
with the `hostname` field from the Gateway spec.
```

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,332 @@
...
+|           Field            |                         Requirement                         |
+|----------------------------|-------------------------------------------------------------|
+| `tls.hostname`             | Must not be empty.                                          |
+| `tls.mode`                 | Must be set to `Terminate`. `Passthrough` is not supported. |
+| `tls.certificateRef.name`  | Cannot be left empty.                                       |
+| `tls.certificateRef.kind`  | Must be set to `Secret`.                                    |
+| `tls.certificateRef.group` | Must be set to `core`.                                      |
+
+In the following example, the three first listener blocks are not going to be
+used to generate Certificate resources:
```
wallrj, line 115:
```suggestion
In the following example, the first three listener blocks will not be
used to generate Certificate resources:
```

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,332 @@
...
+    group: cert-manager.io
+  dnsNames:
+    - foo.example.com
+  secretName: example-com-tls
+```
+
+### Two listeners with the same Secret name
+
+The same Secret name can be re-used in multiple TLS blocks regardless of the
+hostname. Let us imagine that you have these two listeners:
```
wallrj, line 182:
```suggestion
The same Secret name can be re-used in multiple TLS blocks, regardless of the
hostname. Let us imagine that you have these two listeners:
```

content/en/docs/usage/gateway-api.md
```diff
@@ -0,0 +1,332 @@
...
+  secretName: site-org-tls
+```
+
+## Supported Annotations
+
+If you are migrating to using Gateway resources instead of Ingress resources to
+generate Certificate resources, be aware that there are some differences in
+which annotations are [supported on the Ingress
+resource](https://cert-manager.io/docs/usage/ingress/#supported-annotations)
+versus which annotations are supported on the Gateway resource.
```
wallrj, line 301:
```suggestion
If you are migrating to Gateway resources from Ingress resources, be aware that there are some differences between [the annotations for Ingress resources](https://cert-manager.io/docs/usage/ingress/#supported-annotations)
versus the annotations for Gateway resources.
```

wallrj (approved):
I left a few suggestions, but ignore or adapt them as you please.

/lgtm
/hold

content/en/docs/usage/gateway-api.md
maelvls: I like your idea!

I'll go with 1.5 instead of v1.5 as per https://cert-manager.io/docs/installation/supported-releases/#terminology (1.5 is the release, v1.5 is a tag)

```md
**FEATURE STATE**: cert-manager 1.5 [stable]
                                ^^^
```

maelvls: (I re-read the "Terminology" section and it is quite confusing... like release vs. version, patch release vs. patch release...)

maelvls:
Thank you @wallrj for the review! I added your suggestions.

/unassign
/assign @wallrj

wallrj (approved):
Nice docs @maelvls 

Love the diagram.

/lgtm
/approve

maelvls:
The discussion around discovery vs. feature gate is not done:

- https://github.com/cert-manager/website/pull/640#discussion_r682529517

We plan on adding a feature gate for `1.5.0-beta.1`, e.g.

```
--feature-gates=GatewayAPI=true
```

Currently, the documentation says

```
FEATURE STATE: cert-manager 1.5 [stable]
```

I will fix this in #625 before we release `1.5.0-beta.1`.

/unhold

PR #646, opened 2021-07-16
Release-process: I did not know how to update the release branch

maelvls:
Here is a couple of nits that I found while releasing 1.4.0-alpha.0. I added you must have the `admin` Github permission to do the whole process.

### DIFF
--- a/content/en/docs/contributing/release-process.md
+++ b/content/en/docs/contributing/release-process.md
@@ -138,14 +138,14 @@ page if a step is missing or if it is outdated.
 1. Make sure to note which type of release you are doing. That will be helpful
    in the next steps.
 
-    | Type of release          | Example of git tag |
-    | ------------------------ | ------------------ |
-    | initial alpha release    | `v1.3.0-alpha.0`   |
-    | subsequent alpha release | `v1.3.0-alpha.1`   |
-    | initial beta release     | `v1.3.0-beta.0`    |
-    | subsequent beta release  | `v1.3.0-beta.1`    |
-    | final release            | `v1.3.0`           |
-    | patch release            | `v1.3.1`           |
+    |          Type of release           | Example of git tag |
+    |------------------------------------|--------------------|
+    | initial alpha release              | `v1.3.0-alpha.0`   |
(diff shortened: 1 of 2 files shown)


maelvls:
@SgtCoDFish I realized that what was written in the "Supported releases" was off; it said that we wait until the final release to do the patch releases for critical bugs, but we don't wait (and I don't think we should wait!)

I removed the diagram since it does not make sense anymore.

content/en/docs/installation/supported-releases.md
```diff
@@ -138,47 +138,18 @@ them.
 
 **Critical bugs** include both regression bugs as well as upgrade bugs.
 
-Regressions are functionalities that worked in a previous release but no
-longer work. [#3393][] and [#2857][] are two examples of regressions.
+Regressions are functionalities that worked in a previous release but no longer
+work. [#4142][], [#3393][] and [#2857][] are three examples of regressions.
```
SgtCoDFish, line 142: praise: good thinking to add this here, it's a good example :+1:

content/en/docs/contributing/release-process.md
```diff
@@ -155,16 +156,19 @@ page if a step is missing or if it is outdated.
...
+      release). Update the release branch with the latest commits from the
+      master branch, as follows:
 
        ```bash
        # Must be run from the cert-manager repo folder.
        git fetch --all
        git branch --force release-1.0 origin/release-1.0
        git checkout release-1.0
-       git merge --ff-only origin/master # don't run for a point release!
+       git merge --ff-only origin/master
```
SgtCoDFish, line 169: question: was it intentional to remove this comment?

SgtCoDFish (commented):
just one question from me :smile:

content/en/docs/contributing/release-process.md
maelvls: Good catch, that is unintentional 😅

(I noticed that right above it says "initial alpha, subsequent alpha and initial beta" but I realize that it is good to remind people "only for point release" since this command should never be used for patch releases)

PR #650, opened 2021-07-20
Adjust list of Subject Alternative Names

aazon:
A description of resource `Certificate` has `foo.example.com` and `bar.example.com` in the `dnsNames` section, which is different from the diagram that has `example.com` and `foo.example.com` domains. When I was looking at the schema in the article, I thought that the part related to `bar.example.com` is to follow, but it didn't and therefore was a bit confusing.

### DIFF
--- a/content/en/docs/concepts/certificate.md
+++ b/content/en/docs/concepts/certificate.md
@@ -25,8 +25,8 @@ metadata:
 spec:
   secretName: acme-crt-secret
   dnsNames:
+  - example.com
   - foo.example.com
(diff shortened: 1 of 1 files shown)


maelvls:
Thank you @aazon! I realize that I should have update the certificate examples when I added the diagram 😅

/ok-to-test

maelvls:
I rebased the commit and merged it in b11b7747c51c998ff8982d5eb5bb6339af512dff.

Thank you for the patch!

/close

PR #661, opened 2021-07-28
FAQ Tweaks, Including Root Chain FAQ + Terminology

SgtCoDFish:
:rotating_light: Best reviewed on a per-commit basis, each commit is self-contained.

The root chain FAQ was suggested [here](https://github.com/jetstack/cert-manager/pull/4261#pullrequestreview-716984217)

I figured it would also be useful to add a page explaining some of the terminology used in TLS, too - especially terms which are frequently misused or can cause confusion.

### DIFF
--- a/content/en/docs/faq/_index.md
+++ b/content/en/docs/faq/_index.md
@@ -20,6 +20,32 @@ face:
 
 This is a feature in cert-manager starting in `v0.16` using the kubectl plugin. More information can be found on [the renew command's page](../usage/kubectl-plugin/#renew)
 
+### Why isn't my root certificate in my issued Secret's `tls.crt`?
+
+Occasionally, people work with systems which have made a flawed choice regarding TLS chains. The [TLS spec](https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.2)
+has the following section for the "Server Certificate" section of the TLS handshake:
+
+> This is a sequence (chain) of certificates.  The sender's
+> certificate MUST come first in the list.  Each following
+> certificate MUST directly certify the one preceding it.  Because
+> certificate validation requires that root keys be distributed
+> independently, the self-signed certificate that specifies the root
+> certificate authority MAY be omitted from the chain, under the
+> assumption that the remote end must already possess it in order to
+> validate it in any case.
+
+In a standard, secure and correctly configured TLS environment, adding a root certificate to the chain is almost entirely _pure waste_.
+
+There are two ways that a certificate can be trusted:
+
+- explicitly, by including it in a trust store.
+- through a signature, by following the certificate's chain back up to an explicitly trusted certificate.
+
+Crucially, root certificates are by definition self-signed and they cannot be validated through a signature.
+
+As such, if we have a client trying to validate the certificate chain sent by the server, the client must already have the
+root before the connection is started. If the client already has the root, there was no point in it being sent by the server!
+
 ### How can I see all the historic events related to a certificate object ?
 
 cert-manager publishes all events to the Kubernetes events mechanism, you can get the events for your specific resources using `kubectl describe <resource> <name>`.
--- a/content/en/docs/faq/_index.md
+++ b/content/en/docs/faq/_index.md
@@ -34,7 +34,8 @@ has the following section for the "Server Certificate" section of the TLS handsh
 > assumption that the remote end must already possess it in order to
 > validate it in any case.
 
-In a standard, secure and correctly configured TLS environment, adding a root certificate to the chain is almost entirely _pure waste_.
+In a standard, secure and correctly configured TLS environment, adding a root certificate to the chain is
+almost always unnecessary and wasteful.
 
 There are two ways that a certificate can be trusted:
 
@@ -46,19 +47,27 @@ Crucially, root certificates are by definition self-signed and they cannot be va
 As such, if we have a client trying to validate the certificate chain sent by the server, the client must already have the
 root before the connection is started. If the client already has the root, there was no point in it being sent by the server!
 
-### How can I see all the historic events related to a certificate object ?
+The same logic with not sending root certificates applies for servers trying to validate client certificates;
+the [same justification](https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.6) is given in the TLS RFC.
+
+### How can I see all the historic events related to a certificate object?
 
 cert-manager publishes all events to the Kubernetes events mechanism, you can get the events for your specific resources using `kubectl describe <resource> <name>`.
+
 Due to the nature of the Kubernetes event mechanism these will be purged after a while. If you're using a dedicated logging system it might be able or is already also storing Kubernetes events.
 
-### What happens if a renewal is doesn't happen due to issues? Will it be tried again after sometime?
+### What happens if a renewal doesn't happen? Will it be tried again after some time?
 
-cert-manager makes use of exponential back off to retry non-fatal failures (ones that didn't mark the `CertificateRequest` as failed). If the `CertificateRequest` was marked as failed, issuance will be re-tried in 1 hour.
+cert-manager will retry renewal if it encounters temporary failures. It uses an exponential backoff algorithm to calculate the delay between each retry.
+
+A temporary failure is one that doesn't mark the `CertificateRequest` as failed. If the `CertificateRequest` is marked as failed, issuance will be re-tried in 1 hour.
 
 ### Is ECC (elliptic-curve cryptography) supported?
 
 cert-manager supports ECDSA key pairs! You can set your certificate to use ECDSA  in the `privateKey` part of your Certificate resource.
+
 For example:
+
 ```yaml
 apiVersion: cert-manager.io/v1
 kind: Certificate
@@ -77,6 +86,7 @@ spec:
 ```
 
 ### If `renewBefore` or `duration` is not defined, what will be the default value?
+
 Default `duration` is [90 days](https://github.com/jetstack/cert-manager/blob/v1.2.0/pkg/apis/certmanager/v1/const.go#L26). If `renewBefore` has not been set, `Certificate` will be renewed 2/3 through its _actual_ duration.
 
 ## Miscellaneous
--- a/content/en/docs/faq/_index.md
+++ b/content/en/docs/faq/_index.md
@@ -8,7 +8,7 @@ type: "docs"
 Below is an aggregation of solutions to some issues that cert-manager users may
 face:
 
-
+- [TLS Terminology, including commonly misused terms](./terminology/)
 - [Troubleshooting issuing ACME certificates](./acme/)
 - [How to change the Cluster Resource Namespace](./cluster-resource/)
 - [How to sync secrets across namespaces](./kubed/)
--- a/content/en/docs/faq/terminology.md
+++ b/content/en/docs/faq/terminology.md
@@ -1,0 +1,72 @@
+---
+title: "TLS Terminology"
+linkTitle: "TLS Terminology"
+weight: 50
+type: "docs"
+---
+
+With TLS being so widely deployed, terminology can sometimes get confused or be used to mean different things, and that reality
+combined with the complexity of TLS can lead to serious misunderstandings and confusion.
+
+For further reference, you might want to check out some relevant RFCs:
+
+- [RFC 5246: TLS 1.2](https://datatracker.ietf.org/doc/html/rfc5246)
+- [RFC 8446: TLS 1.3](https://datatracker.ietf.org/doc/html/rfc8446)
+- [RFC 5280: X.509](https://datatracker.ietf.org/doc/html/rfc5280)
+
+### What does "publicly trusted" mean?
+
+Broadly speaking, a "publicly trusted" certificate is one that you can use on the Internet and expect
+that most reasonably up-to-date computers will be able to verify it using their system trust store.
+
+There isn't a single standard trust store containing certs which are "publicly trusted", but generally most
+of the commonly seen trust stores are similar. An example would be [Mozilla's CA Certificate Program](https://wiki.mozilla.org/CA).
+
+### What does "self-signed" mean? Is my CA self-signed?
+
+Self-signed means exactly what it says; a certificate is self-signed if it is signed by its own private key.
+
+Self-signed is a commonly confused term, however, and is very frequently misused to mean "not publicly trusted". We tend to use terms
+like "private PKI" to denote the situation where an organization might have their own internal CA certificates which wouldn't
+be trusted outside of the organization.
+
+As an example, there are _many_ self-signed certificates in [Mozilla's CA Certificate Program](https://wiki.mozilla.org/CA), but
+all of those certificates would usually be described as "publicly trusted".
+
+Your certificate is self-signed only if it's signed with its own key.
+
+### What's the difference between "root", "intermediate", and "leaf" certificates?
+
+cert-manager uses the following definitions:
+
+#### Root Certificates
+
+Roots are self-signed certificates and almost always marked as CA certificates. They're usually not sent over the wire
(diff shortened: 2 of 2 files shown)


content/en/docs/faq/terminology.md
```diff
@@ -0,0 +1,65 @@
+---
+title: "TLS Terminology"
+linkTitle: "TLS Terminology"
+weight: 50
+type: "docs"
+---
+
+With TLS being so widely deployed, terminology can sometimes get confused or be used to mean different things, and that reality
+combined with the complexity of TLS can lead to serious misunderstandings and confusion.
+
```
irbekrm, line 10: Suggestion: we already linked it elsewhere, but maybe worth to link the TLS spec here too? https://datatracker.ietf.org/doc/html/rfc5246

irbekrm (approved):
This looks fantastic to me!

Thanks for the terminology section too.

I'd be happy to lgtm, but perhaps other folks want to take a look too.

content/en/docs/faq/terminology.md
```diff
@@ -0,0 +1,65 @@
...
+### What's the difference between "root", "intermediate", and "leaf" certificates?
+
+cert-manager uses the following definitions:
+
+#### Root Certificates
+
+Roots are self-signed certificates and almost always marked as CA certificates. They're usually not sent over the wire
+during a TLS handshake because they need to be explicitly trusted in order to be validated.
+
+Roots are sometimes defined as "CA certificates which are explicitly trusted" - which can include certificates which
```
maelvls, line 41: suggestion: em dash 👼

```suggestion
Roots are sometimes defined as "CA certificates which are explicitly trusted"---which can include certificates which
```

content/en/docs/faq/terminology.md
```diff
@@ -0,0 +1,65 @@
...
+aren't self-signed. cert-manager doesn't use this definition.
+
+Changing trust stores to include new roots or remove old ones is a non-trivial task which can take months or years for publicly
+trusted roots. For this reason, roots are usually issued with very long lifetimes - often on the order of decades.
```
maelvls, line 45:
```suggestion
trusted roots. For this reason, roots are usually issued with very long lifetimes—often on the order of decades.
```

content/en/docs/faq/terminology.md
```diff
@@ -0,0 +1,65 @@
...
+### What does "self-signed" mean? Is my CA self-signed?
+
+Self-signed means exactly what it says; a certificate is self-signed if it is signed by its own private key.
+
+Self-signed is a commonly confused term, however, and is very frequently misused to mean "not publicly trusted". We tend to use terms
+like "private PKI" to denote the situation where an organization might have their own internal CA certificates which wouldn't
+be trusted outside of the organization.
+
+As an example, there are _many_ self-signed certificates in [Mozilla's CA Certificate Program](https://wiki.mozilla.org/CA), but
+all of those certificates would usually be described as "publicly trusted".
```
maelvls, line 28: praise: thank you for giving the mozilla CA bundle example. This is great for newcomers!

content/en/docs/faq/terminology.md
```diff
@@ -0,0 +1,65 @@
...
+Your certificate is self-signed only if it's signed with its own key.
+
+### What's the difference between "root", "intermediate", and "leaf" certificates?
+
+cert-manager uses the following definitions:
+
+#### Root Certificates
+
+Roots are self-signed certificates and almost always marked as CA certificates. They're usually not sent over the wire
+during a TLS handshake because they need to be explicitly trusted in order to be validated.
```
maelvls, line 39: random question: why do we say "trust root"? I assume that we mean "the root of what we trust", but shouldn't we rather say "the trusted root" or "the root that we trust" instead?

content/en/docs/faq/terminology.md
maelvls, line 45: suiggestion: we could mention the case of the [Let's Encrypt DST root CA X3](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/) intermediate which expired in Sept 2021 and broke many old Android phones that were not being updated anymore; these phones cannot access some website due to this

A nice example of "long lifetime" that still broke things 😅

maelvls (approved):
@SgtCoDFish Thank you for writing this, it was a pleasure to read. This is brilliant! 👏

content/en/docs/faq/_index.md
```diff
@@ -20,19 +20,50 @@ face:
...
+
+There are two ways that a certificate can be trusted:
+
+- explicitly, by including it in a trust store.
+- through a signature, by following the certificate's chain back up to an explicitly trusted certificate.
+
+Crucially, root certificates are by definition self-signed and they cannot be validated through a signature.
+
+As such, if we have a client trying to validate the certificate chain sent by the server, the client must already have the
+root before the connection is started. If the client already has the root, there was no point in it being sent by the server!
```
wallrj, line 47: Should we also say something about hypothetical client mounting tls.crt as a client-certificate? Can we link to a corresponding section of the spec that says whether the client MAY / SHOULD / MUST / or MUST NOT send the full chain to the server?

content/en/docs/faq/_index.md
```diff
@@ -20,19 +20,50 @@ face:
...
+### How can I see all the historic events related to a certificate object?
 
 cert-manager publishes all events to the Kubernetes events mechanism, you can get the events for your specific resources using `kubectl describe <resource> <name>`.
+
 Due to the nature of the Kubernetes event mechanism these will be purged after a while. If you're using a dedicated logging system it might be able or is already also storing Kubernetes events.
 
-### What happens if a renewal is doesn't happen due to issues? Will it be tried again after sometime?
+### What happens if a renewal doesn't happen? Will it be tried again after some time?
+
+cert-manager makes use of exponential back off to retry non-fatal failures.
```
wallrj, line 57:
```suggestion
cert-manager will retry the renewal if it encounters temporary failures and it uses an exponential backoff algorithm to calculate the delay between each retry.
```

content/en/docs/faq/_index.md
```diff
@@ -20,19 +20,50 @@ face:
...
 
-cert-manager makes use of exponential back off to retry non-fatal failures (ones that didn't mark the `CertificateRequest` as failed). If the `CertificateRequest` was marked as failed, issuance will be re-tried in 1 hour.
+A non-fatal failure is one that doesn't mark the `CertificateRequest` as failed. If the `CertificateRequest` is marked as failed, issuance will be re-tried in 1 hour.
```
wallrj, line 59:
```suggestion
A temporary failure is one that doesn't mark the `CertificateRequest` as failed. If the `CertificateRequest` is marked as failed, issuance will be re-tried in 1 hour.
```

Fatal just sounds a bit too dramatic in my opinion.

content/en/docs/faq/terminology.md
```diff
@@ -0,0 +1,65 @@
...
+type: "docs"
+---
+
+With TLS being so widely deployed, terminology can sometimes get confused or be used to mean different things, and that reality
+combined with the complexity of TLS can lead to serious misunderstandings and confusion.
+
+### What does "publicly trusted" mean?
+
+Broadly speaking, a "publicly trusted" certificate is one that you can use on the public internet and expect
+that most reasonably up-to-date computers will be able to verify it using their system trust store.
```
wallrj, line 14:
```suggestion
Broadly speaking, a "publicly trusted" certificate is one that you can use on the Internet and expect
that most reasonably up-to-date computers will be able to verify it using their system trust store.
```

content/en/docs/faq/terminology.md
```diff
@@ -0,0 +1,65 @@
...
+
+There isn't a single standard trust store containing certs which are "publicly trusted", but generally most
+of the commonly seen trust stores are similar. An example would be [Mozilla's CA Certificate Program](https://wiki.mozilla.org/CA).
```
wallrj, line 17: I looked to see how Python certifi library get the trusted CAs:

https://github.com/certifi/python-certifi/blob/8110050d8441d363c8057d5ad187c8ad5a862fc2/Makefile#L1-L2

They use https://mkcert.org/ which might be worth linking to?

content/en/docs/faq/terminology.md
wallrj: I went digging around the man pages on my laptop and found

```
$ man update-ca-trust
...
       /etc/pki/tls/certs/ca-bundle.crt
           Classic filename, file contains a list of CA certificates trusted for TLS server authentication usage, in the simple BEGIN/END CERTIFICATE file format, without distrust information. This file is a symbolic link that
           refers to the consolidated output created by the update-ca-trust command.

       /etc/pki/tls/certs/ca-bundle.trust.crt
           Classic filename, file contains a list of CA certificates in the extended BEGIN/END TRUSTED CERTIFICATE file format, which includes trust (and/or distrust) flags specific to certificate usage. This file is a symbolic link
           that refers to the consolidated output created by the update-ca-trust command.

```

Which says something about a distinction between PEM labels "CERTIFICATE" and "TRUSTED CERTIFICATE" and that is described in this SO answer:
 * https://stackoverflow.com/a/5356351

content/en/docs/faq/terminology.md
```diff
@@ -0,0 +1,65 @@
...
+trusted roots. For this reason, roots are usually issued with very long lifetimes - often on the order of decades.
+
+#### Intermediate Certificates
+
+Intermediates are CA certificates signed by another CA. Most intermediates will be signed by a root certificate, but it's
+possible to construct longer chains where an intermediate can be signed by another intermediate.
+
+Intermediate certificates are usually issued with a much shorter lifetime than the CA which signed them. On the public
+internet, intermediate certificates are used on network-connected machines for day-to-day issuance so that the
+highly-valuable root certificates can remain entirely offline.
```
wallrj, line 54:
```suggestion
Intermediate certificates are usually issued with a much shorter lifetime than the CA which signed them. On the Internet, intermediate certificates are used on network-connected machines for day-to-day issuance so that the
highly-valuable root certificates can remain entirely offline.
```

content/en/docs/faq/terminology.md
```diff
@@ -0,0 +1,65 @@
...
+While intermediate certificates can also be explicitly trusted via addition to a trust store, they're usually validated
+by "walking up" the chain and validating signatures until an explicitly trusted self-signed root certificate is found.
+
+#### Leaf Certificates
+
+Leaf certificates are usually not CA certificates, and are used to represent a particular identity; on the public internet,
+leaf certificates usually identify a particular domain, such as `example.com`.
+
+Leaf certificates are generally at the end of a chain of certificates, and must be sent along with one or more intermediates
+to create a chain which can be validated by verifying signatures up to a trusted root certificate.
```
wallrj, line 65:
```suggestion
Leaf certificates are never CA certificates, and are used to represent a particular identity; on the public internet,
leaf certificates usually identify a particular domain, such as `example.com`.

Leaf certificates are always at the end of a chain of certificates, and must be sent along with one or more intermediates
to create a chain which can be validated by verifying signatures up to a trusted root certificate.
```
I referred to: https://en.wikipedia.org/wiki/Public_key_certificate#End-entity_or_leaf_certificate and I know Wikipedia is pretty unreliable but is there anywhere that describes a CA leaf certificate.

I mean does leaf / root have a different meaning when talking about a certificate bundle (e.g. tls.crt) as opposed to certificates used by clients and servers on the Internet. Is a leaf certificate, one without any children in a particular bundle. 
Is a root certificate  any self-signed certificate OR a certificate who's parent is not included in the same bundle?

wallrj (approved):
@SgtCoDFish Some random questions and suggestions below. But I agree with the others that this is a great addition to the docs.

content/en/docs/faq/terminology.md
wallrj: Which downloads trusted CAs from https://hg.mozilla.org/mozilla-central/raw-file/tip/security/nss/lib/ckfw/builtins/certdata.txt I think.

content/en/docs/faq/terminology.md
SgtCoDFish: > I assume I should understand "trust root" as "trusted root", is that correct? The term "trust root" seems a bit weird
> I guess we could also say "root of the chain of trust" too.

Yeah I think it's largely used interchangeably and it's usually pretty clear what's meant.

SgtCoDFish: > Which says something about a distinction between PEM labels "CERTIFICATE" and "TRUSTED CERTIFICATE"

This is good to bear in mind for https://github.com/cert-manager/trust/ perhaps :thinking:

content/en/docs/faq/terminology.md
SgtCoDFish: I think Let's Encrypt's chains are super interesting and might be relevant somewhere in this doc because it's an example of an explicitly trusted _intermediate_ certificate in public trust stores.

I think I'd prefer not to add the example here; I think this issue shows the need for trust stores to be kept up to date, but that's kinda tangential to what this terminology guide is trying to achieve. That's more of an example for a TLS best-practices guide IMO.

content/en/docs/faq/_index.md
```diff
@@ -20,19 +20,50 @@ face:
...
+> This is a sequence (chain) of certificates.  The sender's
+> certificate MUST come first in the list.  Each following
+> certificate MUST directly certify the one preceding it.  Because
+> certificate validation requires that root keys be distributed
+> independently, the self-signed certificate that specifies the root
+> certificate authority MAY be omitted from the chain, under the
+> assumption that the remote end must already possess it in order to
+> validate it in any case.
+
+In a standard, secure and correctly configured TLS environment, adding a root certificate to the chain is almost entirely _pure waste_.
```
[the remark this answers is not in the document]
SgtCoDFish: this reads better for sure, thanks. I've added a "almost always" before "unnecessary" because I can come up with very convoluted scenarios where it might not be entirely wasteful, such as:

Having the root in the chain theoretically allows a validator to more quickly check if the root appears in their trust store... but only really if they were intentionally trying to create pathologically evil trust stores with huge numbers of duplicated certificates :joy:

content/en/docs/faq/_index.md
SgtCoDFish: The [spec](https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.6) points at the same justification for server certificates. I'll add a line explaining this.

content/en/docs/faq/terminology.md
SgtCoDFish: My initial reaction here is that this might be out of scope for a terminology guide, because it's getting into pretty esoteric practices of customising trust stores. Is that fair?

These things would certainly be good to note in cert-manager/trust though.

content/en/docs/faq/terminology.md
SgtCoDFish: haha I'd not really thought about it but I guess "public internet" might be a bit redundant! thanks!

content/en/docs/faq/terminology.md
SgtCoDFish: > I referred to: https://en.wikipedia.org/wiki/Public_key_certificate#End-entity_or_leaf_certificate and I know Wikipedia is pretty unreliable but is there anywhere that describes a CA leaf certificate.

> Is a leaf certificate, one without any children in a particular bundle.

I can't think of why anyone would have a CA leaf certificate, I just know it's _technically_ possible. Probably too far out there to be worth mentioning; I'll remove that bit.

I think you could make the case that such a thing could exist, but I don't know why anyone would do that really and I probably shouldn't have accounted for it in this page.

irbekrm:
I've read through the doc again and I think this is a great addition to the website and will be really useful for folks- thanks for this!

It looks like all the comments have been addressed too.

/approve
/lgtm

PR #664, opened 2021-07-30
Update supported releases

maelvls:
EKS has changed their EOL (https://github.com/jetstack/cert-manager/pull/4253#issuecomment-889847190), so I updated them here. 

I also chose to merge the OpenShift section with the Kubernetes versions, since the purpose of this section was to explain why we don't support versions below Kubernetes 1.16. The table shows that the oldest Kubernetes version supported across vendors is 1.16.

I still kept the "mapping" table between OpenShift and Kubernetes releases since I never know which are which.

### DIFF
--- a/content/en/docs/installation/supported-releases.md
+++ b/content/en/docs/installation/supported-releases.md
@@ -17,14 +17,14 @@ release every two months.
 
 ## Supported releases {#supported-releases}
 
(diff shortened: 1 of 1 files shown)


maelvls:
@SgtCoDFish What do you think about this PR? 🔥

PR #665, opened 2021-07-31
Updates for Venafi as a Service

tr1ck3r:
Venafi Cloud is now Venafi as a Service.  Also clarified a few things around the configuration and updated external links.


tr1ck3r:
Will try resubmitting with DCO signoff

PR #669, opened 2021-08-04
kubectl --export is deprecated

zburgermeiszter:
kubectl --export is deprecated since `kubectl` v1.14 and has been removed in v1.18.

The new command is emulating the behavior of the deprecated `--export` flag using `yq`.

There is another `kubectl` command line with `--export` flag, but in our cluster the email address was provided as an environmental variable rather than a ConfigMap, therefore no tested command for it, but similar commands should work for both.

Signed-off-by: Zoltan Burgermeiszter <zoltan@backenddoctor.com>

### DIFF
--- a/content/en/docs/tutorials/acme/migrating-from-kube-lego.md
+++ b/content/en/docs/tutorials/acme/migrating-from-kube-lego.md
@@ -104,9 +104,17 @@ namespace that it is running in, and we used `cert-manager` when deploying
 cert-manager above. You should change this if you have deployed cert-manager
 into a different namespace.
 
+Install [yq](https://kislyuk.github.io/yq/#installation) then run the command below:
+
 ```bash
-$ kubectl create -f kube-lego-account.yaml \
-    --namespace cert-manager
+$ kubectl get secret kube-lego-account -o yaml \
+    --namespace kube-lego | yq -Y \
+    'del(
+       .metadata.creationTimestamp,
+       .metadata.resourceVersion,
+       .metadata.selfLink,
+       .metadata.uid
+     )' > kube-lego-account.yaml
 ```
 
 ## 4. Creating an ACME `ClusterIssuer` using your old ACME account
--- a/content/en/docs/tutorials/acme/migrating-from-kube-lego.md
+++ b/content/en/docs/tutorials/acme/migrating-from-kube-lego.md
@@ -88,15 +88,20 @@ the value of the `LEGO_SECRET_NAME` environment variable.
 You should download a copy of this secret resource and save it in your local
 directory:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/tutorials/acme/migrating-from-kube-lego.md
```diff
@@ -88,16 +88,21 @@ the value of the `LEGO_SECRET_NAME` environment variable.
...
     --namespace kube-lego \
-    --export > kube-lego-account.yaml
+     | yq -Y \
+    'del(
+       .metadata.creationTimestamp,
+       .metadata.resourceVersion,
+       .metadata.selfLink,
+       .metadata.uid,
+       .metadata.namespace
+     ) | .metadata.name = "letsencrypt-private-key"' > kube-lego-account.yaml
```
wallrj, line 103: I see. So this will set the name automatically. Good.

content/en/docs/tutorials/acme/migrating-from-kube-lego.md
```diff
@@ -88,16 +88,21 @@ the value of the `LEGO_SECRET_NAME` environment variable.
 You should download a copy of this secret resource and save it in your local
 directory:
 
+Install [yq](https://kislyuk.github.io/yq/#installation) then run the command below:
```
wallrj, line 91: I found two versions of `yq` when I googled for it. This seems to be the original version and according to my colleague @maelvls, it is the better of the two : https://github.com/jetstack/cert-manager/pull/3996#issue-643103680

The other tool that might be even better is `kubectl neat` which is mentioned in this SO article: https://stackoverflow.com/questions/61392206/kubectl-export-is-deprecated-any-alternative 
and in https://github.com/kubernetes/kubernetes/issues/90066

In that issue they also complain about the `managedFields` content which is apparently ommitted by `kubectl get` since v1.21 https://github.com/kubernetes/kubernetes/pull/96878

wallrj (changes requested):
Thanks @zburgermeiszter 

I haven't tested this, but I trust that you have. 

I left a few comments and suggestions which you can address or ignore as you please.

content/en/docs/tutorials/acme/migrating-from-kube-lego.md
zburgermeiszter: Unfortunately with my current setup I'm unable to verify the behavior of the tools with the `managedField.
The cluster is still on `1.18.20` and my `kubectl` version is `1.19.13`.

`kubectl neat` is a krew plugin and krew needs to be installed separately.
It feels an overkill for me compared to a simple yq installation.

What would be your preference?

PR #671, opened 2021-08-05
Fix docs for DNS solver for AzureDNS

psibi:
Having it as `azureDNS` throws this error message:

```
error validating data:
ValidationError(ClusterIssuer.spec.acme.solvers[0].dns01): unknown
field "azureDNS" in
io.cert-manager.v1alpha2.ClusterIssuer.spec.acme.solvers.dns01; if you
choose to ignore these errors, turn validation off with
--validate=false
```

Changing it to `azuredns` fixes it.

Signed-off-by: Sibi Prabakaran <sibi@psibi.in>

### DIFF
--- a/content/en/docs/configuration/acme/dns01/azuredns.md
+++ b/content/en/docs/configuration/acme/dns01/azuredns.md
@@ -79,7 +79,7 @@ kind: AzureIdentity
 metadata:
   annotations:
     # recommended to use namespaced identites https://azure.github.io/aad-pod-identity/docs/configure/match_pods_in_namespace/
-    aadpodidentity.k8s.io/Behavior: namespaced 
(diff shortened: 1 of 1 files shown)


wallrj (changes requested):
I think you must be using an old version of cert-manager and old version of the API.

The v1 API uses `azureDNS` https://cert-manager.io/docs/reference/api-docs/#acme.cert-manager.io/v1.ACMEChallengeSolverDNS01 

The v1beta1 API used `azuredns`: https://cert-manager.io/docs/reference/api-docs/#acme.cert-manager.io/v1beta1.ACMEChallengeSolverDNS01

And since cert-manager v1.0.0 will convert any legacy resource versions to v1 and then perform the validation.

So I recommend that you upgrade cert-manager.

PR #673, opened 2021-08-09
Fix broken certificaterequest yaml

SgtCoDFish:
"d" is not a valid character for a duration

/kind cleanup


content/en/docs/concepts/certificaterequest.md
```diff
@@ -27,7 +27,7 @@ spec:
   - signing
   - digital signature
   - server auth
-  duration: 90d
+  duration: 90h
```
jakexks, line 30:
```suggestion
  # 90 days
  duration: 2160h
```

PR #684, opened 2021-08-12
Don't force GOPROXY when generating docs

SgtCoDFish:
Forcing the value is harmful in environments which use a local caching GOPROXY (such as my own :grin:)

This is very similar to something we did in the main cert-manager repo: https://github.com/jetstack/cert-manager/commit/c86f20f4b54d47388dd5f93f2d8de53e9973e0c8

### DIFF
--- a/scripts/gendocs/generate
+++ b/scripts/gendocs/generate
@@ -40,7 +40,6 @@ trap cleanup EXIT
 # Create fake GOPATH
 echo "+++ Creating temporary GOPATH"
 export GOPATH="${tmpdir}/go"
-export GOPROXY="https://proxy.golang.org"
 export GO111MODULE="on"
 GOROOT="$(go env GOROOT)"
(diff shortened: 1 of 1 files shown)


scripts/gendocs/generate
```diff
@@ -40,7 +40,6 @@ trap cleanup EXIT
 # Create fake GOPATH
 echo "+++ Creating temporary GOPATH"
 export GOPATH="${tmpdir}/go"
-export GOPROXY="https://proxy.golang.org"
```
wallrj, line 43: It was probably added before this became the default
> For versions earlier than Go 1.13, you can configure the go command to download modules using this module mirror by setting GOPROXY=https://proxy.golang.org.
-- https://proxy.golang.org/#EnvironmentSetup

wallrj (approved):
/approve
/lgtm

PR #685, opened 2021-08-12
Remove alex check

SgtCoDFish:
We never looked at the output from this command - it didn't fail the build. It has so many false positives that it would take a
significant amount of effort to be able to use the output of this tool at all.

I like the idea of tools like this, but I don't think this specific tool - with the false-positives - is providing any value to us here.

We discussed removing it during the cert-manager 1.5 release on 2021-08-11 - I remember @wallrj saying it. This PR removes it, and also updates the package lock since we remove it from package.json too.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,29 @@
 {
   "name": "tech-doc-hugo",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
-  "dependencies": {
-    "@algolia/cli": {
+  "packages": {
+    "": {
+      "name": "tech-doc-hugo",
+      "version": "0.0.1",
+      "license": "ISC",
+      "devDependencies": {
+        "@algolia/cli": "^4.0.8",
+        "autoprefixer": "^9.4.6",
+        "hugo-algolia": "git+https://git@github.com/meyskens/hugo-algolia.git#bf56cc21e81d6344338e807d143e64b446613b7e",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1,23 +1,4012 @@
 {
   "name": "tech-doc-hugo",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
wallrj, line 4: Is this due to a different npm version or something?
Seems a bit weird to have a diff with twice as many additions as removals as a result of removing a dependency.

wallrj (approved):
Thanks Ash,

Hopefully this'll speed up the builds significantly.

Is there anyway to avoid the massive re-arrangement of the package-lock.json file.

irbekrm:
Glad we are removing this, agree with your points and I think a tool used in CI should simply fail or not instead of expecting devs to inspect output and take action.
 Perhaps this will improve the situation with the node module dependency weirdness too.

/lgtm
/approve

package-lock.json
irbekrm: (Sorry Richard, I didn't spot your comment before lgtm-ing)

SgtCoDFish: My process was to remove alex from `packages.json` and then running `./scripts/verify-lint` which calls `./scripts/update-node_modules`, which caused the huge change to `package-lock.json`. I'm not a node expert by any means, but this kind of thing seems par-for-the-course when it comes to node.

As for why the lockfile version changed, I don't really know! My local npm is v7.20.5 (nodejs v16.6.2) and it just... upgraded the number.

I figured since the tests passed it was probably fine to upgrade that lockfileVersion number.

PR #686, opened 2021-08-12
Update plugin download command to always download latest version

irbekrm:
Update kubectl cert-manager plugin install command so that it always downloads the latest version instead of v1.4.0


wallrj (approved):
/lgtm
/approve

PR #690, opened 2021-08-25
Bump helm installed cm version v1.5.0 -> v1.5.3

irbekrm:
Bump Helm installed cert-manager version v1.5.0 -> v1.5.3

I think we should always encourage users who want to install a particular minor version to install the latest patch version for that minor version. It feels suboptimal to bump it here every time after releasing a patch. I am not sure what would be the best solution for that - it is possible to specify a range of versions to helm install command with `--version ^X.Y`, but that doesn't seem to work for versions that are prefixed with `v`.
Perhaps we should take the version out of the install command alltogether.

### DIFF
--- a/content/en/docs/contributing/release-process.md
+++ b/content/en/docs/contributing/release-process.md
@@ -460,3 +460,9 @@ page if a step is missing or if it is outdated.
        updates to the website configuration. To do this, take inspiration from
        [Maartje's PR
        example](https://github.com/cert-manager/website/pull/309/files).
+
+    5. Ensure that any installation commands in
+       [`cert-manager/website`](https://github.com/cert-manager/website) install
+       the latest version. This should be done after every release, including
+       patch releases as we want to encourage users to always install the latest
+       patch.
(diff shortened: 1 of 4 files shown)


wallrj (changes requested):
I agree. 

We discovered yesterday that the use of "latest" in https://cert-manager.io/docs/installation/#default-static-install is also wrong.
It would redirect to the latest 1.4.x release if that happened to be released after the 1.5.x release. So that should also be changed to 1.5.3.

And there's another version reference in https://cert-manager.io/docs/installation/kubectl/

Maybe add a note to the release process doc explaining that these versions need to be updated as part of a patch release.

irbekrm:
Thanks @wallrj I've updated the PR.

> Maybe add a note to the release process doc explaining that these versions need to be updated as part of a patch release.

Added that, but I guess eventually we'll need a better solution, maybe some script that always returns latest release number like https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#install-kubectl-on-linux

wallrj (approved):
Strictly we should create a PR against the release-1.4 docs updating those too:
 * https://cert-manager.io/v1.4-docs/installation/

The release process is becoming such a bore!

/lgtm

irbekrm:
> Strictly we should create a PR against the release-1.4 docs updating those too

I've added that in #691 .

This is painful 😄

PR #694, opened 2021-09-01
Remove trailing `/` in docker registry

artificial-aidan:
Signed-off-by: Aidan Jensen <aidan@artificial.com>

### DIFF
--- a/content/en/docs/contributing/building.md
+++ b/content/en/docs/contributing/building.md
@@ -101,7 +101,7 @@ $ bazel build //cmd/ctl
 If you need the Docker images you can generate these using:
 ```bash
 $ export APP_VERSION="dev"
-$ export DOCKER_REGISTRY="quay.io/jetstack/"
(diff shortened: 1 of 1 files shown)


irbekrm:
Thanks @artificial-aidan !

/ok-to-test
/approve
/lgtm

As a side note, I wonder if it would make sense to eventually replace these Bazel commands with the make ones from [cert-manager/Makefile](https://github.com/jetstack/cert-manager/blob/master/Makefile), to make this more contributor friendly.

artificial-aidan:
> Thanks @artificial-aidan !
> 
> /ok-to-test
> /approve
> /lgtm
> 
> As a side note, I wonder if it would make sense to eventually replace these Bazel commands with the make ones from [cert-manager/Makefile](https://github.com/jetstack/cert-manager/blob/master/Makefile), to make this more contributor friendly.

I actually had problems with the makefile commands. Bazel worked better odly enough.

PR #695, opened 2021-09-02
Add snippet for plausible

james-w:


maelvls:
Is this PR about swapping Google Analytics with Plausible.io pixel-based analytics?

It looks like we are still using Google Analytics: https://github.com/cert-manager/website/blob/45706c794507db87f07a628183fc0f5b6edf0900/layouts/partials/head.html#L20. Should we remove the Google Analytics JS too?

Last note about the CNCF Sandbox: the website analytics need to be owned by the CNCF as per [project-onboarding.md](https://github.com/cncf/toc/blob/0f1c830a73c79eb3cd42ee7724fb814e575bd898/.github/ISSUE_TEMPLATE/project-onboarding.md). With Plausible.io, would that be the case?

layouts/partials/scripts.html
```diff
@@ -12,4 +12,13 @@
...
+{{ partial "hooks/body-end.html" . }}
+<script type="text/javascript">
+  (function (d, u, h, s) {
+    h = d.getElementsByTagName('head')[0];
+    s = d.createElement('script');
+    s.async = 1;
+    s.src = u + new Date().getTime();
+    h.appendChild(s);
+  })(document, 'https://grow.clearbitjs.com/api/pixel.js?v=');
+</script>
```
maelvls, line 24: Can we have this added only in production builds like what is done for Google Analytics in [layouts/partials/head.html](https://github.com/cert-manager/website/blob/45706c794507db87f07a628183fc0f5b6edf0900/layouts/partials/head.html#L19-L21)?

E.g.:

```go
{{ if eq (getenv "HUGO_ENV") "production" }}
{{ end }}
```

RiichardCC: We need to keep GA in the header as well. pls do not remove this as the CNCF will wan it. The plausible snippet is just an initial view to see the analytics this service can produce until handover to the CNCF is complete.

james-w: I have updated to put it in that file with the production check, thanks.

maelvls:
Great!! Let's merge that.

/lgtm

maelvls (approved):
Oops, apparently I also need to do a GitHub "Approve"

PR #698, opened 2021-09-07
Improve docs for fargate networking to avoid confusion

### LINKED ISSUE #696: Fargate caveats page made me think that I should set `webhook.HostNetwork=true` but you actually must not
(cert-manager 1.5.3 / k8s 1.21 in EKS on Fargate)

On [the cloud compatibility page](https://cert-manager.io/docs/installation/compatibility/#aws-fargate) there's the sentence

>  It’s worth noting that using AWS Fargate to run cert-manager will force you to run using the host’s network

This made me think that I had to set `hostNetwork: true` on the deployment.

But as far as I can tell (not super clear on this), that happens implicitly on Fargate, and setting it manually doesn't work, since it causes a `hostPort` to be generated which you can't have, resulting in the webhook pod never starting with this:

```
  Warning  FailedScheduling  112s  fargate-scheduler  Pod not supported on Fargate: fields not supported: HostNetwork, port contains HostIP or HostPort
```

I wonder if that could be called out more explicitly somehow.

SgtCoDFish:
Fixes #696

/kind documentation

### DIFF
--- a/content/en/docs/installation/compatibility.md
+++ b/content/en/docs/installation/compatibility.md
@@ -74,9 +74,13 @@ port; see the warning at the top of the page for details.
 
 ### AWS Fargate
 
-It's worth noting that using AWS Fargate to run cert-manager will force you to
-run using the host's network, and will force a port clash with the kubelet
-running on port 10250, as seen in [#3237](https://github.com/jetstack/cert-manager/issues/3237).
+It's worth noting that using AWS Fargate doesn't allow much network configuration and
+will cause the webhook's port to clash with the kubelet running on port 10250, as seen
+in [#3237](https://github.com/jetstack/cert-manager/issues/3237).
 
 When deploying cert-manager on Fargate, you _must_ change the port on which
 the webhook listens. See the warning at the top of this page for more details.
+
+Because Fargate forces you to use its networking, you cannot manually set the networking
+type and options such as `webhook.hostNetwork` on the helm chart will cause your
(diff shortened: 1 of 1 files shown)


jakexks (commented):
Looks good. Would it be worth adding this report: https://kubernetes.slack.com/archives/C4NV3DWUC/p1631004960113100?thread_ts=1630950059.102500&cid=C4NV3DWUC to the docs too? (stating that runAsUser: 1001 was required on fargate)

SgtCoDFish:
> Looks good. Would it be worth adding this report: https://kubernetes.slack.com/archives/C4NV3DWUC/p1631004960113100?thread_ts=1630950059.102500&cid=C4NV3DWUC to the docs too? (stating that runAsUser: 1001 was required on fargate)

I don't think I'm fully convinced that the reported issue relating to STS/IAM and `runAsUser` are related, there.  I don't see why runAsUser would have anything to do with that error :thinking: 

I think pending further evidence that runAsUser is required I'd rather leave that advice out, or I'd worry it could become a cargo cult thing.

SgtCoDFish:
@jakexks friendly prod:

> Looks good. Would it be worth adding this report: https://kubernetes.slack.com/archives/C4NV3DWUC/p1631004960113100?thread_ts=1630950059.102500&cid=C4NV3DWUC to the docs too? (stating that runAsUser: 1001 was required on fargate)

do you think this is is required for this PR or can we get it merged?

PR #700, opened 2021-09-10
DNS01 webhooks : add cert-manager-alidns-webhook

olivierboudet:


irbekrm:
Thanks for the contibution!

The project looks good to me, I see docs and tests 😄 

A bit curious why it was chosen to write a new webhook as I see that there is already one external webhook for alidns https://github.com/pragkent/alidns-webhook?

/ok-to-test

olivierboudet:
Thanks !

> A bit curious why it was chosen to write a new webhook as I see that there is already one external webhook for alidns https://github.com/pragkent/alidns-webhook?

Just because I did the first commit before the pragkent's one :)

irbekrm:
That makes sense 👍🏼 

Glad to see this going in- thanks for the contribution!

/lgtm
/approve

PR #701, opened 2021-09-13
Issuer with IRSA needs ambient credentials flag

rossigee:
This should help reduce the amount of time people might waste trying to figure out how to resolve the following error:

```
error instantiating route53 challenge solver: unable to construct route53 provider: empty credentials; perhaps you meant to enable ambient credentials?
```

A couple of related bug reports:

* https://github.com/jetstack/cert-manager/issues/3009
* https://github.com/jetstack/cert-manager/issues/3079

### DIFF
--- a/content/docs/configuration/acme/dns01/route53.md
+++ b/content/docs/configuration/acme/dns01/route53.md
@@ -178,7 +178,20 @@ spec:
 
 Note that, as mentioned above, the pod is using `arn:aws:iam::XXXXXXXXXXX:role/cert-manager` as a credentials source in Account X, but the `ClusterIssuer` ultimately assumes the `arn:aws:iam::YYYYYYYYYYYY:role/dns-manager` role to actually make changes in Route53 zones located in Account Y.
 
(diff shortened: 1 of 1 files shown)


tstraley:
I'm curious if any of the maintainers can chime in here with _why_ the Issuer and ClusterIssuer have different functionality here? Shouldn't either one be trying to use the modern AWS SDK to its full extent and collecting credentials from all supported sources?

wallrj (changes requested):
Thanks @rossigee 

And apologies for the delayed review.
Judging by the number of issues mentioning the ambient-credentials flags, 
this is a much needed addition to the documentation.

I was surprised at the lack of docs and when I went digging into the origin of this feature, 
I found that that original author did write documentation for this:
 * https://github.com/jetstack/cert-manager/blob/95883c47dd31ad6c0e7652a192e4c51981d67592/docs/user-guides/ambient-credentials.md
 * https://github.com/jetstack/cert-manager/pull/363

That document contains some important extra information about why ambient-credentials are disabled by default for Issuer. 
Please review that document and copy over all the content that you think is still relevant.

Thanks.

PR #704, opened 2021-09-17
Fix supported releases example timeline

mkozal:
Signed-off-by: Michał Kozal <michal@kozal.eu>


maelvls:
Well spotted, thanks! 🙏 

/good-to-test

PR #705, opened 2021-09-20
Fix the plusible snippet

james-w:
When I added the Plausible snippet in e8a62ef1d786c856566defe70927f8704c9d9611
I copy and pasted the wrong tracking code, and used the one for
Clearbit instead of Plusible. This corrects that error and uses
the tracking code for Plausible.


james-w:
This time I used /etc/hosts to be able to test from my local machine and verified that the
hits are appearing in plausible.io.

maelvls (approved):
/approve

PR #715, opened 2021-10-15
Adds documentation for the new cmctl command.

JoshVanL:

This PR adds a page for the new `cmctl` CLI. This page mostly replaces the existing `kubectl-plugin` page taking the same content but changing the commands and some wording.

The exiting `kubectl-plugin` page has been stripped down to just installing and linking to `cmctl`.

Users can still find all the docs they need, but moving the docs this way helps nudge users into using `cmctl` which has a better UX.

/assign @jakexks


JoshVanL:
> This is documentation for 1.6, should we wait until the final release or merge it to release-next instead?

Don't mind, whatever you think will be easiest.

PR #718, opened 2021-10-19
Change links in usage certificate to kubectl plugin #renew to be relative

JoshVanL:
/assign @jakexks 

This PR changes references in release-1.4/usage/certificate to usage/kubectl-plugin/#renew to use a relative path. When using an absolute path, it looks like it will reference the latest docs, which will no longer exist after this PR #715. Changing to the relative path should now link to the page within the correct version.


JoshVanL:
@jakexks once this has merged, we can run a /retest on #715 which should tell us that this is how it is working.

PR #740, opened 2021-10-30
Update digitalocean.md

benjick:
This doesn't work if you base64 encode the access token, it will instead give a 401 error. Unencoded works fine

### DIFF
--- a/content/en/docs/configuration/acme/dns01/digitalocean.md
+++ b/content/en/docs/configuration/acme/dns01/digitalocean.md
@@ -16,7 +16,7 @@ metadata:
   name: digitalocean-dns
 data:
   # insert your DO access token here
-  access-token: "base64 encoded access-token here"
+  access-token: "access-token here"
   ```
 
 The access token must have write access.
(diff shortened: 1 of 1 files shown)


content/en/docs/configuration/acme/dns01/digitalocean.md
```diff
@@ -16,7 +16,7 @@ metadata:
   name: digitalocean-dns
 data:
   # insert your DO access token here
-  access-token: "base64 encoded access-token here"
+  access-token: "access-token here"
   ```
```
wallrj, line 20: If you prefer to use plain text in the Secret manifest, you can use the `stringData` stanza rather than the `data` stanza,
and then kubectl / kube-apiserver (not sure which) will automatically base64 encode the values for you.

So the documentation as it stands is correct.

https://kubernetes.io/docs/concepts/configuration/secret (search for `stringData`)

wallrj (changes requested):
I think the documentation is correct, but you could add a note about using `stringData` instead if you think that would improve the docs.

SgtCoDFish:
Closing since the docs are currently correct. As Richard said, adding a note about stringData might be nice if you want to re-open or open a new PR.

PR #744, opened 2021-11-03
Don't show removed API reference docs

irbekrm:
This PR depends on #745 which needs to be merged and cherry-picked into release-1.6 branch first as else this will fail because we still refer to alpha/beta API versions from the API reference doc in a few places .

This PR adds a workaround to ensure that API reference docs generated from cert-manager release-1.6 branch do not include the documentation for the removed alpha and beta APIs.
The 'next' docs that are generated from release-1.7 branch still refers to alpha and beta APIs, but this will not need a workaround as we are going to remove those packages in cert-manager 1.7

Note: as the deploy preview does not include API reference docs, this fix can be verified by running the website locally:
1. checkout this branch
2. run `./scripts/gendocs/generate`
3. run `./scripts/server`
4. go to `http://localhost:1313/` and verify that the 'current' API reference docs do not include the alpha and beta API docs

### DIFF
--- a/scripts/gendocs/generate
+++ b/scripts/gendocs/generate
@@ -45,7 +45,7 @@ GOROOT="$(go env GOROOT)"
 export GOROOT
 GOBIN="${tmpdir}/bin"
 export GOBIN
-go get github.com/ahmetb/gen-crd-api-reference-docs@v0.2.0
+go get github.com/ahmetb/gen-crd-api-reference-docs@v0.3.0
 
 mkdir -p "${GOPATH}/src/github.com/jetstack"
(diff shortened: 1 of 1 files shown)


wallrj (approved):
I think this makes sense,
but don't think we need to  cherry-pick in to the release-1.6 branch.
The current release is `1.6` and the docs for this are in `master` branch (until we release 1.7)

And if I use the correct preview link (there are now two!), I can see that the legacy APIs are no longer shown :+1: 
 * https://deploy-preview-744--cert-manager.netlify.app/docs/reference/api-docs/

/lgtm
/approve

PR #747, opened 2021-11-03
adding redirect rules to prevent google links from reaching page not …

chance19:
…found

Signed-off-by: Chance Whittaker <chance19@gmail.com>

Currently some links from Google, such as https://docs.cert-manager.io/en/release-0.11/ would route to https://cert-manager.io/docs/en/release-0.11/ which would take the browser to a 404 page not found (see screenshot)
image

This change should ensure that these links are at least routed back to the specific release-note page that was referenced in the link. This is not a full solution to the problem just an improvement on the current situation.

PR #748, opened 2021-11-04
Add temp 302 redirect to all other paths than the root

mattbates:
Based on feedback from @munnerz, propose we do the following (for now):

* Use a permanent (301) redirect for `docs.cert-manager.io` 
* Use a temporary (302) redirect for all other paths (which may or may not work currently)

We should follow-up to this PR with redirect rules for specific pages that we know are linked from other pages and feature prominently in search engine listings.

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -38,4 +38,5 @@ https://cert-manager.io/docs/usage/kubectl-plugin/#certificatesigningrequest htt
 
 # docs.cert-manager.io was previously a separately hosted service. The dns has since been redirect to the main site and the following redirect
 # rule will force people with historical links or browser to docs.cert-manager.io to be redirected to the /docs endpoint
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm

From reading [netlify docs](https://docs.netlify.com/routing/redirects/redirect-options/) this looks correct to me!

munnerz (approved):
/approve

munnerz:
Seems like Netlify dropped webhooks or something.. 🤔 you may want to `git commit --amend && git push -f origin head` (where 'origin' is the name of your `mattbates` remote/fork) to give it a kick

chance19:
@mattbates to your point re:

> We should follow-up to this PR with redirect rules for specific pages that we know are linked from other pages and feature prominently in search engine listings.

I started work on this with https://github.com/cert-manager/website/pull/747

PR #750, opened 2021-11-04
Add specific redirect rule for docs subdomain

SgtCoDFish:
http://docs.cert-manager.io/en/latest/ seems to be another link that has been shared

I'm not 100% confident that this is exactly the right format for netlify

/kind cleanup


j-fuentes:
/lgtm

based on the other routes in the redirect file.

jakexks (approved):
/lgtm

PR #751, opened 2021-11-05
Added kubectl config for recursive nameservers

rhjensen79:
The documentation, is lacking how to set recursive nameservers, when deploying using kubectl.

I found the answer in the following issue : 
https://github.com/jetstack/cert-manager/issues/1163#issuecomment-912671595

I have testet it, and it works perfect.

### DIFF
--- a/content/en/docs/configuration/acme/dns01/_index.md
+++ b/content/en/docs/configuration/acme/dns01/_index.md
@@ -80,6 +80,15 @@ with `--set`:
 --set 'extraArgs={--dns01-recursive-nameservers-only,--dns01-recursive-nameservers=8.8.8.8:53\,1.1.1.1:53}'
 ```
 
+If you are deploying using kubectl, then you can add the following line, to the `spec.template.spec.containers.args`
+```bash
+- args:
+        - --v=2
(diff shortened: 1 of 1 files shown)


content/en/docs/configuration/acme/dns01/_index.md
```diff
@@ -80,6 +80,15 @@ with `--set`:
...
 ```
 
+If you are deploying using kubectl, then you can add the following line, to the `spec.template.spec.containers.args`
+```bash
+- args:
+        - --v=2
+        - --cluster-resource-namespace=$(POD_NAMESPACE)
+        - --leader-election-namespace=kube-system
+        - --dns01-recursive-nameservers="8.8.8.8:53"
+```
```
wallrj, line 90: We're in the process of adding a config file mechanism, which will allow these settings to be configured using a ConfigMap, but for now, this will be helpful.

Perhaps you should make it clear that the user will need to edit the downloaded cert-manager manifests file, before applying it.

wallrj (approved):
Thanks @rhjensen79 

/ok-to-test
/approve

PR #754, opened 2021-11-10
Broaden redirect for "latest/tasks" path on old docs site

SgtCoDFish:
This is intended to fix a broken link under https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#set-up-a-signer which points to https://docs.cert-manager.io/en/latest/tasks/issuers/setup-ca.html

Thanks to Jon for reporting!

/kind cleanup

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -40,8 +40,21 @@ https://cert-manager.io/docs/usage/kubectl-plugin/#certificatesigningrequest htt
 # These rules are in place to capture traffic that is specifically referencing these source urls
 https://docs.cert-manager.io/en/latest/getting-started/index.html https://cert-manager.io/docs/tutorials/
 https://docs.cert-manager.io/en/latest/tutorials/acme/index.html https://cert-manager.io/docs/tutorials/
-https://docs.cert-manager.io/en/latest/tasks/issuers/index.html https://cert-manager.io/docs/configuration/
 https://docs.cert-manager.io/en/latest/ https://cert-manager.io/docs/
+
+# Issuer-specific redirects; these were mined from a wayback machine crawl of the old site:
+# https://web.archive.org/web/20190802192846/http://docs.cert-manager.io/en/latest/tasks/issuers/index.html
+https://docs.cert-manager.io/en/latest/tasks/issuers/setup-acme/http01/* https://cert-manager.io/docs/configuration/acme/http01/
+https://docs.cert-manager.io/en/latest/tasks/issuers/setup-acme/dns01/* https://cert-manager.io/docs/configuration/acme/dns01/
+https://docs.cert-manager.io/en/latest/tasks/issuers/setup-acme/* https://cert-manager.io/docs/configuration/acme/
+https://docs.cert-manager.io/en/latest/tasks/issuers/setup-ca.html https://cert-manager.io/docs/configuration/ca/
+https://docs.cert-manager.io/en/latest/tasks/issuers/setup-selfsigned.html https://cert-manager.io/docs/configuration/selfsigned/
+https://docs.cert-manager.io/en/latest/tasks/issuers/setup-vault.html https://cert-manager.io/docs/configuration/vault/
+https://docs.cert-manager.io/en/latest/tasks/issuers/setup-venafi.html https://cert-manager.io/docs/configuration/venafi/
+
+# fallback in case there are any pages we missed:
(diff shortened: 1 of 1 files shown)


james-w:
I think this will mean that that URL redirects to `docs/configuration/setup-ca.html` but that is a 404. Should it be `docs/configuration/ca/` instead?

_redirects
```diff
@@ -40,7 +40,7 @@ https://cert-manager.io/docs/usage/kubectl-plugin/#certificatesigningrequest htt
 # These rules are in place to capture traffic that is specifically referencing these source urls
 https://docs.cert-manager.io/en/latest/getting-started/index.html https://cert-manager.io/docs/tutorials/
 https://docs.cert-manager.io/en/latest/tutorials/acme/index.html https://cert-manager.io/docs/tutorials/
-https://docs.cert-manager.io/en/latest/tasks/issuers/index.html https://cert-manager.io/docs/configuration/
+https://docs.cert-manager.io/en/latest/tasks/issuers/* https://cert-manager.io/docs/configuration/
```
mattbates, line 43: you'll need to add the `:splat` if you wish for the path to be added to the redirected URL - ie

```
https://docs.cert-manager.io/en/latest/tasks/issuers/* https://cert-manager.io/docs/configuration/:splat
```

SgtCoDFish:
> I think this will mean that that URL redirects to `docs/configuration/setup-ca.html` but that is a 404. Should it be `docs/configuration/ca/` instead?

I wouldn't have expected that :thinking: I would've thought for the redirect to be to `docs/configuration/setup-ca.html` to happen the rule would've had to be:

```text
https://docs.cert-manager.io/en/latest/tasks/issuers/* https://cert-manager.io/docs/configuration/:splat
```

E.g. [in this line we already have](https://github.com/cert-manager/website/pull/754/files#diff-fd40fde62534b781806d746f82ba621d39c640f85029c1524582d4b137701cfeR47) the redirect works as I expect. But I don't know this area well and I don't really know of a good way to test this before merging.

My instinct would be to merge as-is and then I can quickly raise a new PR to fix if it's not correct

james-w:
> > I think this will mean that that URL redirects to `docs/configuration/setup-ca.html` but that is a 404. Should it be `docs/configuration/ca/` instead?
> 
> I wouldn't have expected that 🤔 I would've thought for the redirect to be to `docs/configuration/setup-ca.html` to happen the rule would've had to be:
> 
> ```
> https://docs.cert-manager.io/en/latest/tasks/issuers/* https://cert-manager.io/docs/configuration/:splat
> ```
> 
> E.g. [in this line we already have](https://github.com/cert-manager/website/pull/754/files#diff-fd40fde62534b781806d746f82ba621d39c640f85029c1524582d4b137701cfeR47) the redirect works as I expect. But I don't know this area well and I don't really know of a good way to test this before merging.
> 
> My instinct would be to merge as-is and then I can quickly raise a new PR to fix if it's not correct

Ah, I misunderstood the syntax, sorry.

This redirect therefore takes the link to configure a CA issuer and just points to the config section, do we want
to point to the new CA issuer docs?

_redirects
SgtCoDFish: That wasn't my intention, no. There's no obvious mapping of the names of individual html pages on the `docs.cert-manager.io` site to pages under `https://cert-manager.io/docs/` - I think using `:splat` would be the incorrect behaviour here?

jakexks (approved):
/lgtm
/hold

I think we should merge this, (but it's always a bit scary as you can't test it)

We should also update the actual Kubernetes website.

SgtCoDFish:
> This redirect therefore takes the link to configure a CA issuer and just points to the config section, do we want
to point to the new CA issuer docs?

That's a fair point - we can do better with the redirects. From looking at [an old wayback machine crawl](https://web.archive.org/web/20190725192351/https://docs.cert-manager.io/en/latest/tasks/issuers/setup-ca.html) I can see these pages are guides for setting up issuers - I'll add redirects for each individual issuer.

jakexks (commented):
/lgtm

SgtCoDFish:
/unhold

(we'll see if this gets us to a happy place; I can easily raise another if needed)

SgtCoDFish:
... it redirects to https://cert-manager.io/docs/en/latest/tasks/issuers/setup-ca.html now, flying in the face of all logic I'd expected :joy:

PR #755, opened 2021-11-10
Redirect all 404s to the root of the site

SgtCoDFish:
I think this is required because the theme we use uses relative links for CSS files, which break if the 404 page isn't served from the root.

Currently a [404 on the root](https://cert-manager.io/dasdub) looks vaguely correct while a [404 under /docs](https://cert-manager.io/docs/asdasdasd) looks totally broken. This should hopefully fix that.

This line was taken from the [hugo docs](https://gohugo.io/templates/404/#automatic-loading). The [netlify docs](https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling) seem to contradict and suggest that this line is not needed, but I think it's only not needed if the 404 page uses absolute paths.

/kind cleanup


irbekrm:
Looks good, I think there's no other way to prove for sure than to try it 😄 

/lgtm
/approve

PR #761, opened 2021-11-18
Add kubectl plugin Windows installation instructions

abousias:
There where no instructions for Windows installation of the kubectl plugin, so these are the steps I did. 
My plan is to make a chocolatey package as well like it is in [kubectl's docs](https://kubernetes.io/docs/tasks/tools/install-kubectl-windows/#install-on-windows-using-chocolatey-or-scoop) (not sure about scoop yet), but that will follow in a future PR.

Signed-off-by: Angelos Bousias <angelos@bousias.net>

### DIFF
--- a/content/en/docs/usage/kubectl-plugin.md
+++ b/content/en/docs/usage/kubectl-plugin.md
@@ -15,12 +15,20 @@ While the kubectl plugin is supported, it is recommended to use
 ## Installation
 You need the `kubectl-cert-manager.tar.gz` file for the platform you're using, these can be found on our [GitHub releases page](https://github.com/jetstack/cert-manager/releases).
 In order to use the kubectl plugin you need its binary to be accessible under the name `kubectl-cert_manager` in your `$PATH`.
+
+### macOS/Linux
 Run the following commands to set up the plugin:
 ```console
 $ curl -L -o kubectl-cert-manager.tar.gz https://github.com/jetstack/cert-manager/releases/latest/download/kubectl-cert_manager-linux-amd64.tar.gz
 $ tar xzf kubectl-cert-manager.tar.gz
 $ sudo mv kubectl-cert_manager /usr/local/bin
 ```
+
+### Windows
+1. Download the [latest version](https://github.com/jetstack/cert-manager/releases/latest/download/kubectl-cert_manager-windows-amd64.tar.gz).
+2. Extract the archive.
+3. Add the `.exe` file extension to the extracted `kubectl-cert_manager`.
+4. Copy `kubectl-cert_manager.exe` to a location which is also in your `PATH`.
 
 You can run `kubectl cert-manager help` to test the plugin is set up properly:
 ```console


content/en/docs/usage/kubectl-plugin.md
```diff
@@ -15,13 +15,21 @@ While the kubectl plugin is supported, it is recommended to use
...
+### macOS/Linux
 Run the following commands to set up the plugin:
 ```console
 $ curl -L -o kubectl-cert-manager.tar.gz https://github.com/jetstack/cert-manager/releases/latest/download/kubectl-cert_manager-linux-amd64.tar.gz
 $ tar xzf kubectl-cert-manager.tar.gz
 $ sudo mv kubectl-cert_manager /usr/local/bin
 ```
 
+### Windows
+1. Download the [latest version](https://github.com/jetstack/cert-manager/releases/latest/download/kubectl-cert_manager-windows-amd64.tar.gz).
```
wallrj, line 28: I need to double check, but I think `/latest` is unsuitable, because it redirects to what ever happens to be the newest release on the releases page, which might be alpha, beta, or patch release of a previous minor release.

content/en/docs/usage/kubectl-plugin.md
```diff
@@ -15,13 +15,21 @@ While the kubectl plugin is supported, it is recommended to use
...
+2. Extract the archive.
+3. Add the `.exe` file extension to the extracted `kubectl-cert_manager`.
+4. Copy `kubectl-cert_manager.exe` to a location which is also in your `PATH`.
```
wallrj, line 31: Thanks for figuring all this out.
In https://github.com/jetstack/cert-manager/pull/4554 @SgtCoDFish is going to add the `.exe` extension by default, 
so we'll have to remember to update these docs when that happens.

wallrj (approved):
Thanks @abousias 

If the tests all pass, I'll be happy to merge this.

/approve
/ok-to-test

content/en/docs/usage/kubectl-plugin.md
SgtCoDFish: We'll also add a zip version, which should be much more natural on Windows compared with tar.gz!

content/en/docs/usage/kubectl-plugin.md
abousias: @wallrj `/latest` does not include the `pre-releases`, it actually points to the the one "tagged" `latest`.
image
To make sure though I double checked it  with an old repo 😃 
The latest pre-release has a `test.txt` asset:
https://github.com/abousias/corepkgappveyor/releases/download/1.0.2-pre/test.txt
But it is not available in `/latest` because it gets redirected to `1.0.1`:
https://github.com/abousias/corepkgappveyor/releases/latest/download/test.txt

content/en/docs/usage/kubectl-plugin.md
abousias: With a bit of free time I hope to continue with the chocolatey package, so a zip with both executables would be perfect for that too! Thanks!

wallrj (approved):
Thanks @abousias  for checking the behaviour of the `/latest` redirect.
In that case let's merge this and we could consider using the `/latest` URL elsewhere in the docs.

One point to consider is that we serve the docs for our older cert-manager versions, 
so it would be good if the instructions for those versions pointed to the appropriate binaries,
but we can fix that when we do the releases and archive the docs.

/lgtm
/approve

PR #767, opened 2021-12-02
venafi: service-generated CSRs must not be locked

maelvls:
It seems our documentation isn't precise enough and does not mention the possibility for "Service Generated CSRs" to be a valid configuration as long as it is not locked; for example the following works with cert-manager:

Screenshot 2021-12-02 at 16 34 13

Signed-off-by: Maël Valais <mael@vls.dev>

cc @SpectralHiss

### DIFF
--- a/content/en/docs/configuration/venafi.md
+++ b/content/en/docs/configuration/venafi.md
@@ -118,9 +118,22 @@ of the connection parameters are slightly different.
 > this is the only type supported by cert-manager at this time.
 >
 > More specifically, the valid configurations of the "CSR handling" are:
+>
 > - "User Provided CSRs" selected and unlocked,
 > - "User Provided CSRs" selected and locked,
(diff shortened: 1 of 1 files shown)


SpectralHiss:
if we just add that cert-manager can work with the policy set to server provided csr without it being locked it would be good.
In that  case vcert (and by extension, cert-manager) will set organisation fields etc. will be set to the policy value.

I don't think the CSR itself is modified but this is a leaky abstraction by TPP, it's just that in that mode the certificate will have different properties derived from the policy than what the CSR requested.

maelvls:
@SpectralHiss Got it. I added an explanation of what will happen if you use "Service Generated CSR" unlocked. What do you think?

wallrj (approved):
Thanks both of you for figuring this out.

/lgtm
/approve

PR #773, opened 2021-12-06
Describing two more usecases through documentation

lboix:


content/en/docs/configuration/acme/http01/_index.md
```diff
@@ -139,6 +139,7 @@ No other fields of the `podTemplate` exist.
 ### `ingressTemplate`
 
 It is possible to add labels and annotations to the solver ingress resources.
+It can be really useful when you are managing several Ingress Controllers across your cluster and you want to make sure that the right one will pick up and expose the solver (for the upcoming challenge to resolve).
```
irbekrm, line 142: > It can be really useful when you are managing several Ingress Controllers across your cluster

Do you mean using the deprecated [ingress class annotation](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation)? Since cert-manager v1.5 the Ingress resources are always created with `ingressClassName` field set (which the users can configure via `acme.solvers.http01.ingress.class` field on issuer spec) - as I undertand ingress classes should be the standard way how to manage multiple ingress controllers in a cluster, see [kube docs](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class)

irbekrm (changes requested):
Thank you very much for the PR, @lboix!

I've added some comments- I think that the first change is very useful as it was not clear by looking at the docs what the behaviour would be if issuer was configured via multiple sources. I've only added some wording suggestions.
I'm not so sure about the second change as I think ingress classes should now be used to manage multiple ingress controllers.

Let me know what you think.

/ok-to-test

content/en/docs/configuration/acme/http01/_index.md
lboix: Yes you are right we can add the precision like "If you are using a old version of Ingress Controller (who rely on old annotation `kubernetes.io/ingress.class`), it can be really useful..."

What do you think ? Is it too much information ?

irbekrm: > If you are using a old version of Ingress Controller (who rely on old annotation kubernetes.io/ingress.class

Those old versions of ingress controllers will anyway not work (at least the ingress-shim without `edit-in-place` annotation will not work) with cert-manager 1.5+ as since cert-manager 1.5 the temporary `Ingress` resource that _cert-manager_ creates has the `ingressClassName` field _only_. There is some discussion around what to do as this is breaking for users of some ingress implementations (see https://github.com/jetstack/cert-manager/issues/4537), but as it is now I would be a little reluctant to add that line because this PR will get merged to latest docs. What do you think?

irbekrm:
The `pull-cert-manager-website-verify` test failure looks like it might be a flake

/test all

content/en/docs/configuration/acme/http01/_index.md
lboix: Yes I do agree

lboix:
@irbekrm is there a way for you to override the dco step if I add something in a new commit for example ?

In this mindset for example : https://github.com/todogroup/gh-issues/issues/50#issuecomment-1043070927

irbekrm:
Hi @lboix 

Thanks for your contribution- sorry for leaving this PR open for so long!

Yes, we can override DCO, happy to do that for this case to get this merged 👍🏼 

/lgtm
/approve

lboix:
That's great thank you very much for your help @irbekrm and I wish you a great day :)

PR #774, opened 2021-12-07
Update release date (with note) for 1.7

SgtCoDFish:
We're clearly not releasing today ( :sweat_smile: ) and I don't think we have a good handle on when exactly we'll release yet. This leaves it up in the air to reflect reality.

/kind cleanup


maelvls:
Thank you for updating the doc, I did not thought about it.

/lgtm

PR #775, opened 2021-12-09
Fix incorrect indentation in documentation

tete17:
The selector field belongs to the `ACMEChallengeSolver` according to the api
documentation, and we should not add another entry in the list for the `dns01` key

In the `acme` section the selector key was not properly indented.


irbekrm (approved):
Thank you for spotting these errors and most importantly for submitting fixes! 🙏🏼 

/ok-to-test
/lgtm

PR #777, opened 2021-12-10
typo

BryanStenson-okta:
official docs from AWS PCA here: https://github.com/cert-manager/aws-privateca-issuer#awspcaissuer


irbekrm (approved):
Thanks for the fix! 🙏🏼 

/lgtm
/ok-to-test

PR #781, opened 2021-12-19
Documentation for using reflector for cross-namespace secret sync-ing

winromulus:
Added documentation for using [reflector](https://github.com/emberstack/kubernetes-reflector) for cross-namespace sync-ing of secrets.
There were numerous requests for documenting this with cert-manager from the extension's support channels.


JoshVanL:
Thanks @winromulus, would it be possible to add a redirect [here](https://github.com/cert-manager/website/blob/master/_redirects), so that links to the old page (which _only_ contained kubed) will now be directed to this new renamed page?

winromulus:
@JoshVanL good point, I did not spot that. Adding now.

JoshVanL:
Thanks @winromulus, this looks like it is doing the right thing.

/lgtm
/approve
/ok-to-test

PR #782, opened 2021-12-20
release-process: mention the removal/re-enabling of next periodics

maelvls:
As per #779, @SgtCoDFish made the point that the links in "next periodic" line should be removed until we have released an `alpha.0`. In this PR, I add more steps in the release process so that we don't forget about those steps.


SgtCoDFish (approved):
/lgtm
/approve

Obvious improvement, great stuff. Thanks :+1:

PR #786, opened 2021-12-21
Install 1.7.0-alpha.0

wallrj:
Signed-off-by: Richard Wall <richard.wall@jetstack.io>


irbekrm:
Agree that since the docs for 1.7 are already there, the install instructions should actually install 1.7. I do wonder whether we want to show 1.7 docs yet, but it's not what this PR is about

/lgtm
/approve

PR #793, opened 2022-01-04
Update instructions for upgrading cert-manager resources and CRDs

irbekrm:
Update the instructions how to ensure that cert-manager resources are stored in etcd at v1 and that the deprecated API versions are not recorded on CRD status fields to refer to the new `cmctl upgrade migrate-api-version` command

Testing the upgrade 
1. Create a kind cluster with Kubernetes <= 1.21
2. `helm install cert-manager jetstack/cert-manager -ncert-manager --create-namespace --set installCRDs=true --version v0.16`
3. Create some legacy resources such these:
```yaml
kind: Issuer
metadata:
  name: selfsigned-issuer-legacy
spec:
  selfSigned: {}
---
kind: Certificate
metadata:
  name: test-old
spec:
  secretName: test-old
  dnsNames:
  - foo.bar
  issuerRef:
    name: selfsigned-issuer-legacy
    kind: Issuer
    group: cert-manager.io
```

These will be stored in etcd at `v1beta1` which is the preferred storage version for cert-manager v0.16

7. Upgrade to cert-manager v1.6 (or any other cert-manager version that is v1 or higher) 
`helm upgrade cert-manager jetstack/cert-manager -ncert-manager --create-namespace --set installCRDs=true --version v1.6.1`
8.Run `cmctl upgrade migrate-api-version`
10. Upgrade to cert-manager `v1.7.0-alpha.0`
11. Verify that the upgrade works and you can still retrieve all cert-manager resources

/kind cleanup

### DIFF
--- a/content/en/docs/installation/upgrading/remove-deprecated-apis.md
+++ b/content/en/docs/installation/upgrading/remove-deprecated-apis.md
@@ -1,11 +1,11 @@
 ---
-title: "Removing Deprecated API Resources"
-linkTitle: "Removing Deprecated API Resources"
+title: "Migrating Deprecated API Resources"
+linkTitle: "Migrating Deprecated API Resources"
 weight: 20
 type: "docs"
 ---
 
-We have deprecated the following cert-manager APIs: 
+The following cert-manager APIs were deprecated in cert-manager `v1.4`:
 
 - `cert-manager.io/v1alpha2`
 - `cert-manager.io/v1alpha3`
@@ -14,37 +14,18 @@ We have deprecated the following cert-manager APIs:
 - `acme.cert-manager.io/v1alpha3`
 - `acme.cert-manager.io/v1beta1`
 
-These APIs were removed in cert-manager `v1.6.0`. If you have a cert-manager installation that is using or has previously used these deprecated APIs you'll need to upgrade your cert-manager custom resources and CRDs. This needs to be done before upgrading to cert-manager `v1.6.0`.
+These APIs are no longer served in cert-manager `v1.6` and are fully removed in cert-manager `v1.7`. If you have a cert-manager installation that is using or has previously used these deprecated APIs you might need to upgrade your cert-manager custom resources and CRDs. This should be done before upgrading to cert-manager `v1.6` or later.
 
-## Upgrading existing cert-manager resources
+{{% alert title="Note" color="warning" %}}
+An earlier version of this document listed a number of kubectl commands to run to migrate resources. These steps have now been encoded in [`cmctl upgrade migrate-api-version` command](https://cert-manager.io/docs/usage/cmctl/#migrate-api-version). If you have already run the kubectl commands, your resources should have been migrated and there should be no need to also run the `cmctl` command. However, if you are not sure, you can still run the `cmctl` command as well- it will be a no-op if no actions are needed.
+{{% /alert %}}
+
+# Upgrading existing cert-manager resources
 
 1. Familiarize yourself with the [official Kubernetes documentation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects) on CRD versioning.
-2. Make sure your installed cert-manager deployment is `v1.0.0` or later.
 
-3. Make sure any version-controlled cert-manager custom resource config files that still use the deprecated APIs are updated to use the `cert-manager.io/v1` API and re-applied. This should update stored versions of the given resources.
+2. Make sure your cert-manager deployment is currently at version `v1.0` or later.
 
-After that, follow the official Kubernetes documentation [here](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version):
+3. Make sure that any cert-manager custom resource manifests that refer to the deprecated APIs are updated to use the `cert-manager.io/v1` API and re-applied. You can use the [cmctl convert command](https://cert-manager.io/docs/usage/cmctl/#convert)to convert manifests.
 
-1. If there are other resources (e.g. `Certificate`s created by ingress-shim, `CertificateRequest`s, etc.) that are not version controlled and may have been created using the deprecated APIs, run
-   ```bash
-   kubectl get <resource_name> -oyaml | kubectl apply -f -
-   ```
-   to force the resources to be stored in `etcd` at the current storage version.
-
-2. To remove legacy API versions from cert-manager CRDs run the following `curl` commands:
-
-```bash
-kubectl proxy &
-
-curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v1"] }]' -H "Content-Type: application/json-patch+json"  -X PATCH http://localhost:8001/apis/apiextensions.k8s.io/v1/customresourcedefinitions/certificates.cert-manager.io/status
-
-curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v1"] }]' -H "Content-Type: application/json-patch+json"  -X PATCH http://localhost:8001/apis/apiextensions.k8s.io/v1/customresourcedefinitions/certificaterequests.cert-manager.io/status
-
-curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v1"] }]' -H "Content-Type: application/json-patch+json"  -X PATCH http://localhost:8001/apis/apiextensions.k8s.io/v1/customresourcedefinitions/issuers.cert-manager.io/status
-
-curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v1"] }]' -H "Content-Type: application/json-patch+json"  -X PATCH http://localhost:8001/apis/apiextensions.k8s.io/v1/customresourcedefinitions/clusterissuers.cert-manager.io/status
-
-curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v1"] }]' -H "Content-Type: application/json-patch+json"  -X PATCH http://localhost:8001/apis/apiextensions.k8s.io/v1/customresourcedefinitions/orders.acme.cert-manager.io/status
-
-curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v1"] }]' -H "Content-Type: application/json-patch+json"  -X PATCH http://localhost:8001/apis/apiextensions.k8s.io/v1/customresourcedefinitions/challenges.acme.cert-manager.io/status
-```
+4. Run [cmctl upgrade migrate-api-version command](https://cert-manager.io/docs/usage/cmctl/#migrate-api-version) to migrate any resources that might be stored in `etcd` at the deprecated API versions and to patch the cert-manager CRDs.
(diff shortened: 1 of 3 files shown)


content/en/docs/installation/upgrading/remove-deprecated-apis.md
```diff
@@ -14,27 +14,110 @@ We have deprecated the following cert-manager APIs:
...
+
+4. Make sure that any non-version controlled cert-manager resources (e.g. `Certificate`s created by ingress-shim, `CertificateRequest`s, etc.) are updated to `v1`. The Kubernetes documentation describes two alternative ways how to do that:
+
+A. Use [storage version migrator](https://github.com/kubernetes-sigs/kube-storage-version-migrator)
+We have tested Storage Version Migrator [`v0.0.5`](https://github.com/kubernetes-sigs/kube-storage-version-migrator/releases/tag/v0.0.5):
+- Follow the [storage version migrator installation instructions](https://github.com/kubernetes-sigs/kube-storage-version-migrator/blob/master/USER_GUIDE.md). (To use the published `v0.0.5` storage migrator images, modify the command that creates local manifests to `make local-manifests REGISTRY=eu.gcr.io/k8s-artifacts-prod/storage-migrator VERSION=v0.0.5`). Alternatively, if using managed Kubernetes, you may already have Storage Migrator installed- look for `storageversionmigrations.migration.k8s.io` and `storagestates.migration.k8s.io` `CustomResourceDefinition`s in cluster to verify if it installed.
```
irbekrm, line 40: The bit about using v0.0.5 images is needed as the installation instructions for storage migrator doesn't include that part https://github.com/kubernetes-sigs/kube-storage-version-migrator/pull/98

irbekrm:
I've tried to add a wording that suggests that these instructions should be applied with caution and also some scripts to verify that things have been upgraded, but it does feel like this is now a bit verbose- happy to remove some of the verbosity or scripts if suggested :)

content/en/docs/installation/upgrading/remove-deprecated-apis.md
```diff
@@ -14,27 +14,110 @@ We have deprecated the following cert-manager APIs:
...
+These APIs are no longer served in cert-manager `v1.6` and are fully removed in cert-manager `v1.7`. If you have a cert-manager installation that is using or has previously used these deprecated APIs you'll need to upgrade your cert-manager custom resources and CRDs. This should be done before upgrading to cert-manager `v1.6`.
 
-## Upgrading existing cert-manager resources
+{{% alert title="Warning" color="warning" %}}
+Although the deprecated APIs are no longer served in cert-manager `v1.6`, it is possible that at this point some cert-manager resources are still stored in `etcd` at the deprecated API versions and that the deprecated versions are still recorded on the cert-manager `CustomResourceDefinition`s' statuses.
+You can verify the storage versions on CRD status fields by running `kubectl get crd <crd-name>  -ojsonpath='{.status.storedVersions}`, but we are not aware of any straightforward way how to verify what versions of resources are stored in `etcd`.
```
munnerz, line 21: This sentence is a little confusing IMO - what does it mean/what's the reason for verifying the storage version on CRD status fields if this *doesn't* reflect what's in etcd? What's actionable here from an end-user standpoint? :)

irbekrm: The upgrade will fail if there is a removed storage version recorded on CRD fields no matter whether the resources have been stored in etcd at v1 or not

munnerz: 👍 yep that makes sense - but I don't think this message is making that clear. Can we adjust it to be a bit more actionable for a user?

"Before the upgrade can proceed, you must ensure that the `status.storedVersions` field only contains `v1`. If the upgrade is attempted whilst there are still other versions recorded in this field, the upgrade will fail and your installation may be left inoperable until it is resolved." or something like that?

irbekrm: I _think_ I've come across some upstream issues where the point of the stored versions is explained, but cannot find those now..

irbekrm: > yep that makes sense - but I don't think this message is making that clear. Can we adjust it to be a bit more actionable for a user?

That warning was there because we previously believed (and told users) that upgrade to v1.6 would show if there are resources at the deprecated versions in the cluster (as v1.6 stops no longer serves those APIs)- it is meant to convey that it's possible to be at v1.6 and not notice that you have old resources/not updated CRD status fields. It's not meant to be a general warning that CRD statuses need to be updated as that is already what this doc is about.
Agree that it's not best worded, I can remove it 👍🏼

irbekrm:
/hold

some related conversation here https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1641311819078800

We might end up putting all those steps into a tool

content/en/docs/installation/upgrading/remove-deprecated-apis.md
```diff
@@ -14,27 +14,110 @@ We have deprecated the following cert-manager APIs:
...
+
+4. Make sure that any non-version controlled cert-manager resources (e.g. `Certificate`s created by ingress-shim, `CertificateRequest`s, etc.) are updated to `v1`. The Kubernetes documentation describes two alternative ways how to do that:
+
+A. Use [storage version migrator](https://github.com/kubernetes-sigs/kube-storage-version-migrator)
+We have tested Storage Version Migrator [`v0.0.5`](https://github.com/kubernetes-sigs/kube-storage-version-migrator/releases/tag/v0.0.5):
+- Follow the [storage version migrator installation instructions](https://github.com/kubernetes-sigs/kube-storage-version-migrator/blob/master/USER_GUIDE.md). (To use the published `v0.0.5` storage migrator images, modify the command that creates local manifests to `make local-manifests REGISTRY=eu.gcr.io/k8s-artifacts-prod/storage-migrator VERSION=v0.0.5`). Alternatively, if using managed Kubernetes, you may already have Storage Migrator installed- look for `storageversionmigrations.migration.k8s.io` and `storagestates.migration.k8s.io` `CustomResourceDefinition`s in cluster to verify if it installed.
```
maelvls, line 40:
```suggestion
- Follow the [storage version migrator installation instructions](https://github.com/kubernetes-sigs/kube-storage-version-migrator/blob/master/USER_GUIDE.md). (To use the published `v0.0.5` storage migrator images, modify the command that creates local manifests to `make local-manifests REGISTRY=eu.gcr.io/k8s-artifacts-prod/storage-migrator VERSION=v0.0.5`). Alternatively, if you are using a managed Kubernetes cluster, you may already have the Storage Migrator installed. Look for `storageversionmigrations.migration.k8s.io` and `storagestates.migration.k8s.io` `CustomResourceDefinition`s in your cluster to verify if it is installed.
```

wallrj:
The cmctl tool has now merged, so we should reword this PR to explain how and when to use the tool
* https://github.com/jetstack/cert-manager/pull/4711
* xref: https://github.com/jetstack/cert-manager/issues/4686

content/en/docs/installation/upgrading/remove-deprecated-apis.md
irbekrm: This has now been changed to use the cmctl command

content/en/docs/installation/upgrading/remove-deprecated-apis.md
irbekrm: I've changed the contents to refer to the cmctl install command

maelvls (commented):
I have tested the instructions performed:

- one upgrade from v1.6.1 to v1.7.0-alpha.0,
- one upgrade from v1.5.3 to v1.6.1.

Both went smoothly.

That said, could we give some indication to people who use a GitOps-like flow? The step (4) seems to indicate the command has to be run locally, which isn't possible for people using a GitOps flow (cc @jsoref)

irbekrm:
> That said, could we give some indication to people who use a GitOps-like flow?

I think they should just run it from a script then, unless they use helm or something else that has a concept of pre-install hook.

content/en/docs/installation/upgrading/remove-deprecated-apis.md
```diff
@@ -14,37 +14,18 @@ We have deprecated the following cert-manager APIs:
...
+1. Familiarize yourself with the [official Kubernetes documentation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects) on CRD versioning.
 
-curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v1"] }]' -H "Content-Type: application/json-patch+json"  -X PATCH http://localhost:8001/apis/apiextensions.k8s.io/v1/customresourcedefinitions/clusterissuers.cert-manager.io/status
+2. Make sure your cert-manager deployment is currently at version `v1.0` or later.
 
-curl -d '[{ "op": "replace", "path":"/status/storedVersions", "value": ["v1"] }]' -H "Content-Type: application/json-patch+json"  -X PATCH http://localhost:8001/apis/apiextensions.k8s.io/v1/customresourcedefinitions/orders.acme.cert-manager.io/status
+3. Make sure that any cert-manager custom resource manifests that refer to the deprecated APIs are updated to use the `cert-manager.io/v1` API and re-applied. You can use the [cmctl convert command](https://cert-manager.io/docs/usage/cmctl/#convert)to convert manifests.
```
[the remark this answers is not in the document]
irbekrm: It's a remnant from the previous version, but I suppose you do always want to ensure that you've applied whatever you have in Git as opposed to doing one thing to Git, another thing in cluster

content/en/docs/installation/upgrading/migrating-deprecated-apis.md
```diff
@@ -0,0 +1,31 @@
...
+{{% /alert %}}
+
+# Upgrading existing cert-manager resources
+
+1. Familiarize yourself with the [official Kubernetes documentation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects) on CRD versioning.
+
+2. Make sure your cert-manager deployment is currently at version `v1.0` or later.
+
+3. Make sure that any cert-manager custom resource manifests that refer to the deprecated APIs are updated to use the `cert-manager.io/v1` API and re-applied. You can use the [cmctl convert command](https://cert-manager.io/docs/usage/cmctl/#convert)to convert manifests.
```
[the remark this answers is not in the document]
irbekrm: cmctl convert converts manifests, it does not apply anything to cluster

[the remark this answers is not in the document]
irbekrm: > such as manifests stored in Git for people using a GitOps flow

I'm not keen on adding this part. I would like to think that version controlled manifests is a default, this makes it sound like an exception

content/en/docs/installation/upgrading/migrating-deprecated-apis.md
```diff
@@ -0,0 +1,31 @@
...
+
+4. Run [cmctl upgrade migrate-api-version command](https://cert-manager.io/docs/usage/cmctl/#migrate-api-version) to migrate any resources that might be stored in `etcd` at the deprecated API versions and to patch the cert-manager CRDs.
```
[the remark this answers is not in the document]
maelvls: Ah right yes, that's a good point. I don't really know how people manage this, since sometimes the only possibility is to commit to a git repository (as opposed to manually running `kubectl`).

I'll ping people on #argo-cd (CNCF slack) just to make sure I understand what people think about this upgrade path.

maelvls:
Thanks for writing this PR 🙏🙏

/lgtm

maelvls:
@irbekrm Do you think we can remove `WIP` from the issue title?

PR #795, opened 2022-01-06
Updates supported releases page

irbekrm:
Updates supported releases (updates the info about what tests we're running as we're now testing against 1.23)

/kind cleanup

### DIFF
--- a/content/en/docs/installation/supported-releases.md
+++ b/content/en/docs/installation/supported-releases.md
@@ -183,21 +183,15 @@ The list of supported Kubernetes versions displayed in the [Supported
 Releases](#supported-releases) section depends on what the cert-manager
 maintainers think is reasonable to support and to test.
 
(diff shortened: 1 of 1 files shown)


irbekrm:
As a side note (unrelated to this PR's contents) I feel like this page has a lot in it that is likely to become out-of-date/is a pain to update manually such as dates- not sure what to do about that though

SgtCoDFish:
/lgtm

(I'm assuming the test will pass on a retry)

> As a side note (unrelated to this PR's contents) I feel like this page has a lot in it that is likely to become out-of-date/is a pain to update manually such as dates- not sure what to do about that though

Agreed, it's a pain - ideally large chunks of this page could be automatically generated from other things, although that's maybe a task for another time

irbekrm:
Thanks for the quick review 🚀 😄

PR #798, opened 2022-01-19
fix arrow character in testing coverage table

dsch:
Other Kubernetes version ranges use "Rightwards Arrow" (unicode 2192) character to express a range of versions.


wallrj (approved):
Thanks for spotting that.

/lgtm
/approve
/ok-to-test

PR #800, opened 2022-01-20
generate cli reference docs

jahrlin:
Fixes https://github.com/jetstack/cert-manager/issues/4458

adds a new function `genclireference` that will attempt to run `main.go --help` for our CLIs and writes the output to markdown files for each version of the docs.

contains some crude checks to cover some corner cases where some CLIs don't exist for older releases, or when they don't produce any output from `--help` (pre-cobra)

the preview deployment does not seem to include generated docs.

result examples:
image
image


jahrlin:
Updated after discussion with team today.

Previous version CLI reference docs are added to git, current and next version will always be generated - as they are likely to change while you are working with the docs

scripts/gendocs/generate
```diff
@@ -83,6 +100,54 @@ gendocs() {
...
+  fi
+
+  outputdir="$1"
+  target="$2"
+  name="$3"
+  echo "+++ Generating CLI reference docs for $target ..."
+
+  mkdir -p "${REPO_ROOT}/content/en/${outputdir}/cli/"
+
+  output=$(go run "$target/main.go" --help)
```
JoshVanL, line 131: question: was continuing the work of generating the documentation via cobra considered? https://github.com/jetstack/cert-manager/blob/master/tools/cobra/main.go

Not a huge thing, but it does  give a _sightlier_ better output format, and can be configured with headers etc.

jahrlin: Going forward it could definitely make sense to build on top of that, the new website will not be based on Hugo so it could be we need something purpose-built rather than this simple script.

At this stage we decided to simply use the output of `--help`, and we wanted the CLI reference for older versions.

jahrlin: Tested it a bit now and it's not a huge change to use the tool you linked for the versions where it's available, but it requires some work on both ends and I'm not sure it's worth it?

jakexks: Thought: I'm surprised that --help doesn't output to standard error.

jakexks (approved):
This is definitely an improvement over not having anything at all, so I think it should merge.

I agree with Josh that the output format could use some work in future, but the website is about to go through a big style overhaul so I'm not too worried about it right now.

The .gitignore overrides were discussed with others in the team, I have no real opinion on this as long as it doesn't conflict with the other doc generation steps.

/lgtm
/hold

Placing a hold for anyone who chatted with @jahrlin earlier to also review, if they like.

PR #801, opened 2022-01-21
Secret needs to be in cert-manager namespace

p0l0:
To be able to use the secret, it must be in the _cert-manager_ namespace, if not you will get an error saying that the secret can not be found.

### DIFF
--- a/content/en/docs/configuration/acme/dns01/cloudflare.md
+++ b/content/en/docs/configuration/acme/dns01/cloudflare.md
@@ -26,6 +26,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: cloudflare-api-token-secret
+  namespace: cert-manager
 type: Opaque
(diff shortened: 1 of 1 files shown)


wallrj (changes requested):
Thanks @p0l0 

But it's not quite right.
When creating an **Issuer**, the Secret MUST be in the same namespace as the Issuer.
When creating a **ClusterIssuer**, the Secret MUST be in the `--cluster-resource-namespace`, which is USUALLY `cert-manager` (but not always).
See:
 * https://cert-manager.io/docs/faq/cluster-resource/

I think it would be helpful to add a short note to this doc explaining this  caveat and linking to fhe FAQ for further details.

SgtCoDFish:
Going to close this since it's been open with a request for changes for months. Please feel free to re-open with the suggested changes!

PR #805, opened 2022-01-24
Add release step for updating algolia index

SgtCoDFish:
We've missed this in every release since 0.15 so now might be a good time to start :grin:

### DIFF
--- a/content/en/docs/contributing/release-process.md
+++ b/content/en/docs/contributing/release-process.md
@@ -511,3 +511,8 @@ page if a step is missing or if it is outdated.
        the latest version. This should be done after every release, including
        patch releases as we want to encourage users to always install the latest
        patch.
+
+    9. Open a PR against our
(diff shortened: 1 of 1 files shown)


jakexks (approved):
/lgtm

content/en/docs/contributing/release-process.md
```diff
@@ -511,3 +511,8 @@ page if a step is missing or if it is outdated.
        the latest version. This should be done after every release, including
        patch releases as we want to encourage users to always install the latest
        patch.
+
+    9. Open a PR against our
+       [Algolia indexing configuration](https://github.com/algolia/docsearch-configs/blob/master/configs/cert-manager.json#L7-L13)
+       including the new version for search indexing, as in
```
munnerz, line 517: It's actually the *old* version that we need to include here (e.g. when v1.7 is release, v1.6 should be added to that list)

PR #809, opened 2022-01-25
feat: next website

PatrickHeneise:
new cert-manager website based on next.js

Needs an `.env` file or environment variables set up:

```
DOCS_FALLBACK="README"
DOCS_PATH="content"
DOCS_FOLDER="docs"
NEXT_PUBLIC_ALGOLIA_KEY=""
NEXT_PUBLIC_ALGOLIA_ID=""
NEXT_PUBLIC_ALGOLIA_INDEX="docs"
DOCS_USE_MDX=false
```


jahrlin:
**Change request**

keyboard navigation is incomplete on the refresh

http://localhost:3000/docs/

- while the `select` for version is focused, pressing enter should open the dropdown (space works, but not enter)
- the menu items on the left side are targetable by tabbing, but lack has no visible focus indicator
- search box has no focus indicator

jahrlin:
Thanks for the accessibility fixes. 

I'd like to have the browser-native `:focus` `outline` property. 

I think we can simply **remove** this CSS rule from global.scss:

```scss
a:focus, button:focus, input:focus, select:focus {
  outline: 2px solid transparent;
  outline-offset: 2px;
} 
```

After that I am happy with the visual stuff.

C-Mejlak:
> I'd like to have the browser-native `:focus` `outline` property.

@jahrlin while totally possible, I'm not 100% sure that's what we want.
For example, this is how the search box would end up looking when focused (whether by keyboard or mouse).

image

It is quite common to customise the `:focus` styles. In fact even the current cert-manager site has custom `:focus` styles.
However, if you prefer to have the browser control that, we're happy to make the change

JoshVanL:
Having some trouble running the website locally. I ran the following commands at the root of the repo at this PR, but my browser is stuck "loading".

```console 
npm i
npm dev run
```

->

```terminal
> cert-manager.io@2.0.0 dev
> next dev

ready - started server on 0.0.0.0:3000, url: http://localhost:3000
info  - Loaded env from /home/josh/go/src/github.com/cert-manager/website/.env
```

image

```console
$ npm version
{
  'cert-manager.io': '2.0.0',
  npm: '8.4.1',
  node: '17.3.0',
  v8: '9.6.180.15-node.12',
  uv: '1.42.0',
  zlib: '1.2.11',
  brotli: '1.0.9',
  ares: '1.18.1',
  modules: '102',
  nghttp2: '1.46.0',
  napi: '8',
  llhttp: '6.0.4',
  openssl: '1.1.1m',
  cldr: '40.0',
  icu: '70.1',
  tz: '2021a3',
  unicode: '14.0'
}
```

---

As an aside, it doesn't look like npm is respecting `SIGINT`s (`ctrl-C`), and I have to close it using `SIGKILL`.

SgtCoDFish:
I think @JoshVanL's comment above might need addressing before this can be merged - I've had the same issue before, as has Joakim, and all 3 of us on different machines.

SgtCoDFish:
There are 3 public key paths which are served on the cert-manager website today. The following two work on a local checkout:

- https://cert-manager.io/public-keys/cert-manager-keyring-2021-09-20-1020CF3C033D4F35BAE1C19E1226061C665DF13E.gpg
- https://cert-manager.io/public-keys/cert-manager-pgp-2021-09-20-1020CF3C033D4F35BAE1C19E1226061C665DF13E.asc

This one _doesn't_ work on a local checkout and appears to have been deleted.  This file can't move - it needs to exist at this path:

- https://cert-manager.io/public-keys/cert-manager-pubkey-2021-09-20.pem

SgtCoDFish:
I found a great doc explaining the requirements on the cert-manager project's website when mentioning Jetstack as the "origin company": https://github.com/cncf/foundation/blob/main/website-guidelines.md

Several people have mentioned to me about the "Enterprise" heading on the refreshed site, and I think changing that to "Support" might be good, as well as changing the path to "/support" (it's currently "/enterprise").

In addition, the "Find Out More" link on that page is currently just a link to the same page: https://cert-manager.zntrd.cloud/enterprise/

C-Mejlak:
> In addition, the "Find Out More" link on that page is currently just a link to the same page: https://cert-manager.zntrd.cloud/enterprise/

What URL should this be pointing to pls?

SgtCoDFish:
I think https://www.jetstack.io/jetstack-secure/ would be a good start; we can always change it down the road.

PatrickHeneise:
> I think @JoshVanL's comment above might need addressing before this can be merged - I've had the same issue before, as has Joakim, and all 3 of us on different machines.

There was a similar issue before. A clean checkout resolved the issue.

C-Mejlak:
"Enterprise" renamed to "Support"
"Find out more" link updated

jahrlin:
> > I'd like to have the browser-native `:focus` `outline` property.
> 
> @jahrlin while totally possible, I'm not 100% sure that's what we want. For example, this is how the search box would end up looking when focused (whether by keyboard or mouse).
> 
> image
> 
> It is quite common to customise the `:focus` styles. In fact even the current cert-manager site has custom `:focus` styles. However, if you prefer to have the browser control that, we're happy to make the change

Most important to me is that there is a high contrast indicator for focus, which we have - so I'm happy with the current state!

Holding for others

/lgtm
/hold

SgtCoDFish:
/remove-lgtm

I think the substance here is pretty much good to go, but I _don't_ think this is ready to merge into `master` as is:

1. The deploys will fail without changing some netlify config at the least
2. There are changes on master which haven't been incorporated into the new site yet, leading to content "disappearing" if this goes live in its current state. It won't be hard to update this PR to include the changes, but that will need doing.
3. This PR (intentionally) doesn't include a solution for older versions of docs, and so merging to master means we'd lose old content until that solution is in place (which could be us doing the changes [I described in a doc](https://docs.google.com/document/d/14BWiKcBsvsbahtp7rcrZ0XBujkslQPxiZv4pq7nkGXs/edit#))

I think we need a deployment plan for this to be sure we don't cause an outage of the cert-manager website, and at the very least we might need a temporary branch in this repo (e.g. `next-website`) if the old docs are going to be merged in a separate PR.

If we do what I suggested [in the doc](https://docs.google.com/document/d/14BWiKcBsvsbahtp7rcrZ0XBujkslQPxiZv4pq7nkGXs/edit#) - which I'll get started on now - that'll solve point number 3 for us quite nicely.

SgtCoDFish:
I changed the base to "next-website" so we can take a more careful approach to merging, and so that we can add new docs to that branch after they've been run through the script I've been working on :+1:

SgtCoDFish:
I've manually cherry picked the latest commit from `zentered:feat/next-website` to the next-website branch on the cert-manager/website repo, without any content. As such I think this PR is redundant and I'll close it; we can work against the next-website branch :+1:

PR #811, opened 2022-01-26
fix CLI reference generation for current and next versions

jahrlin:
This PR fixes an issue where CLI reference docs for the current release was generated from master, meaning they would contain flags included in the upcoming release.

Signed-off-by: Joakim Ahrlin <joakim.ahrlin@gmail.com>

### DIFF
--- a/scripts/gendocs/generate
+++ b/scripts/gendocs/generate
@@ -63,18 +63,20 @@ gencliversions="${1:-0}"
 genversion() {
 	checkout "$1"
 	gendocs "$2"
+}
 
-  if [ "$gencliversions" == 1 ]; then
-    genclireference "$2" "cmd/acmesolver" "acmesolver"
-    genclireference "$2" "cmd/cainjector" "cainjector"
-    genclireference "$2" "cmd/ctl" "cmctl"
-    genclireference "$2" "cmd/controller" "controller"
-    genclireference "$2" "cmd/webhook" "webhook"
+genversionwithcli() {
+  genversion "$1" "$2"
 
-    # if any of the above steps succeeded copy over the index file
-    if [ -d "$REPO_ROOT/content/en/$2/cli" ]; then
-      cp "$REPO_ROOT/content/en/docs/cli/_index.md" "${REPO_ROOT}/content/en/$2/cli/"
-    fi
+  genclireference "$2" "cmd/acmesolver" "acmesolver"
+  genclireference "$2" "cmd/cainjector" "cainjector"
+  genclireference "$2" "cmd/ctl" "cmctl"
+  genclireference "$2" "cmd/controller" "controller"
+  genclireference "$2" "cmd/webhook" "webhook"
+
+  # if any of the above steps succeeded copy over the index file
+  if [ -d "$REPO_ROOT/content/en/$2/cli" ]; then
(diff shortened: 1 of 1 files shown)


scripts/gendocs/generate
```diff
@@ -142,21 +144,22 @@ $output
...
 # The branches named here exist in the `jetstack/cert-manager` repo.
 
-genversion "release-1.7" "next-docs"
+genversionwithcli "release-1.7" "next-docs"
 
 # In cert-manager 1.6 cert-manager.io and acme.cert-manager.io alpha and beta
 # API versions are no longer served, but the apis are still in the codebase. We
 # don't want to show the docs for those API versions so this is a workaround.
 # This will only be necessary for the release-1.6 branch.
 checkout "release-1.6"
```
wallrj, line 155: This `checkout` and the comment above relate to the `rm` and `gendocs` command below,
so I think it the `genclireference` commands should be moved after those.

scripts/gendocs/generate
```diff
@@ -142,21 +144,22 @@ $output
...
+
+genclireference "docs" "cmd/acmesolver" "acmesolver"
+genclireference "docs" "cmd/cainjector" "cainjector"
+genclireference "docs" "cmd/ctl" "cmctl"
+genclireference "docs" "cmd/controller" "controller"
+genclireference "docs" "cmd/webhook" "webhook"
```
wallrj, line 161: And if these lines are deliberately placed after the `checkout` command, then I'd add a comment explaining why it's important, so that someone doesn't come alone and rearrange this code.

scripts/gendocs/generate
```diff
@@ -174,9 +177,9 @@ rm -r pkg/apis/certmanager/v1alpha3
 rm -r pkg/apis/certmanager/v1beta1
 gendocs "v1.6-docs"
 
-# when adding a new version here run
-# ./scripts/gendocs/generate 1 to force generation of CLI docs
-# and add them to the git index
+# when adding a new version here use
+# genversionwithcli to force generation of CLI docs
+# and add them to the git index, then switch back to genversion
```
wallrj, line 182: I don't understand this comment. 
Does it mean that I should add another `genversionwithcli` line somewhere in the scipt.
Please add a little more detail about how this script should be modified in future.

wallrj (changes requested):
Thanks @jahrlin 

I ran it on my laptop without errors.
And I see that the preview now shows the current webhook CLI flags rather than the flags in master: https://deploy-preview-811--cert-manager.netlify.app/docs/cli/webhook/

There are just one or two things I didn't understand about the script.

scripts/gendocs/generate
jahrlin: I have re-formulated the comment and included an example for the next change in this section

scripts/gendocs/generate
jahrlin: Good point, I have moved the comment.

`genclireference` must happen before we delete the legacy API versions because the code does not compile without them. I added a comment on that as well

scripts/gendocs/generate
jahrlin: I have added a comment here covering this!

jahrlin:
Thanks @wallrj - very good feedback and I have tried to simplify and clarify the comments

wallrj (approved):
Thanks @jahrlin 

/lgtm

PR #812, opened 2022-01-26
Add notes on Ingress Class compatibility for 1.5-1.7

jakexks:
Hopefully this addresses concerns found during testing of the Ingress Annotation fallout

### DIFF
--- a/content/en/docs/installation/upgrading/ingress-class-compatibility.md
+++ b/content/en/docs/installation/upgrading/ingress-class-compatibility.md
@@ -28,7 +28,7 @@ create an `IngressClass` object matching the field to work.
 cert-manager aims to be compatible with as many ingress controllers as possible. According to the
 Ingress v1 [Kubernetes enhancement proposal], the deprecated annotation, if present, takes
 precedence over the new field. From our perspective, the option that maintains the highest
-compatibility is to only use the annotation, even when creating `v1` Ingresses
+compatibility is to only use the annotation, even when creating `v1` Ingresses.
 
 [Kubernetes enhancement proposal]: https://github.com/kubernetes/enhancements/tree/44dd2975dc6cdad96ca73e7b0ba1794f1196f604/keps/sig-network/1453-ingress-api#interoperability-with-previous-annotation
 
@@ -47,13 +47,13 @@ If you chose not to use the IngressClass `nginx` that is created by default by t
 ## Istio
 
 If you are using Istio and you had to create an IngressClass while migrating to cert-manager 1.5 or 1.6
-and that you chose to create an IngressClass that isn't named `istio` (e.g., you named it `istio-internal`),
+and you chose to create an IngressClass that isn't named `istio` (e.g., you named it `istio-internal`),
 you will need to change the `class` field on those Issuers back to `istio`.
 
 ## Traefik
 
 If you are using Traefik and you had to create an IngressClass while migrating to cert-manager 1.5
-or 1.6 and that the IngressClass you created isn't named `traefik` (for example, you called
+or 1.6 and the IngressClass you created isn't named `traefik` (for example, you called
 the IngressClass `traefik-external`), you will need to add a command-line argument to your
 Traefik deployment:
 
@@ -63,6 +63,6 @@ Traefik deployment:
 
 ## Ambassador
 
-If you are using Ambassador and found that you had to create an IngressClass while migrating to
-cert-manager 1.5 or 1.6, and that the IngressClass you created isn't named `ambassador`
+If you are using Ambassador and you had to create an IngressClass while migrating to
+cert-manager 1.5 or 1.6, and the IngressClass you created isn't named `ambassador`
 (e.g., `ambassador-internal`), you will need to change the `class` field on the affected Issuers back to `ambassador`.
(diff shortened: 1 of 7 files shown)


wallrj (changes requested):
Could you also add a note to https://cert-manager.io/next-docs/installation/upgrading/upgrading-1.6-1.7/ with a link to this information.

jakexks:
I think there should be upgrade notes for 1.5, 1.6 and 1.7, I shall add them now

content/en/docs/installation/upgrading/ingress-class-compatibility.md
```diff
@@ -0,0 +1,68 @@
...
+create an `IngressClass` object matching the field to work.
+
+cert-manager aims to be compatible with as many ingress controllers as possible. According to the
+Ingress v1 [Kubernetes enhancement proposal], the deprecated annotation, if present, takes
+precedence over the new field. From our perspective, the option that maintains the highest
+compatibility is to only use the annotation, even when creating `v1` Ingresses
+
+[Kubernetes enhancement proposal]: https://github.com/kubernetes/enhancements/tree/44dd2975dc6cdad96ca73e7b0ba1794f1196f604/keps/sig-network/1453-ingress-api#interoperability-with-previous-annotation
+
+# Notes For Specific Ingress Controllers
```
maelvls, line 35: People reading this page in a "standalone" way (without reading the 1.7 release notes first) won't know whether they should read the rest of the document or not. I think we should explain if they are affected or not by adding which cert-manager versions are affected.

jakexks: https://github.com/cert-manager/website/pull/812/files#diff-0253be5163bbb956f00cbf9d08ea6281c91d9b75ea5cd83ff331ff0c256d6923R8 

It is at the top of the page, but it may not be good enough. Do you have a suggestion?

content/en/docs/installation/upgrading/ingress-class-compatibility.md
```diff
@@ -0,0 +1,68 @@
...
+During the 1.5 and 1.6 cert-manager release cycles, we discovered that ingress controllers have
+handled the graduation of Ingress to `v1` differently. Some treat the class as an opaque string,
+similarly to the annotation. Some were unintentionally broken, as their default ingress class name
+contains characters that are disallowed in object references, e.g. (`/`). Some now require you to
+create an `IngressClass` object matching the field to work.
+
+cert-manager aims to be compatible with as many ingress controllers as possible. According to the
+Ingress v1 [Kubernetes enhancement proposal], the deprecated annotation, if present, takes
+precedence over the new field. From our perspective, the option that maintains the highest
+compatibility is to only use the annotation, even when creating `v1` Ingresses
```
maelvls, line 31:
```suggestion
compatibility is to only use the annotation, even when creating `v1` Ingresses.
```

wallrj (approved):
Nice clear description of the regression and the upgrade steps.

I spotted a couple of typos (I think) but ignore those if you prefer to get this merged.

content/en/docs/installation/upgrading/ingress-class-compatibility.md
maelvls:
```suggestion
# Notes For Specific Ingress Controllers

The regression is only present in cert-manager 1.5.0, 1.5.1, 1.5.2, 1.5.3, 1.5.4, 1.6.0, 1.6.1, 1.6.2, and 1.6.3, and applies to clusters running Kubernetes 1.19+ when using an Issuer or ClusterIssuer with an ACME HTTP-01 solver configured. 

If you are upgrading from one of the above versions to 1.5.5+, 1.6.4+ or 1.7.0+, and that you are in the following case:

- You are using Gloo, Contour, Skipper, or kube-ingress-aws-controller;
- You are using Traefik, Istio, Ambassador, or ingress-nginx with their default "class" (e.g., `istio` for Istio);

then upgrading from a version that contains the regression to 1.7.0, 1.6.4 or 1.5.5 won't have any impact and you can skip the rest of this document.

If you are using an ingress controller that doesn't belong to the above, upgrading from a version that contains the regression to 1.7.0, 1.6.4 or 1.5.5 may have an impact. The rest of this document details the measures you will need to take.

```

maelvls:
```suggestion
# Notes For Specific Ingress Controllers

When upgrading from one of the affected versions of cert-manager (1.5.0, 1.5.1, 1.5.2, 1.5.3, 1.5.4, 1.6.0, 1.6.1, 1.6.2, and 1.6.3) to 1.5.5, 1.6.4 or 1.7.0 and above, and that you are in one or multiple of the following cases:

- You are using Gloo, Contour, Skipper, or kube-ingress-aws-controller,
- You are using Traefik, Istio, Ambassador, or ingress-nginx with their default "class" (e.g., `istio` for Istio),

then you don't need to worry and the upgrade will have no impact. You can skip the rest of this document.

If you are using an ingress controller that doesn't belong to the above, upgrading from a version that contains the regression to 1.5.5, 1.6.4 or 1.7.0 and above may have an impact. The rest of this document details the measures you will need to take.

```

maelvls: I am OK for moving this suggestion to a new PR.

PR #815, opened 2022-01-26
Update release-notes with latest output of the release-notes tool

wallrj:
I re-ran the `release-notes` tool during the 1.7 release and it picked up the following change.


wallrj:
@irbekrm This is the change you made to that PR release-note but I removed the Bugfix prefix since it does seem like feature.

jakexks (approved):
/lgtm

irbekrm:
> @wallrj I changed it from feature to bugfix because it was not actually a new feature 

It's probably not super-important, but marking it as bugfix would make it easier to understand in which release this got fixed (like in case of the person who commented on that PR and had ran into the bug with 1.6.1)

wallrj:
I changed it back to bugfix. The problem was that the PR originally had both `kind/feature` and `kind/bug` so the release-notes tool chose feature. PTAL

PR #826, opened 2022-01-27
Add the missing entries to the release notes 1.5 and 1.6

maelvls:
- I had to use `##` (section level 2) instead of `#` (section level 1) because the section level 1 is meant for titles and does not show in the table of contents.
- I reviewed all the change log entries and rephrased the ones that I could not understand. I also updated the `release-note` block for those PRs.

PR #830, opened 2022-01-30
release-process: disambiguate the example 3 "final release"

maelvls:
From https://github.com/cert-manager/website/pull/826#discussion_r795197260.

### DIFF
--- a/content/en/docs/contributing/release-process.md
+++ b/content/en/docs/contributing/release-process.md
@@ -235,7 +235,7 @@ page if a step is missing or if it is outdated.
         |                   | initial alpha    | subsequent alpha | beta release     | final release | patch release |
         |                   | `v1.3.0-alpha.0` | `v1.3.0-alpha.1` | `v1.3.0-beta.0`  | `v1.3.0`      | `v1.3.1`      |
         |                   |                  |                  |                  |               |               |
-        | `START_REV`\*     | `v1.2.0`         | `v1.3.0-alpha.0` | `v1.3.0-alpha.1` | `v1.2.0`      | `v1.3.0`      |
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm

seems reasonable to me, thanks!

content/en/docs/contributing/release-process.md
```diff
@@ -261,7 +266,8 @@ page if a step is missing or if it is outdated.
         ```sh
         # Must be run from the cert-manger folder.
         export GITHUB_TOKEN=*your-token*
-        go install k8s.io/release/cmd/release-notes@v0.7.0
```
maelvls, line 264: I removed this line because we already install the tool in the first 3 steps of the release process.

content/en/docs/contributing/release-process.md
```diff
@@ -261,7 +266,8 @@ page if a step is missing or if it is outdated.
...
+        git fetch origin $BRANCH:$BRANCH
+        export START_SHA="$(git rev-list --reverse --ancestry-path $START_TAG..$BRANCH | head -1)"
```
maelvls, line 270: This won't work when releasing v1.8.0. Let's imagine I am about to release v1.8.0 and that I want to generate the release notes. Since v1.7.0 does not belong to `master`, the command:

```sh
git rev-list --reverse --ancestry-path v1.7.0..master | head -1
```

returns nothing.

To work around the fact that v1.7.0 does not belong to master, we could consider the "base" commit between master and v1.7.0:

```sh
git rev-list --reverse --ancestry-path $(git merge-base v1.7.0 master)..master | head -1
```

I will update the release process page.

PR #837, opened 2022-02-04
Release notes and installation links for for v1.7.1

wallrj:
Release notes and updated installation links for https://github.com/cert-manager/cert-manager/releases/tag/v1.7.1

I've replaced some jetstack/cert-manager with cert-manager/cert-manager, but only where they relate to the download of versioned assets.

We can replace the remaining references in another PR.

I also updated the `cmctl` and `kubectl cet-manager` installation instructions to use a specific version rather than `/latest/`, which may not point to the newest semver version, but simply the last released version.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -352,6 +352,7 @@ v1.6.1
 v1.6.2
 v1.7
 v1.7.0
+v1.7.1
 v1alpha2
 v1alpha3
 v1beta1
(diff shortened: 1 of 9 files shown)


SgtCoDFish (approved):
/lgtm
/approve

Thanks!

PR #846, opened 2022-02-10
Ssa installation warning

irbekrm:
Adds a note about the known issue when upgrading to cert-manaer 1.7 using kubectl server side apply and how to fix it

See https://github.com/cert-manager/cert-manager/issues/4831 for context

If this gets merged, I will sync the update to GitHub release note too.

Also fixes the cert-manager version at which the alpha/beta APIs were deprecated that I fixed on GitHub some time ago, but not here

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -147,6 +147,7 @@ SelfSigned
 SgtCoDFish
 Smallstep
 SubjectAccessReview
+stevehipwell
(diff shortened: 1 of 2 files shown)


SgtCoDFish (approved):
/lgtm
/approve

Nice one :+1:

PR #848, opened 2022-02-11
Add redirects for "project" documentation

SgtCoDFish:
For some sections - e.g. contributing, about jetstack, release notes, upgrading docs - there's no reason that someone would want to consume older versions of the docs. Only the latest version is relevant.

This change adds redirects for each of these older docs pages, pointing to the latest version of each page.

**NOTE FOR REVIEWERS**: The netlify preview has broken docs for all old docs sites. This is common for all website PRs and isn't caused by _this_ PR specifically.


jahrlin (approved):
Makes sense to me!

Holding in case someone else wants to take a look, unhold at your convenience

/lgtm
/approve
/hold

PR #849, opened 2022-02-11
Check all docs into the default branch

SgtCoDFish:
## Change Notifications / Pleas

This treats release-next the same as it treats old release docs. That means if we change something in `docs` we need to remember to change it in `next-docs` too, or else they'll become out-of-sync.

I know having to remember this is a burden but allowing this makes things _so much easier_ as we transition towards the website refresh. We also don't use release-next much except when doing a release, and we've got a while to go before the next one, giving us a good chunk of time to get a better solution in place. I'm willing to own this and keep an eye on website PRs.

Essentially, I'm asking if we can accept this in the name of pragmatism now to make life easier during the refresh. I want to come back to the build process and try to make improvements, but now isn't the time for it with the refresh coming.

## Description

Previously, each version was stored on its own branch, but only the markdown files from each branch were actually used from each branch to construct the final website.

Now, we commit all the files into the default branch (master at the time of writing). This simplifies the build and simplifies changes to older docs. See [this doc](https://docs.google.com/document/d/14BWiKcBsvsbahtp7rcrZ0XBujkslQPxiZv4pq7nkGXs/edit#) for more information on the approach.

Some files from older branches are not included; these are broadly "project" documentation as opposed to "product" documentation, and will have redirects in place from #848.

The bulk of the work was done through via the following, in order:

### 1. Edit `content/en/.gitignore`

This un-ignores the files we want to check in to master.

### 2. Copy files to master

Run `./scripts/build-release` one last time

### 3. Run cleanup script

The following script will delete any files for old releases for which redirects are in place thanks to #848:

```bash
#!/usr/bin/env bash

set -eu -o pipefail

find content/en/v* -name "contributing"  -type d | xargs rm -r
find content/en/v* -name "aboutjetstack" -type d | xargs rm -r
find content/en/v* -name "release-notes" -type d | xargs rm -r
find content/en/v* -name "upgrading"     -type d | xargs rm -r

find content/en/v* -name "supported-releases.md" | xargs rm

find content/en/v* -name ".gitignore"  | xargs rm
```

### 4. Update old links

This essentially does what #848 did, but replacing hyperlinks in pages (e.g. in 1.4 docs, don't link to `../supported-releases/`, but rather `/docs/installation/supported-releases/`.

Linking to `../supported-releases/` isn't actually a problem on the deployed site thanks to netlify redirects, but our htmltest check will fail on it.

The following script updates the links:

```bash
#!/usr/bin/env bash

set -eu -o pipefail

TARGET="content/en/v*"
SED="sed -i"

grep -lR "../../../../contributing/dns-providers/" $TARGET  | xargs $SED 's/..\/..\/..\/..\/contributing\/dns-providers\//\/docs\/contributing\/dns-providers\//g'
grep -lR "../../../contributing/dns-providers/" $TARGET  | xargs $SED 's/..\/..\/..\/contributing\/dns-providers\//\/docs\/contributing\/dns-providers\//g'
grep -lR "../../contributing/external-issuers" $TARGET  | xargs $SED 's/..\/..\/contributing\/external-issuers\//\/docs\/contributing\/external-issuers\//g'
grep -lR "../../installation/upgrading/upgrading-1.2-1.3/" $TARGET  | xargs $SED 's/..\/..\/installation\/upgrading\/upgrading-1.2-1.3\//\/docs\/installation\/upgrading\/upgrading-1.2-1.3\//g'
grep -lR "../supported-releases/" $TARGET | xargs $SED 's/..\/supported-releases\//\/docs\/installation\/supported-releases\//g'
grep -lR "../upgrading/upgrading-0.15-0.16/#helm" $TARGET | xargs $SED 's/..\/upgrading\/upgrading-0.15-0.16\/#helm/\/docs\/installation\/upgrading\/upgrading-0.15-0.16\/#helm/g'
grep -lR "../upgrading/upgrading-0.15-0.16/#issue-with-older-versions-of-kubectl" $TARGET | xargs $SED 's/..\/upgrading\/upgrading-0.15-0.16\/#issue-with-older-versions-of-kubectl/\/docs\/installation\/upgrading\/upgrading-0.15-0.16\/#issue-with-older-versions-of-kubectl/g'
```


wallrj:
/unhold because @SgtCoDFish explained what *I think* is the key motivation for this change, which is that the metadata in all the markdown files of current and past releases will need to be modified from the Hugo schema to a `next.js` schema. 
And this will be done (by hand) by the web agency who are carrying out the website redesign, so having all the past markdown content in `master` will make it easier for them to edit all that content in a single PR, 
Additionally, if we changed all the markdown files metadata to `next.js` scheme in the existing `release-` branches it would make those incompatible with the Hugo rendering scripts which are also versioned in those branches.
 
I am not persuaded that these are good enough reasons to commit all the old markdown into master and to commit further duplicate content to master on every release. I'd prefer we stuck with Hugo, but if mine is a minority opinion on that, I'm happy to live with it for the sake of getting the website refresh over the line.

jakexks (approved):
(I have reviewed the steps and commit messages rather than +118792!)

As far as I can see, the PR is technically sound, and the deploy preview all works. This has the advantage of making the website build much faster, at the disadvantage of having to duplicate edits between master / next-docs sometimes.

I think that tradeoff is OK, as the old flow for merging into release-next and back to master could have conflicts too.

If this unblocks progress on the website redesign, I think it should merge, as having massive PRs get out of sync leads to a lot more toil work that could be avoided.

The only thing I can think of to add in future is an extra website contributing page telling people how to document how to make changes.

/lgtm

PR #856, opened 2022-02-24
Remove a few broken tags which were introduced by mistake

SgtCoDFish:


jakexks (approved):
/lgtm

I wonder if a newline `<br />` was originally intended here but the deploy preview looks ok to me.

PR #865, opened 2022-03-18
Add a paragraph about releasing OLM packages

wallrj:
Eventually I should probably move all the OLM release documentation into a page here, but for now I explain the motivation for this release step and link to the README in the cert-manager-olm repo.


maelvls:
Thanks for updating the release process document!! 

/lgtm

PR #867, opened 2022-03-18
Proposal for Google Season of Docs

wallrj:
Mael and I sat down and sketched out a possible proposal for [Google Season of Docs](https://developers.google.com/season-of-docs/docs/admin-guide).

See also:
 * https://github.com/google/season-of-docs/pull/483

/cc @maelvls


content/en/docs/contributing/google-season-of-docs/improve-navigation-and-structure/index.md
```diff
@@ -0,0 +1,22 @@
...
+Nor will a Cluster Administrator easily find information about how to optimise cert-manager for a large cluster with many Certificates.
+The information exists but is spread across multiple pages and is often not at the obvious page.
+
+We would like a technical writer:
+1. to help us identify our target audiences and
+2. to identify the key tasks of each of these audiences and
+3. re-structure the cert-manager.io website with this in mind.
+
+For example, we have discussed the following audiences and tasks: Beginner, Cluster Administrator, User, Integrator, New Contributor
+and each of these people will be interested in a different set of tasks.
```
irbekrm, line 14: Question: does defining the audience in the docs (as opposed to just having a list of named tasks) matter?

(Had this thought after looking at https://linkerd.io/2.11/tasks/ where it seems like task names are already descriptive enough to help the reader understand whether that's something they're interested in doing)

content/en/docs/contributing/google-season-of-docs/improve-navigation-and-structure/index.md
```diff
@@ -0,0 +1,22 @@
+# Proposal: Improve the navigation structure of the cert-manager website
+
+Right now the content is not designed with our target audiences in mind.
```
irbekrm, line 3: Note to self: so this aims to introduce the 'use case based documentation' that has been talked about previously.

content/en/docs/contributing/google-season-of-docs/improve-navigation-and-structure/index.md
maelvls: I think the intent of this paragraph isn't to mandate the technical writer to define the audience in the docs, i.e., we don't need to have these audiences listed somewhere on the website. Instead, it is to study these audiences for the purpose of restructuring the website with these audiences in mind.

wallrj:
I think somethings broken in the build scripts `F0324 15:58:18.120451     241 main.go:135] no API packages found in ./pkg/apis`

SgtCoDFish:
I didn't expect this to be a problem, but our API doc generation process doesn't like the renamed cert-manager repo. We never fast-forwarded the `release-1.8` branch on the cert-manager repo until recently, so the module path changing was "invisible" until just now.

I _think_ this might be because the tool we use - https://github.com/ahmetb/gen-crd-api-reference-docs - is not module aware and so we have to do some GOPATH setup for it to work. I'm looking for a workaround now.

SgtCoDFish:
Should be unblocked once #879 merges!

PR #875, opened 2022-03-22
 Remove old Helm v2 Tiller section

### LINKED ISSUE #600: Remove Helm v2 from Ingress Nginx tutorial
I think the later versions of `cert-manager` don't support Helm v2 so we may want to clean up the Ingress Nginx tutorial https://cert-manager.io/docs/tutorials/acme/ingress/

derba:
Fixes #600


SgtCoDFish:
/lgtm
/approve

I think the tutorials could still use a little more work but this is a great improvement that we should've made long ago. Thank you for doing it :star:

PR #879, opened 2022-03-24
Fix API doc generation for new import path

SgtCoDFish:
Our tool for generating API docs is not module aware, and so it chokes for versions of cert-manager >= 1.8 where the import path has changed and we consequently don't setup the GOPATH correctly, leading the API doc tool to think there aren't any API docs available.

This change is _intentionally_ brute force. It simply copies the script twice, one for the old import path and one for the new import path. There are certainly more elegant ways of doing this, but this is very simple and easy to understand and should unblock any PRs for the website for now without too much of an overhead.


wallrj (approved):
/approve
/lgtm

PR #884, opened 2022-03-27
remove duplicate code

azadious:


content/en/v1.7-docs/faq/sync-secrets.md
```diff
@@ -62,9 +62,8 @@ spec:
   secretTemplate:
     annotations:
       reflector.v1.k8s.emberstack.com/reflection-allowed: "true"  
-      reflector.v1.k8s.emberstack.com/reflection-allowed-namespaces: "dev,staging,prod"  # Control destination namespaces
-      reflector.v1.k8s.emberstack.com/reflection-auto-enabled: "true" # Auto create reflection for matching namespaces
```
irbekrm, line 66: Thanks for spotting the duplicate!
After having a brief read through this (I've not worked with reflector,  I think the duplicate is actually a typo and was meant to be `reflector.v1.k8s.emberstack.com/reflection-auto-namespaces` - then the comment would also make sense.

irbekrm (changes requested):
Thanks for the contribution!

I think the duplication might be a typo - see the comment

SgtCoDFish:
This has been stale for a long time, so I'll close. Feel free to re-open if you think this is still a problem!

PR #898, opened 2022-04-05
style: fix active style for docs in nav

C-Mejlak:
Active styling for docs in navigation
image

Correctly comment out using `{/* ... */}`
Before
image

After
image


SgtCoDFish:
Thanks for this, the pictures are great too.

Since this is so small (and since the comment fix is to `content/`) I'll merge it by hand to keep the next-website branch minimal :+1: Thanks again!

PR #902, opened 2022-04-11
Fully update supported releases page after 1.8 release

SgtCoDFish:
Despite the title I'm not actually _fully_ finished with editing this page - I think a chunk of it should be automated and that doing that automation would be valuable - but I think this is enough of an improvement to raise it now!

/assign @maelvls


SgtCoDFish:
Part of https://github.com/cert-manager/cert-manager/issues/5021

maelvls:
Thank you for updating the Supported release page and for finding an automated way of doing it.

PR #903, opened 2022-04-12
Don't run build script for old crd docs by default

SgtCoDFish:
This avoids re-running setup steps when the old versions are not being built. since building the old versions is a manual process anyway, we might as well add an extra step to that manual process and save time on our CI


jakexks (approved):
/lgtm

PR #910, opened 2022-04-13
Various post-release updates to scripts

SgtCoDFish:
Importantly this includes changes to the README to make it clear how the site can be contributed to in the new next.js world!

A potentially controversial change is that I've commented out `scripts/index` - we can't currently run it because we don't know the variables. I'm hoping that won't be a lasting change, but for now I'm trying to reflect reality.


scripts/bin/htmltest
```diff
@@ -40,14 +43,14 @@ if ! test -f "${htmltest}"; then
 
     tmpdir="$(mktemp -d)"
     if [ "$GOOS" = "darwin" ]; then
-        curl -Lo "${tmpdir}/file" "https://github.com/wjdp/htmltest/releases/download/v${VERSION}/htmltest_${VERSION}_osx_amd64.tar.gz"
-        check_sha "${tmpdir}/file" "761cc0ab30edf9c327b897cf3f2b5c6c8bb78c11a45a75092f23ed3f35df7759"
+        curl -Lo "${tmpdir}/file" "https://github.com/wjdp/htmltest/releases/download/v${VERSION}/htmltest_${VERSION}_macos_amd64.tar.gz"
```
jakexks, line 46: darwin/arm64 users are forever forgotten

jakexks (approved):
The changes make sense to me

/lgtm

README.md
```diff
@@ -1,116 +1,101 @@
...
+If you've already run `npm install`, you can manually run `npm run dev` as another option.
 
-You also need to ensure you have the following tool(s) installed on your system
-as they are not managed by the `scripts/` directory:
+In any case, a server will spin up at `http://localhost:3000`.
 
-* npm/nodejs
-* Golang 1.12+
+Initial builds of a page on the development server can be quite slow - a few seconds - but
+after the initial build changes should be picked up quickly and the development server
```
jsoref, line 41: after the initial build, changes should be picked up quickly and the development server

README.md
```diff
@@ -1,116 +1,101 @@
...
+./scripts/gendocs/generate
 ```
 
-This will automatically run a number of checks against your changes.
+Since there are many old versions of cert-manager, none of which change regularly (or at all),
+the website build process does not re-generate documentation for older versions, on the assumption
+that doing so would be a waste of effort.
+
+The solution for achieving this is simple; the generation scripts for older cert-manager versions
+are commented out. To rebuild, uncomment them and then re-comment after.
```
jsoref, line 88: To regenerate documentation for older versions...

scripts/_verify
```diff
@@ -18,11 +18,15 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-REPO_ROOT="${REPO_ROOT:-$(cd "$(dirname "$0")/.." && pwd)}"
+# NOTE: Usually you'd not run this directly.
```
jsoref, line 21: Usually you wouldn't run...

README.md
SgtCoDFish: Yeah would maybe be better with the comma but probs not worth changing; thanks for the suggestion though!

README.md
SgtCoDFish: I feel this is clear enough as-is; thanks for the suggestion though!

scripts/_verify
SgtCoDFish: This is a stylistic choice - I'd say "you'd not run" like this in speech, personally, and I think it's understandable; thanks for the suggestion though!

PR #916, opened 2022-04-14
Various updates to nginx ingress tutorial

SgtCoDFish:
This fixes a bunch of language, tweaks links, and generally makes improvements. The tutorial isn't finished by any means, but this feels like a bunch of incremental improvements.

Also changes links on the tutorial from cert-manager.io to github user content - the links were broken by the refresh, and this patches over that (and might actually be an improvement, anyway)

Also removes the leading `$` on several install commands for the main docs page, which makes the new copy button on the refreshed website much more useful. That's not replicated on older docs versions because there are links which need sorting on those.

### DIFF
--- a/content/docs/installation/README.md
+++ b/content/docs/installation/README.md
@@ -14,7 +14,7 @@ install methods are listed below for each of the situations.
 The default static configuration can be installed as follows:
 
 ```bash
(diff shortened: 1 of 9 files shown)


maelvls:
This looks great to me. I very much like the fact you added a few words about what an "issuer" and "certificate" is, since people may not be aware of those when getting started.

/lgtm

PR #919, opened 2022-04-15
gsod2022: add a title, and tell that we were accepted

maelvls:
cc @wallrj


maelvls:
If that's OK, I'll go ahead and merged this PR since people are asking me how to apply since last Saturday. cc @wallrj

Issue #559, opened 2021-05-12
Update uninstall process doc


irbekrm:
There have been a number of questions related to failed uninstalls after attempting to uninstall by deleting `cert-manager` namespace see [cert-manager#3986](https://github.com/jetstack/cert-manager/issues/3986), [cert-manager#3474](https://github.com/jetstack/cert-manager/issues/3474) [cert-manager#3870](https://github.com/jetstack/cert-manager/issues/3870).

We already have some notes on how to recover from this scenario [here](https://cert-manager.io/docs/installation/uninstall/kubernetes/#namespace-stuck-in-terminating-state), but we could update that section with how to manually delete pending custom resources (as mentioned in the above linked issues).

irbekrm:
Thanks for updating the doc! I've added a couple more comments to #920.

PR #920, opened 2022-04-18
Update uninstall documentation

### LINKED ISSUE #860: Update uninstall documentation
At the moment our uninstall docs suggest that to uninstall cert-manager all cert-manager resources need to be deleted https://cert-manager.io/docs/installation/kubectl/#uninstalling. In reality this may not be desirable in cases where folks aim to keep the CRDs in th cluster and want to uninstall and reinstall, and would rather not lose their resources.

We should:
- update the docs to explain in which cases the resources need to be deleted before uninstalling (how to avoid getting into the problem with finalizers, see i.e [cert-manager#3851](https://github.com/cert-manager/cert-manager/issues/3851)
- Also add some note about what can potentially happen with `Secret`s _if_ the `Certificate`s have been deleted, see the discussion on https://github.com/cert-manager/cert-manager/issues/4788

enjeck:
Closes #860 and #559

### DIFF
--- a/content/docs/installation/kubectl.md
+++ b/content/docs/installation/kubectl.md
@@ -50,16 +50,25 @@ $ kubectl create clusterrolebinding cluster-admin-binding \
 > and potentially broken states. Please ensure you follow the below steps when
 > uninstalling to prevent this happening.
 
-Before continuing, ensure that all cert-manager resources that have been created
+Before continuing, ensure that unwanted cert-manager resources that have been created
 by users have been deleted. You can check for any existing resources with the
 following command:
 
 ```bash
 $ kubectl get Issuers,ClusterIssuers,Certificates,CertificateRequests,Orders,Challenges --all-namespaces
 ```
+It is recommended that you delete all these resources before uninstalling cert-manager. 
+If plan on reinstalling later and don't want to lose some custom resources, you can keep them. 
+However, this can potentially lead to problems with finalizers. Some resources, like 
+`Challenges`, should be deleted to avoid [getting stuck in a pending state](#namespace-stuck-in-terminating-state). 
 
-Once all these resources have been deleted you are ready to uninstall
+Once the unneeded resources have been deleted, you are ready to uninstall
 cert-manager using the procedure determined by how you installed.
+
+> **Warning**: Uninstalling cert-manager or simply deleting a `Certificate` resource can result in 
+> TLS `Secret`s being deleted if they have `metadata.ownerReferences` set by cert-manager.
+> You can disable this behavior by setting `-enable-certificate-owner-ref=false` before deleting
+> a `Certificate` or `cert-manager`.
 
 ### Uninstalling with regular manifests
 
@@ -87,4 +96,24 @@ experiencing issues then run:
 
 ```bash
 $ kubectl delete apiservice v1beta1.webhook.cert-manager.io
+```
+
+#### Deleting pending challenges
+
+`Challenge`s can get stuck in a pending state when the finalizer is unable to complete 
+and Kubernetes is waiting for the cert-manager controller to finish. 
+This happens when the controller is no longer running to remove the flag, 
+and the resources are defined as needing to wait.
+You can fix this problem by doing what the controller does manually. 
+
+First, delete existing cert-manager webhook configurations, if any:
+
+```bash
+$ kubectl delete validatingwebhookconfigurations cert-manager-webhook
+$ kubectl delete validatingwebhookconfigurations cert-manager-webhook
+```
+Then change the `.metadata.finalizers` field to an empty list by editing the challenge file:
+
+```bash
+$ kubectl edit challenge <the-challenge>
 ```
--- a/.spelling
+++ b/.spelling
@@ -236,6 +236,8 @@ erikgb
 errored
 erroring
 external-dns
+finalizer
+finalizers
 foosinn
 francescsanjuanmrf
 freeipa-issuer


enjeck:
Does it take a while for the DCO to sign off? I just ran `git rebase --signoff <current-branch>` and has nothing changed yet...

enjeck:
Yes! Finally figured out how to signoff. Thanks, @jakexks

jakexks:
Hi there,

The verify job has failed because the spell checker (which isn't amazing) has flagged finalizer / finalizsers. I believe you should be able to add them to https://github.com/cert-manager/website/blob/master/.spelling in this PR.

jakexks (approved):
/lgtm
thanks!

content/docs/installation/kubectl.md
```diff
@@ -50,17 +50,26 @@ $ kubectl create clusterrolebinding cluster-admin-binding \
...
-Before continuing, ensure that all cert-manager resources that have been created
+Before continuing, ensure that unwanted cert-manager resources that have been created
 by users have been deleted. You can check for any existing resources with the
 following command:
 
 ```bash
 $ kubectl get Issuers,ClusterIssuers,Certificates,CertificateRequests,Orders,Challenges --all-namespaces
 ```
+It is recommended that you delete all these resources before uninstalling cert-manager. 
+If plan on reinstalling later and don't want to lose some custom resources, you can keep them.
```
irbekrm, line 61: You cannot keep the custom resources if you delete the `CustomResourceDefinition`s as per the following steps in these instructions. Users may choose to manage `CustomResourceDefinition`s separately in which case they don't need to delete custom resources - it might be worth to add a note about that somewhere in the doc.

content/docs/installation/kubectl.md
```diff
@@ -50,17 +50,26 @@ $ kubectl create clusterrolebinding cluster-admin-binding \
...
 ```bash
 $ kubectl get Issuers,ClusterIssuers,Certificates,CertificateRequests,Orders,Challenges --all-namespaces
 ```
+It is recommended that you delete all these resources before uninstalling cert-manager. 
+If plan on reinstalling later and don't want to lose some custom resources, you can keep them. 
+However, this can potentially lead to problems with finalizers. Some resources, like 
+`Challenges`, should be deleted to avoid [getting stuck in a pending state](#namespace-stuck-in-terminating-state). 
 
-Once all these resources have been deleted you are ready to uninstall
+Once the unneeded resources have been deleted, you are ready to uninstall
```
irbekrm, line 65: Same as above- if we assume that CRDs are deleted there is no way to keep custom resources. If CRDs are managed separately there will be no need to delete custom resources.

content/docs/installation/kubectl.md
```diff
@@ -88,3 +97,23 @@ experiencing issues then run:
...
+and Kubernetes is waiting for the cert-manager controller to finish. 
+This happens when the controller is no longer running to remove the flag, 
+and the resources are defined as needing to wait.
+You can fix this problem by doing what the controller does manually. 
+
+First, delete existing cert-manager webhook configurations, if any:
+
+```bash
+$ kubectl delete validatingwebhookconfigurations cert-manager-webhook
+$ kubectl delete validatingwebhookconfigurations cert-manager-webhook
```
irbekrm, line 113: I think this should be `kubectl delete mutatingwebhookconfigurations cert-manager-webhook`

enjeck:
@irbekrm Whoops, this has already been merged. I'll address your review in a new PR

enjeck:
@irbekrm Or maybe you can revert the PR?

content/docs/installation/kubectl.md
```diff
@@ -50,17 +50,26 @@ $ kubectl create clusterrolebinding cluster-admin-binding \
...
+However, this can potentially lead to problems with finalizers. Some resources, like 
+`Challenges`, should be deleted to avoid [getting stuck in a pending state](#namespace-stuck-in-terminating-state). 
 
-Once all these resources have been deleted you are ready to uninstall
+Once the unneeded resources have been deleted, you are ready to uninstall
 cert-manager using the procedure determined by how you installed.
 
+> **Warning**: Uninstalling cert-manager or simply deleting a `Certificate` resource can result in 
+> TLS `Secret`s being deleted if they have `metadata.ownerReferences` set by cert-manager.
+> You can disable this behavior by setting `-enable-certificate-owner-ref=false` before deleting
```
irbekrm, line 70: Suggestion: `-enable-certificate-owner-ref=false` will ensure that the owner reference is not added to a `Secret`, it does not disable the garbage collection as such. If the flag was set previously set to true and the user sets it to false, the owner references will be eventually removed, but this is only the case since cert-manager v1.8 (see https://github.com/cert-manager/cert-manager/pull/4888).
Perhaps we could add a little bit more detail to this warning- mention that `-enable-certificate-owner-ref=false` will ensure that owner refs are removed in cert-manager 1.8+ (I know that these docs are for 1.8 however I think it's possible that someone using an older version uninstalls by looking at the latest docs) and suggest that users check the owner references to ensure that they actually are removed. Also worth mentioning that `-enable-certificate-owner-ref` is set to false by default.

content/docs/installation/kubectl.md
```diff
@@ -88,3 +97,23 @@ experiencing issues then run:
...
+and the resources are defined as needing to wait.
+You can fix this problem by doing what the controller does manually. 
+
+First, delete existing cert-manager webhook configurations, if any:
+
+```bash
+$ kubectl delete validatingwebhookconfigurations cert-manager-webhook
+$ kubectl delete validatingwebhookconfigurations cert-manager-webhook
+```
+Then change the `.metadata.finalizers` field to an empty list by editing the challenge file:
```
irbekrm, line 115: Suggestion: `Then change the `.metadata.finalizers` field to an empty list by editing the challenge resource`.

irbekrm:
Thanks for working on this @enjeck !

I think another PR would be good 👍🏼

PR #937, opened 2022-04-25
add info on GKE Autopilot regarding kube-system namespace

jahrlin:
Signed-off-by: Joakim Ahrlin <joakim.ahrlin@gmail.com>

Fixes https://github.com/cert-manager/cert-manager/issues/5061

### DIFF
--- a/content/docs/installation/compatibility.md
+++ b/content/docs/installation/compatibility.md
@@ -55,6 +55,33 @@ out version 1.21 for Kubernetes masters. Installation via the helm chart
 may end in an error message but cert-manager is reported to be working by
 some users. Feedback and PRs are welcome.
 
+**Problem**: GKE Autopilot does not allow modifications to the `kube-system`-namespace.
+
+Installing cert-manager in these environments with default configuration can cause issues with bootstrapping.
+Some signals are:
+
+* `cert-manager-cainjector` logging errors like:
+
+```text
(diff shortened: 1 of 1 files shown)


content/docs/installation/compatibility.md
```diff
@@ -55,6 +55,35 @@ out version 1.21 for Kubernetes masters. Installation via the helm chart
...
+```
+
+**Solution**: Configure cert-manager to use a different namespace for leader election, like this:
+
+```console
+helm install \
+  cert-manager jetstack/cert-manager \
+  --namespace cert-manager \
+  --create-namespace \
+  --version v1.8.0 --set global.leaderElection.namespace=cert-manager
```
SgtCoDFish, line 84: suggestion: it scares me a little to add another place where we need to update the cert-manager version when a new version is released. It's already a bit of a messy manual job.

In this use case would it be OK to use `v<version-name>` instead of `v1.8.0`?

(this kind of thing should be automated for `/docs` but I don't think we're there yet)

SgtCoDFish (approved):
/lgtm
/approve
/hold

Adding a hold because I have a suggestion, but I'd be happy to merge.

content/docs/installation/compatibility.md
jahrlin: Good point! I'll put a placeholder there

SgtCoDFish (approved):
/lgtm
/approve

Thanks :+1:

PR #940, opened 2022-04-26
Update venafi.md

sabixx:
The provided zone example was misleading. One would assume that he does not need the full policy path with \VED\Policy\...


SgtCoDFish (approved):
/lgtm
/approve

Thanks for the improvement, much appreciated :grin:

Issue #934, opened 2022-04-25
Review opengraph / socials for site


SgtCoDFish:
Check we have an image and appropriate title / description for each page

Prompted by a question from Richard Collins at Jetstack.

PatrickHeneise:
A few tools and services that help with that:

- https://og-image.xyz
- https://www.bannerbear.com
- https://og-image.vercel.app

PR #950, opened 2022-04-27
Document how to work on the site a little better

SgtCoDFish:


README.md
```diff
@@ -25,7 +25,53 @@ The following tool(s) must be installed on your system prior to developing the w
...
+document you're working on and should point at a named markdown file.
+
+For example, take [this snapshot](https://raw.githubusercontent.com/cert-manager/website/d398905baef9841590fab6c2b854b74f0eecb006/content/docs/concepts/certificate.md)
+of the "Certificate Concepts" page.
+
+There's an external link to an RFC which is fully specified as expected:
+
+> `see [RFC 5246 section 7.4.2](https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.2)`
+
+But towards the end of the page we link to the "Certificate Usage" page:
```
irbekrm, line 53: Praise: I've previously spent ages searching through the docs for an example of how to link a file in the same repo as I always forget how to do that. Adding this example is absolutely awesome

README.md
```diff
@@ -25,7 +25,53 @@ The following tool(s) must be installed on your system prior to developing the w
 We also assume you've got GNU coreutils installed, which is usually the case by default on Linux
 and should be installable via Homebrew on macOS.
 
-## Website Development
+## Website Development Guides
+
+### Where's the documentation content?
+
+All docs go under `content/`; most changes are usually made in `content/docs` which is conceptually
+the "latest" version of the docs and includes other sections which don't make sense to version. For
```
irbekrm, line 33: Question: is it not the case that currently for the stuff for which it _does_ make sense to be versioned /content/docs should _only_ include documentation relevant for latest released cert-manager version- should this be explicitly stated here (perhaps as well as the fact that documentation aimed at a next release should be PR-ed against next-docs?
(This has been discussed multiple times, however it's late and my brain fails)

irbekrm (commented):
Thanks for working on this, I think this will be a great improvement and will be useful for me as I always struggle to find a 'local' link example.
I've added a question/optional suggestion apart from that happy to lgtm

README.md
SgtCoDFish: Fair comment yeah - I think this might look better with less words, I'll take another crack at it

SgtCoDFish:
@irbekrm I've updated the "where to put stuff" section as more of a question and answer format, and added a section on editing social sharing tags too :+1:

README.md
```diff
@@ -25,7 +25,61 @@ The following tool(s) must be installed on your system prior to developing the w
...
+  Add it to `docs/` and possibly to the specific version of cert-manager that's latest (e.g. `v1.8-docs/`)
+- Something which only applies to the next major version of cert-manager? <br />
+  Add it to `next-docs`
+- Something which isn't "versioned", e.g. a page under "contributing", release notes or our supported-releases page? <br />
+  Add it to `docs/`, which is the only place such pages should appear
+- Something which applies only to versions of cert-manager which have already been released? <br />
+  Change specific version docs (e.g. `v1.8-docs/`)
+
+If you're not sure, _ask_! Reach out to us on [slack](https://cert-manager.io/docs/contributing/#slack) and
+we'll point you in the right direction!
```
irbekrm, line 47: Praise: this is great and reads very clear.

README.md
```diff
@@ -79,6 +79,15 @@ site you're working on.
 For example, the [manifest for the docs section](https://github.com/cert-manager/website/blob/master/content/docs/manifest.json)
 contains the expected path for every file.
 
+### Task: Changing OpenGraph / social sharing tags
```
irbekrm, line 82: I don't really know what OpenGraph and these tags are. However, I can google that if needed, else may also be valuable to ad some comments to the javascript in ./pages (not in this PR)

irbekrm (approved):
Looks good to me, thank you for making these improvements!

/lgtm

README.md
SgtCoDFish: Yeah fair enough really! I've got a task to do more on this in #934 so I might add more docs there!

PR #972, opened 2022-05-06
Remove a header which previously had autofilled hugo content

SgtCoDFish:
We could maintain a hand-rolled list here, but the sidebar already has
all the navigation information. In any case, these headings are
currently pure noise and need to be changed from their current state.


wallrj (approved):
I agree that these (now) content-free headed sections are extremely confusing but I think I'd like  some closing remark that invites the reader to:
> explore the available issuers in the left-hand menu

But that might be something to discuss in the GSoD project.

/lgtm

PR #973, opened 2022-05-06
Update ingress-class-compatibility.md

interone-ms:
Per https://kubernetes.github.io/ingress-nginx/#how-to-easily-install-multiple-instances-of-the-ingress-nginx-controller-in-the-same-cluster, the nginx-ingress controller needs a bit more parameters set.

Signed-off-by: Marco Schuster <marco.schuster@interone.de>

### DIFF
--- a/content/docs/installation/upgrading/ingress-class-compatibility.md
+++ b/content/docs/installation/upgrading/ingress-class-compatibility.md
@@ -50,11 +50,21 @@ compatibility is to only use the annotation, even when creating `v1` Ingresses.
 ## ingress-nginx
 
 If you chose not to use the IngressClass `nginx` that is created by default by the Helm chart
-(e.g., you named the IngressClass `nginx-outside`), you will need to add the flag
-`--ingress-class` to your ingress-nginx deployment:
+(e.g., you named the IngressClass `nginx-outside`), you will need to add the flags
+`--ingress-class` and `--ingress-class-by-name` to your ingress-nginx deployment:
 
 ```
---ingress-class=nginx-outside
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
The documentation for ingressClassByName is lacking a little so it's been hard to review this :sweat_smile: 

I think [this](https://github.com/kubernetes/ingress-nginx/blob/fc58e92b3951bd313a36eba227933bb4e931cfbd/internal/ingress/controller/ingressclass/ingressclass.go#L48-L50) is the best definition there is. I've not seen much in the code to help.

In any case this change seems harmless so I'll approve - thank you for raising it! :grin: 

/lgtm
/approve

interone-ms:
Sorry for leaving the context out. For posterity:

1. `controller.ingressClassByName` causes the controller template (https://github.com/kubernetes/ingress-nginx/blob/4b4895b53bb5fd00874deeeb914d5d36fd19556e/charts/ingress-nginx/templates/_params.tpl) to add the flag `--ingress-class-by-name=true` to be added to the nginx controller pod.
2. The flag ends up recognized in the controller code at https://github.com/kubernetes/ingress-nginx/blob/89ed571d2a8c7c2486e9671808e27a7cfbe3b40d/internal/ingress/controller/store/store.go#L515, where it allows the controller to recognize Ingress objects that *only* have the old-style `kubernetes.io/ingress.class` annotation set.

SgtCoDFish:
Oh no need to apologise, I was struggling with the docs for ingress-nginx!

Thank you for the additional context though :heart:

PR #983, opened 2022-05-19
Document how to configure cert-manager via OLM

wallrj:
I've tried to document the OLM configuration options that are available today and mention some of the limitations.

Part of:
 * https://github.com/cert-manager/cert-manager-olm/issues/22
 * https://github.com/cert-manager/cert-manager/issues/4410

### DIFF
--- a/content/docs/installation/operator-lifecycle-manager.md
+++ b/content/docs/installation/operator-lifecycle-manager.md
@@ -93,22 +93,22 @@ If you have any issues with your installation, please refer to the
 
 ## Configuration
 
-The configuration options are quite limited you install cert-manager using OLM.
+The configuration options are quite limited when you install cert-manager using OLM.
 There are a few Deployment settings which can be overridden permanently in the Subscription
-and most other elements of the cert-manager manifests can be changed by patching the ClusterServiceVersion,
+and most other elements of the cert-manager manifests can be changed by editing the ClusterServiceVersion,
 but changes to the ClusterServiceVersion are temporary and will be lost if OLM upgrades cert-manager,
 because an upgrade results in a new ClusterServiceVersion resource.
 
 ### Configuration Via Subscription
 
-When you create an OLM Subscription you can override *some* of the cert-manager Deployment settings,
+When you create an OLM Subscription you can override **some** of the cert-manager Deployment settings,
 but the options are quite limited.
 The configuration which you add to the Subscription will be applied immediately to the current cert-manager Deployments.
 It will also be re-applied if OLM upgrades cert-manager.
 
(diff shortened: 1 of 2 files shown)


content/docs/installation/operator-lifecycle-manager.md
```diff
@@ -91,9 +91,146 @@ in accordance with [OLM's Recommended Channel Naming][].
...
+but the options are quite limited.
+The configuration which you add to the Subscription will be applied immediately to the current cert-manager Deployments.
+It will also be re-applied if OLM upgrades cert-manager.
+
+> 🔰 Read the [Configuring Operators deployed by OLM](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md#configuring-operators-deployed-by-olm) design doc in the OLM repository.
+>
+> 🔰 Refer to the [Subscription API documentation](https://pkg.go.dev/github.com/operator-framework/api@v0.14.0/pkg/operators/v1alpha1#Subscription).
+
+Here are some examples of configuration that can be achieved by modifying the Subscription resource.
+In each case we assume that you are starting with the following [default Subscription from OperatorHub.io]((https://operatorhub.io/install/cert-manager.yaml)):
```
jahrlin, line 114:
```suggestion
In each case we assume that you are starting with the following [default Subscription from OperatorHub.io](https://operatorhub.io/install/cert-manager.yaml):
```

content/docs/installation/operator-lifecycle-manager.md
```diff
@@ -91,9 +91,146 @@ in accordance with [OLM's Recommended Channel Naming][].
 If you have any issues with your installation, please refer to the
 [FAQ](../faq/README.md).
 
+## Configuration
+
+The configuration options are quite limited when you install cert-manager using OLM.
```
jahrlin, line 96: Just out of curiosity: are the options limited because of limitations with OLM or is it limitation we have imposed?

jahrlin (approved):
Great!

/lgtm

content/docs/installation/operator-lifecycle-manager.md
wallrj: The current version of OLM has limited support for modifying the deployment, as documented in this PR. 
For example, it does not provide an easy way for the OLM package author to allow the container command line flags to be overridden. 
But on the other hand, if all the cert-manager components could be configured using a config file in a configmap and if all the existing command line options could also be declared inthose config files, that particular problem would be solved.

There are some other limitations such as not being able to customize the resource requests of individual Deployments in the OLM package:
 * https://github.com/operator-framework/operator-lifecycle-manager/issues/1794

The current API seems to be designed around there only being a single Deployment in each package, but cert-manager has three.

PR #992, opened 2022-06-01
Initial feature gate documentation

### LINKED ISSUE #988: Document how feature gated fields can be added to API
See https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1653494694006459 for context

irbekrm:
This PR:

- attempts to define a loose feature policy just so that we can graduate some of the existing alpha features to beta and communicate to users what they should expect from alpha/beta features

- attempts to document the current mechanism how we implement feature gated API fields (for implementors)

All 'policy' bits are open to discussion, I've largely just pulled things from kube docs and what we discussed in the previous dev meeting

Fixes #988 


content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,48 @@
...
+- Ensure that the feature gate is added to cert-manager installation scripts for CI and local tests in [make](https://github.com/cert-manager/cert-manager/blob/134398e939bb2b1401697eaf589405ad469cd609/make/e2e-setup.mk#L165) and [bazel](https://github.com/cert-manager/cert-manager/blob/fd747b42b9ab4b6409b61b7946e8dc14d532e950/devel/addon/certmanager/install.sh#L26) scripts
+- Default cert-manger e2e CI tests run with all feature gates for all components enabled. There is an additional optional e2e test that runs with all feature gates disabled. You can trigger that for your PR with `/test pull-cert-manager-e2e-feature-gates-disabled` to verify that all works as expected both with and without the new feature gate.
+
+### Potential issues
+
+- A user has to remember to set two feature gates, one of the webhook one on the controller for the feature to function. Forgetting to set one of them might result in unexpected behavior
```
munnerz, line 34: I think it's worth noting that it's actually the *operator* or administrator, not the end-user that has to do this (depending on how we define these different roles/personas! 😄)

content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,48 @@
...
+
+- A user must remember to remove the alpha fields from their manifests when disabling a previously enabled API feature. Failing to do so might result in unexpected behavior- for example forgetting to remove an feature gated field from a `Certificate` resource might result in failed renewals at some later point when cert-manager's controller will attempt to update the `Certificate` spec, but the webhook will reject the update due to the feature gated field being set
```
munnerz, line 36: 👍 your idea of only gating on CREATE and not UPDATE sounds reasonable (though I need to consider/mull over any side-effects of this..). That said, this _is_ why we advise people to not use alpha features and is exactly why they are gated. We could remove the feature *altogether* in release N+1 as well.

irbekrm: Thanks I will also look into how CREATE vs UPDATE would look like

Agree that alpha could be less stable.. but I see a bunch of proposals/issues/PRs that would result in new alpha fields against API, some of which may be more popular than others.. so I think in reality we may end up with this mechanism being used quite a bit

content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,48 @@
...
+
+As of v1 release cert-manager is considered stable. We aim to follow Kubernetes API compatibility policy when making API changes, see [API compatibility](../installation/api-compatibility.md) to avoid breaking users' existing cert-manager installations.
+This means that as developers we are somewhat limited in regards to changing existing behavior, i.e renaming or removing API elements or changing their behavior.
+
+
+New functionality that is not yet stable[^1] can still be added, but it needs to be placed behind a feature gate.
+
+## Feature gated API fields
+
+Feature gated API fields are implemented using `--feature-gates` flags of cert-manager [webhook](../cli/webhook.md) and [controller](../cli/controller.md).
```
jsoref, line 14: Should they always have the same name? If so, it'd be good to spell that out.

content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,48 @@
...
+
+A feature gated API field is always visible to the user (i.e when running `kubectl explain <some-resource>`), but is only functional if the relevant feature is explicitly enabled via both feature flags.
```
jsoref, line 16:
```suggestion
A feature gated API field is always visible to the user (i.e when running `kubectl explain <some-resource>`), but is only functional if the relevant feature is explicitly enabled via feature flags for both the webhook and controller.
```

content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,48 @@
...
+
+If a user attempts to apply a resource with the feature gated field set to a non-nil value, but the feature gate is not enabled, the resource will get rejected by the webhook validation.
```
jsoref, line 18:
```suggestion
If a user attempts to apply a resource with the feature gated field set to a non-nil value, but the feature gate is not enabled, the resource will get rejected by webhook validation.
```

content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,48 @@
...
+This mechanism differs from [the one that Kubernetes uses for feature gated API field implementation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#new-field-in-existing-api-version) where the field will be simply set to nil if the feature gate is disabled. We chose to use webhook validation instead to make debugging easier for users who are attempting to use the feature gated field, but have forgotten to set the feature gate.
```
jsoref, line 19:
```suggestion
This mechanism differs from [the one that Kubernetes uses for feature gated API field implementation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#new-field-in-existing-api-version) where the field will be simply set to nil if the feature gate is disabled. We chose to use webhook validation instead to make debugging easier for users who are attempting to use the feature gated field, but have forgotten to enable the feature gate.
```

content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+### Implementation
+
+- Implement the new field and document that it is feature gated and in order to use it feature gates need enabling
```
jsoref, line 24:
```suggestion
- Implement the new field and document that it is feature gated and in order to use it the webhook and controller feature gates that need to be enabled
```

content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,48 @@
...
+- Add a new [webhook feature gate](https://github.com/cert-manager/cert-manager/blob/3a055cc2f56c1c2874807af4a8f84d0a1c46ccb4/internal/webhook/feature/features.go#L25-L39) for the field
+- Update webhook validation checks for the relevant resource kind to ensure that if the feature gated field is set, but the webhook feature gate is not enabled, the resource gets rejected
+- Add a new [controller feature gate](https://github.com/cert-manager/cert-manager/blob/2417132b3cd017b5f0974006e03c2b8a540efe3f/internal/controller/feature/features.go#L26-L54) for the field
+- Ensure that any control loops that use the feature, check that the feature gate is actually enabled. (This is required to cover edge cases such if webhook runs a version of cert-manager where the feature is in GA whereas controller runs an older version where the feature is still in experimental state)
```
jsoref, line 28:
```suggestion
- Ensure that any control loops that use the feature, check that the feature gate is actually enabled. (This is required to cover edge cases such as if the webhook runs a version of cert-manager where the feature is in GA whereas controller runs an older version where the feature is still in experimental state.)
```

content/docs/contributing/featuregates.md
jsoref, line 36:
```suggestion
- A user must remember to remove the alpha fields from their manifests when disabling a previously enabled API feature. Failing to do so might result in unexpected behavior- for example forgetting to remove a feature gated field from a `Certificate` resource might result in failed renewals at some later point when cert-manager's controller will attempt to update the `Certificate` spec, but the webhook will reject the update due to the feature gated field being set
```

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+A feature that is GA is on by default and cannot be disabled.
+With regards to API fields and their functionality, we keep Kubernetes API compatibility promise https://cert-manager.io/docs/installation/api-compatibility/.
+The feature flag for a GA feature will eventually be removed.
+
+## Graduation
+
+The graduation criteria can be different for each feature.
+Generally, we find user feedback most valuable when determining if a feature is sufficiently mature to graduate. If you are using an alpha or beta feature and would like to see it graduate, it would be great if you could give us some feedback about how you use it and whether you find the API useful to initiate graduation process. Feel free to [open a Github issue](https://github.com/cert-manager/cert-manager/issues/new/choose) or [join one of our meetings](https://cert-manager.io/docs/contributing/#meetings) to talk about this.
+
+## List of current feature gates
```
irbekrm, line 37: Inspiration from https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+Beta features will not be removed, but can be changed.
+If the feature gets changed in incompatible ways, we will provide migration instructions.
+A beta feature will work with all versions of Kubernetes supported by cert-manager.
+
+**GA**
+
+A feature that is GA is on by default and cannot be disabled.
+With regards to API fields and their functionality, we keep Kubernetes API compatibility promise https://cert-manager.io/docs/installation/api-compatibility/.
+The feature flag for a GA feature will eventually be removed.
+
```
irbekrm, line 31: This is largely adapted from https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages

content/docs/contributing/featuregates.md
irbekrm: This can be looked into separately and probably won't happen now when a bunch of things have already been implemented using the 2 flags approach

content/docs/contributing/featuregates.md
irbekrm: This is really up to the implementor

irbekrm:
Kube also has feature metrics apparently https://kubernetes.io/blog/2023/02/03/sig-instrumentation-spotlight-2023/#:~:text=We%E2%80%99ve%20also%20added%20feature%20metrics%20so%20that%20you%20can%20look%20at%20your%20metrics%20to%20determine%20which%20features%20are%20enabled%20in%20your%20cluster%20at%20a%20given%20time

content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Feature gated API fields
+
+Feature gated API fields are implemented using `--feature-gates` flags of cert-manager [webhook](../cli/webhook.md) and [controller](../cli/controller.md).
+
+A feature gated API field is always visible to the user (i.e when running `kubectl explain <some-resource>`), but is only functional if the relevant feature is explicitly enabled via feature flags for both the webhook and controller.
+
+If a user attempts to apply a resource with the feature gated field set to a non-nil value, but the feature gate is not enabled, the resource will get rejected by the webhook validation.
+This mechanism differs from [the one that Kubernetes uses for feature gated API field implementation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#new-field-in-existing-api-version) where the field will be simply set to nil if the feature gate is disabled. We chose to use webhook validation instead to make debugging easier for users who are attempting to use the feature gated field, but have forgotten to enable the feature gate.
```
jsoref, line 18: If an object is created using a feature gated thing while everything is enabled and then the controller/webhook are disabled, what happens when the controller restarts? Does anything flag for this?

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+New cert-manager features and functionality is often initially implemented behind a feature gate. This is so as to not break users with functionality that has not yet been tested in production as well as to give us a chance to remove or modify API fields and functionality following user feedback.
+
+We have alpha and beta features. We do not aim to keep any of the features in alpha or beta stage indefinitely. Feature gating should only be used for functionality that can eventually be turned on by default for all users (GA).
+
+A feature gate can be toggled on/off using `--feature-gates` flags on cert-manager controller. For feature gated functionality that comes with new API fields there is also a corresponding feature gate on webhook that also needs to be enabled using a `--feature-gates` flag if you want to use it.
+
+**Alpha**
+
+All alpha features are off by default.
+We keep the right to remove or change alpha features without a warning.
```
jsoref, line 15:
```suggestion
We retain the right to change or remove alpha features without warning.
```

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+An API field that is part of an alpha feature and requires a webhook feature flag to be used also can be removed from the API without a warning.
```
jsoref, line 16:
```suggestion
An API field that is part of an alpha feature and requires a webhook feature flag to be used can also be removed from the API without warning.
```

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+An alpha feature might not work for all versions of Kubernetes supported by cert-manager.
```
jsoref, line 17:
```suggestion
An alpha feature might not work for all of cert-manager's supported Kubernetes versions.
```

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+
+**Beta**
+
+All beta features are off by default.
+Beta features will not be removed, but can be changed.
```
jsoref, line 22:
```suggestion
Beta features will not be removed, but may be changed.
```

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+If the feature gets changed in incompatible ways, we will provide migration instructions.
+A beta feature will work with all versions of Kubernetes supported by cert-manager.
```
jsoref, line 24:
```suggestion
A beta feature will work with all versions of cert-manager's supported Kubernetes versions.
```

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+
+**GA**
+
+A feature that is GA is on by default and cannot be disabled.
+With regards to API fields and their functionality, we keep Kubernetes API compatibility promise https://cert-manager.io/docs/installation/api-compatibility/.
```
jsoref, line 29: Why is this a bare url?

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
+---
+title: Features
+description: cert-manager feature policy
+---
```
SgtCoDFish, line 4: nitpick: I don't feel like I've got a great suggestion here, but it feels like "features" doesn't quite match what this page describes.  Don't get me wrong - i love the idea of adding this page.

Maybe something like "Unstable Features", "Feature Flags" or something? I think as a user if I see "features" in the sidebar I might expect like a vendor pitch of what the product can do, and I'd skip it. If I saw something like "feature flags" I'd think "oh, more things to try!" and check it out.

What do you think?

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+
+New cert-manager features and functionality is often initially implemented behind a feature gate. This is so as to not break users with functionality that has not yet been tested in production as well as to give us a chance to remove or modify API fields and functionality following user feedback.
+
+We have alpha and beta features. We do not aim to keep any of the features in alpha or beta stage indefinitely. Feature gating should only be used for functionality that can eventually be turned on by default for all users (GA).
```
SgtCoDFish, line 8: question: I think the way in which some features currently behind gates are implemented might break this?

e.g. ValidateCAA is feature gated, but it's [also](https://github.com/cert-manager/cert-manager/blob/b0caedffd915eab1a52e06473dc2b5b527b84460/pkg/controller/acmechallenges/sync.go#L149) a toggle on whether or not to do the validation. We might want to graduate ValidateCAA to GA (or not, it's just an example) but it would seem reasonable for users to want to disable it.

In fact, for this specific feature I'd imagine users would want to opt-in rather than opt-out (and that might even be required so that we don't break people).

Considering this:

> Feature gating should only be used for functionality that can eventually be turned on by default for all users (GA).

For this to be true, we'd have to add a new flag "--validate-caa" which is required _as well as the feature gate_, right? (Or some other way of toggling the feature on in addition to the feature gate)

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+All beta features are off by default.
+Beta features will not be removed, but can be changed.
+If the feature gets changed in incompatible ways, we will provide migration instructions.
+A beta feature will work with all versions of Kubernetes supported by cert-manager.
+
+**GA**
+
+A feature that is GA is on by default and cannot be disabled.
+With regards to API fields and their functionality, we keep Kubernetes API compatibility promise https://cert-manager.io/docs/installation/api-compatibility/.
+The feature flag for a GA feature will eventually be removed.
```
SgtCoDFish, line 30: suggestion: I'm not sure if there'd be any reason to ever _actually_ remove one - it feels like a needless way to break people.

It doesn't seem like a huge burden to move old feature flags to `OldFeatureFlags []string`, where if any of them are passed we just log a warning and then move on.

Removing would make sense if we were ever gonna re-use a flag name but I think that could get confusing tbh

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Graduation
+
+The graduation criteria can be different for each feature.
+Generally, we find user feedback most valuable when determining if a feature is sufficiently mature to graduate. If you are using an alpha or beta feature and would like to see it graduate, it would be great if you could give us some feedback about how you use it and whether you find the API useful to initiate graduation process. Feel free to [open a GitHub issue](https://github.com/cert-manager/cert-manager/issues/new/choose) or [join one of our meetings](https://cert-manager.io/docs/contributing/#meetings) to talk about this.
```
SgtCoDFish, line 35: nitpick: the meetings link is to the website, so it might be worth following the [linking to other pages](https://github.com/cert-manager/website/#task-linking-to-other-documentation-pages) guide.

content/docs/installation/features.md
```diff
@@ -0,0 +1,62 @@
...
+
+## List of current feature gates
+
+### Alpha
+
+See `--feature-gates` flags on cert-manager controller and webhook to enable any of these features.
+
+- `AdditionalCertificateOutputFormats`. Added in cert-manager 1.7.0. Allows to specify additional formats in which cert-manager will store issued certificates and keys. See [release note](https://cert-manager.io/docs/release-notes/release-notes-1.7#additional-certificate-output-formats). Requires the feature to be enabled on both cert-manager controller and webhook
```
SgtCoDFish, line 43: nitpick: another website link here; there are a few more in this section.

SgtCoDFish (commented):
I really like the overall gist of this. Got a few comments on it overall, but I like the core of this! Thank you for your work in pushing this forwards, it's super valuable!

content/docs/installation/features.md
jsoref: `experimental`, `evolving`, `under development`

content/docs/installation/features.md
irbekrm: > For this to be true, we'd have to add a new flag "--validate-caa" which is required as well as the feature gate, right? (Or some other way of toggling the feature on in addition to the feature gate)

Yes, I think that the mechanism for marking a piece of functionality as alpha/beta should be separate from ways how something that should always be opt-in gets toggled on/off. I think that's better than mixing the two up because it simplifies the feature flag mechanism that is already complicated, is how it's implemented in kube and would appear more consistent to users once it's in GA.

It has the downside that when some functionality X that should always be opt-in is in alpha, using it might look like
```
go run <cert-manager-controller> \
 --features-gates=<optional-functionality-x>=true \
 --<enable-optional-functionality-x>
```
but I think this inconvenience is acceptable so that in GA we can have

```
go run <cert-manager-controller> 
 --<enable-optional-functionality-a>
 --<enable-optional-functionality-b>
 --<enable-optional-functionality-x>
```
and not
```
go run <cert-manager-controller> 
 --<enable-optional-functionality-a>
 --<enable-optional-functionality-b>
 --feature-gates=<enable-optional-functionality-x> // X is GA same as A and B, but gets toggled on via --feature-gates flag because it once was an alpha feature
```

I think it's ok if enabling an alpha feature is a bit clunky, perhaps it'd even encourage folks to nudge us to move it through to graduation.

Let me know what you think.

content/docs/installation/features.md
irbekrm: Yeah I agree with you, I think removing would make sense if we ended up with a huge number of features gates, so that the help output would become unreadable, but this is currently not the case.

irbekrm: The downside would be that if it's a feature gated opt-in functionality that gets enabled via flag in GA, it could be confusing that it appears in two places. But since this would be non-functional, we could change it if folks complained.

content/docs/installation/features.md
irbekrm: Thank for the reminder, that guide is great btw 👍🏼 I have changed all internal links.

content/docs/contributing/featuregates.md
irbekrm: You would get errors in controller logs if there are resources with fields that are no longer valid. I've added a note to warn folks that resources need updating before disabling a previously enabled feature gate flags.

content/docs/installation/features.md
irbekrm: > But since the features are feature gated

😆 

yeah, I think in some cases it might not be possible to know whether something will need to opt-in. For example with Gateway, we originally assumed that those CRDs will at some point become part of core Kubernetes.

irbekrm:
Thanks for the review folks! I've now addressed all the comments, PTAL

https://deploy-preview-992--cert-manager-website.netlify.app/docs/installation/featureflags/
https://deploy-preview-992--cert-manager-website.netlify.app/docs/contributing/featuregates/

content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,47 @@
...
+- Default cert-manger e2e CI tests run with all feature gates for all components enabled. There is an additional optional e2e test that runs with all feature gates disabled. You can trigger that for your PR with `/test pull-cert-manager-e2e-feature-gates-disabled` to verify that all works as expected both with and without the new feature gate.
+
+### Potential issues
+
+- The person deploying cert-manager has to remember to set two cert-manager feature gates, one of the webhook one on the controller for the feature to function. Forgetting to set one of them might result in unexpected behavior
+
+- A user must remember to remove the alpha fields from their manifests when disabling a previously enabled API feature. Failing to do so might result in unexpected behavior- for example forgetting to remove feature gated field from a `Certificate` resource might result in failed renewals at some later point when cert-manager's controller will attempt to update the `Certificate` spec, but the webhook will reject the update due to the feature gated field being set
```
jsoref, line 35: Add `.` at end

content/docs/installation/featureflags.md
```diff
@@ -0,0 +1,72 @@
...
+
+- `ServerSideApply`. Added in cert-manager 1.8.0. If this feature is enabled, cert-manager uses [Server side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) when creating or updating API resources. This will speed cert-manager operations and prevent the resource version conflict errors. See [release notes](../release-notes/release-notes-1.8.md#server-side-apply)
+
+- `StableCertificateRequestName`. Added in cert-manager 1.10.0. Will enable generation of `CertificateRequest` resources with a fixed name. See [`cert-manager#5487`](https://github.com/cert-manager/cert-manager/pull/5487)
+
+- `UseCertificateRequestBasicConstraints`. Added in cert-manager 1.11.0. Makes cert-manager add a basic constraints section to certificate signing requests with the CA constraint set to the correct value. See [`cert-manager#5552`](https://github.com/cert-manager/cert-manager/pull/5552)
+
+- `ValidateCAA`. Added in cert-manager 0.7.2. CAA checking when issuing a certificate.
```
jsoref, line 67: Can you add a see link for this one?

irbekrm: There's no description/documentation for this as far as I am aware, it was implemented pre-v1 where we didn't have more descriptive release notes yet

content/docs/installation/featureflags.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Feature flags
+description: Using feature gated functionality
+---
+
+New cert-manager features and functionality is often initially implemented behind a feature gate. This is so as to not break users with functionality that has not yet been tested in production as well as to give us a chance to remove or modify API fields and functionality following user feedback.
```
SgtCoDFish, line 6: nitpick: grammar
```suggestion
New cert-manager features and functionality are often initially implemented behind a feature gate. This is so as to not break users with functionality that has not yet been tested in production as well as to give us a chance to remove or modify API fields and functionality following user feedback.
```

content/docs/installation/featureflags.md
```diff
@@ -0,0 +1,72 @@
...
+**Alpha**
+
+All alpha features are off by default. We retain the right to change or remove
+alpha features without warning. An API field that is part of an alpha feature
+and requires a webhook feature flag to be used also can also be removed from the
+API without warning. An alpha feature might not work for all cert-manager's
+supported Kubernetes versions. If you want to disable a previously enabled alpha
+feature gate, you should make sure that you have updated any resources that have API
+fields set that are only valid if the feature gate is enabled, else the resources
+will end up in invalid state.
```
SgtCoDFish, line 21: nitpick: missing "an"

```suggestion
fields set that are only valid if the feature gate is enabled, or else the resources
will end up in an invalid state.
```

content/docs/manifest.json
```diff
@@ -472,6 +476,10 @@
               "title": "Developing with Kind",
               "path": "/docs/contributing/kind.md"
             },
+            {
+              "title": "Implementing feature gates",
```
SgtCoDFish, line 480: nitpick: The rest of the titles in this section use title case so this one stands out a little

```suggestion
              "title": "Implementing Feature Gates",
```

content/docs/contributing/featuregates.md
```diff
@@ -0,0 +1,47 @@
...
+
+- cert-manager's [API compatibility promise](../installation/api-compatibility.md)
+
+- An example implementation of an alpha field is [`AdditionalOutputFormats` field on `Certificate` spec](https://github.com/cert-manager/cert-manager/blob/dbad3d98f3d7d85cadb4bd2c2493faf8b666b313/internal/apis/certmanager/types_certificate.go#L169-L174)
+
+- [Kubernetes definition of feature stages](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages)
+
+- Kubernetes [API change design](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md)
+
+[^1]: for example, functionality that might change in the future in response to user feedback
```
SgtCoDFish, line 47: serious nitpick (non-blocking): Does this look better if capitalized? Really, really doesn't matter though!

```suggestion
[^1]: For example, functionality that might change in the future in response to user feedback
```

SgtCoDFish (commented):
Spotted a few typos and bits but I'm happy to merge this now!

SgtCoDFish (approved):
/lgtm
/approve

Thanks for this! I think it's good to go - it can always evolve more later if we end up needing that!

I guess this does introduce a need to rework some of the feature gated stuff but I don't really see a huge rush!

PR #1003, opened 2022-06-09
Update docs on alt cert chains

SgtCoDFish:
This was _incredibly_ outdated, but the general idea of it might have some use for some ACME servers in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,12 @@ is configured in the frontmatter for each docs page. The magic happens in in `pa
 
 For pages _except_ docs and for some other tags, look at changing [`next-seo.config.js`](./next-seo.config.js).
 
+### Task: Adding a Comment in Documentation
+
+Sometimes you'll want to add a comment which is only for documentation maintainers.
+
+Use `{/* my comment */}` rather than the HTML-style comments you'd normally use for Markdown files. Other comment types will cause errors.
+
 ## Website Development Tooling
 
 ### Development Server
--- a/content/docs/configuration/acme/README.md
+++ b/content/docs/configuration/acme/README.md
@@ -341,10 +341,10 @@ solvers:
       - 'test.example.com'
       - 'example.dev'
 ```
-In this case the `DNS01` solver for CloudFlare will only be used to solve a
+
+In this case the `DNS01` solver for Cloudflare will only be used to solve a
 challenge for a DNS name if the `Certificate` has a label from
 `matchLabels` _and_ the DNS name matches a zone from `dnsZones`.
-
 
 ## Alternative Certificate Chains
 
--- a/content/docs/configuration/acme/README.md
+++ b/content/docs/configuration/acme/README.md
@@ -73,46 +73,6 @@ Solvers come in the form of [`dns01`](./dns01/README.md) and
 these solver types, visit their respective documentation -
 [DNS01](./dns01/README.md), [HTTP01](./http01/README.md).
 
-### Use an Alternative Certificate Chain
-
-On January 11th 2021, Let's Encrypt will [change over](https://community.letsencrypt.org/t/transition-to-isrgs-root-delayed-until-jan-11-2021/125516) to using its own `ISRG Root` CA.
-This will replace the cross-signed certificates by `Identrust`. This change over needs no changes to your cert-manager configuration, any renewed or new certificates issued after this date will use the new CA root.
-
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -88,6 +88,12 @@ is configured in the frontmatter for each docs page. The magic happens in in `pa
 
 For pages _except_ docs and for some other tags, look at changing [`next-seo.config.js`](./next-seo.config.js).
 
+### Task: Adding a Comment in Documentation
+
+Sometimes you'll want to add a comment which is only for documentation maintainers; in regular markdown you'd use HTML style comments: `<!-- my comment -->`
```
jsoref, line 93: Can I encourage you to write this of the form "while you might be inclined to...". Starting with a note about how to do something only to write in a later paragraph "don't do the thing we mentioned in the previous paragraph" is counterproductive at best and really closer to setting someone up to fail.

content/docs/configuration/acme/README.md
```diff
@@ -383,4 +343,28 @@ solvers:
...
+
+## Alternative Certificate Chains
+
+{/* This empty link preserves old links to #alternative-certificate-chain", which matched the old title of this section */}
+
+<a id="alternative-certificate-chain" className="hidden-link"></a>
+
+It's possible to choose alternative certificate chains when fetching a certificate from an ACME server. This allows issuers to gracefully roll people over to a new root.
+
+This functionality is not Let's Encrypt exclusive; if your ACME server supports signing by multiple CAs you can use `preferredChain` with the value of the Common Name of the chain you want in the Issuer part of the certificate.
```
jsoref, line 357:
```suggestion
This functionality is not exclusive to Let's Encrypt; if your ACME server supports signing by multiple CAs you can use `preferredChain` with the value of the Common Name of the chain you want in the Issuer part of the certificate.
```

content/docs/configuration/acme/README.md
```diff
@@ -383,4 +343,28 @@ solvers:
...
+
+By way of an example, below is how a user would have requested an alternative chain before the [now completed](https://community.letsencrypt.org/t/transition-to-isrgs-root-delayed-until-jan-11-2021/125516) "ISRG Root" changeover. Since this change has already happened, there's no need for this with Let's Encrypt any more:
```
jsoref, line 359:
```suggestion
By way of an example, below is how a user would have requested an alternative chain before the (now completed) ["ISRG Root" changeover](https://community.letsencrypt.org/t/transition-to-isrgs-root-delayed-until-jan-11-2021/125516). Since this change has already happened, there's no need for this with Let's Encrypt anymore:
```

README.md
SgtCoDFish: Really good suggestion yeah, thanks, will reword this :+1:

content/docs/configuration/acme/README.md
SgtCoDFish: I prefer the link that way yeah, thanks :+1:

README.md
```diff
@@ -88,6 +88,12 @@ is configured in the frontmatter for each docs page. The magic happens in in `pa
 
 For pages _except_ docs and for some other tags, look at changing [`next-seo.config.js`](./next-seo.config.js).
 
+### Task: Adding a Comment in Documentation
```
irbekrm, line 91: Thank you for continuously adding these awesome tips!

content/docs/configuration/acme/README.md
```diff
@@ -383,4 +343,28 @@ solvers:
 ```
 In this case the `DNS01` solver for CloudFlare will only be used to solve a
 challenge for a DNS name if the `Certificate` has a label from
-`matchLabels` _and_ the DNS name matches a zone from `dnsZones`.
\ No newline at end of file
+`matchLabels` _and_ the DNS name matches a zone from `dnsZones`.
+
+
+## Alternative Certificate Chains
```
irbekrm, line 349: Suggestion: it would be good to keep the bit of information that if the 'preferredChain' is not found then the default certificate will be returned as I imagine that this is not obvious.

irbekrm (changes requested):
Thank you @SgtCoDFish! I think this is a very useful change as the previous content was obviously out of date. I've added a small nit to keep the info about the defaulting except for that happy to lgtm

content/docs/configuration/acme/README.md
SgtCoDFish: Yeah good point! I've updated now in a new commit (with some better language overall I think)

PR #1004, opened 2022-06-10
Update failed issuance FAQ

irbekrm:
The failed issuance FAQ was out of date, it would be good to be able to link to it in case of questions like https://kubernetes.slack.com/archives/C4NV3DWUC/p1654858541061359


content/docs/faq/README.md
```diff
@@ -54,11 +54,15 @@ cert-manager publishes all events to the Kubernetes events mechanism, you can ge
...
 Due to the nature of the Kubernetes event mechanism these will be purged after a while. If you're using a dedicated logging system it might be able or is already also storing Kubernetes events.
 
-### What happens if a renewal doesn't happen? Will it be tried again after some time?
+### What happens if an issuance failed? Will it be retried?
 
-cert-manager will retry renewal if it encounters temporary failures. It uses an exponential backoff algorithm to calculate the delay between each retry.
+cert-manager will retry a failed issuance except for a few rare edge cases where manual intervention is needed.
 
-A temporary failure is one that doesn't mark the `CertificateRequest` as failed. If the `CertificateRequest` is marked as failed, issuance will be re-tried in 1 hour.
+If an issuance failed because of a temporary error, it will be retried again with a short exponential backoff (currently 5s to 5min). A temporary error is one that does not result in a failed `CertificateRequest`.
```
irbekrm, line 61: 5s to 5min is the exponential backoff for reconciles that return with an error https://github.com/cert-manager/cert-manager/blob/master/pkg/controller/util.go#L45

content/docs/faq/README.md
```diff
@@ -54,11 +54,15 @@ cert-manager publishes all events to the Kubernetes events mechanism, you can ge
 
 Due to the nature of the Kubernetes event mechanism these will be purged after a while. If you're using a dedicated logging system it might be able or is already also storing Kubernetes events.
 
-### What happens if a renewal doesn't happen? Will it be tried again after some time?
+### What happens if an issuance failed? Will it be retried?
```
SgtCoDFish, line 57: nitpick: I'd prefer "issuance fails"  rather than "an issuance failed" but I don't think it _needs_ to change!

nitpick: changing this title changes the anchor link too; do you think we should add a manual `<a>` so we preserve the old link, as I did in [this other PR](https://github.com/cert-manager/website/pull/1003/files#diff-87f6019faccfa562a228b696d2b5a8ff9c71392f66eb62db144f3fdf84c6b905R351-R353)? I don't feel strongly though! (the suggestion would require that https://github.com/cert-manager/website/pull/1003 be merged first since it uses the new `className`

```suggestion
### What happens if issuance fails? Will it be retried?

<a id="what-happens-if-a-renewal-doesnt-happen-will-it-be-tried-again-after-some-time" className="hidden-link"></a>
```

content/docs/faq/README.md
```diff
@@ -54,11 +54,15 @@ cert-manager publishes all events to the Kubernetes events mechanism, you can ge
...
 
-cert-manager will retry renewal if it encounters temporary failures. It uses an exponential backoff algorithm to calculate the delay between each retry.
+cert-manager will retry a failed issuance except for a few rare edge cases where manual intervention is needed.
 
-A temporary failure is one that doesn't mark the `CertificateRequest` as failed. If the `CertificateRequest` is marked as failed, issuance will be re-tried in 1 hour.
+If an issuance failed because of a temporary error, it will be retried again with a short exponential backoff (currently 5 seconds to 5 minutes). A temporary error is one that does not result in a failed `CertificateRequest`.
+
+If the issuance has failed with an error that resulted in a failed `CertificateRequest`, it will be retried with a longer binary exponential backoff (1 hour to 32 hours) to avoid overwhelming external services.
+
+You can always trigger immediate renewal using [`cmctl renew` command](../usage/cmctl.md#renew)
```
SgtCoDFish, line 65: nitpicks: don't think any of these are needed but I thought I'd suggest them anyway:

```suggestion
If issuance fails because of a temporary error, it will be retried again with a short exponential backoff (currently 5 seconds to 5 minutes). A temporary error is one that does not result in a failed `CertificateRequest`.

If issuance fails with an error that resulted in a failed `CertificateRequest`, it will be retried with a longer binary exponential backoff (1 hour to 32 hours) to avoid overwhelming external services.

You can always trigger immediate renewal using the [`cmctl renew` command](../usage/cmctl.md#renew)
```

SgtCoDFish (approved):
/lgtm
/hold

Added a few suggestions but would happily merge!

irbekrm:
Thanks for the good suggestions @SgtCoDFish I somehow missed this code review!

I've made the suggested changes now

SgtCoDFish (approved):
/lgtm
/approve

Looks good, cheers! :+1:

PR #1007, opened 2022-06-14
Document new and old release processes

SgtCoDFish:
[During the release of v1.8.1](https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1655221978315829) it was discovered that the path to which we write staged artifacts was unintentionally changed by using `make publish-release` rather than our old `cmrel` magic which constructed a longer path.

This change actually simplifies our release process and will simplify `cmrel makestage` too. This PR documents that, and also documents how _older_ versions of cert-manager (1.7 and before) should be released, in case we find a reason to do that!

This also removes the step where we're instructed to do something with algolia, because we don't actually know how to do that.

Each commit has more details in its full commit message.

### DIFF
--- a/content/docs/contributing/release-process.md
+++ b/content/docs/contributing/release-process.md
@@ -311,7 +311,7 @@ page if a step is missing or if it is outdated.
         cmrel makestage --ref=v1.8.0-beta.0
         ```
 
-        This step takes ~5 minutes. It will build all Docker images and create
+        This step takes ~5 minutes. It will build all container images and create
         all the manifest files, sign Helm charts and upload everything to a storage
         bucket on Google Cloud. These artifacts will then be published and released
         in the next steps.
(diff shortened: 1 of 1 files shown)


SgtCoDFish:
Linked to https://github.com/cert-manager/release/pull/78

content/docs/contributing/release-process.md
```diff
@@ -314,7 +317,7 @@ page if a step is missing or if it is outdated.
         in the next steps.
 
         <div className="pageinfo pageinfo-info"><p>
-        🔰 Remember to keep open the terminal where you run <code>cmrel stage</code>. Its output will be used in the next step.
+        🔰 Remember to keep open the terminal where you run <code>cmrel makestage</code>. Its output will be used in the next step.
```
irbekrm, line 320: Suggestion: remove as this actually is no longer needed (the output was the printed longer path of the staged artifacts)

SgtCoDFish: Haha I saw that I'd not updated the `stage` bit before and it blinded me to the fact that this wasn't needed any more! Good catch, will remove.

irbekrm:
Thanks @SgtCoDFish, nice to see that the release notes get clearer!

/lgtm
/approve

PR #1009, opened 2022-06-16
Add other backports section

SgtCoDFish:
This attempts to formalise that PRs such as [cert-manager#5209](https://github.com/cert-manager/cert-manager/pull/5209)  are OK, and e.g. if we wanted to backport improvements to the descriptions on our CRDs that would be OK too.


jahrlin (commented):
What about Helm charts and kube manifests? Should we allow backports for tooling changes that also produces new version of charts and manifests? It could potentially be a side-effect, no?

content/docs/installation/supported-releases.md
```diff
@@ -167,12 +166,30 @@ Kubernetes API or the command line flags. We avoid making breaking changes
...
+We aim to be extremely conservative in what we back-port. That applies especially for anything
+which could be a _runtime_ change - that is, a change which might alter behavior for someone
+upgrading between patch releases.
+
+That means that if a candidate for back-porting has a chance of having a runtime impact we're
+unlikely to accept the change unless it addresses a security issue or a critical bug.
+
+We reserve the right to back-port other changes which are unlikely to have a runtime impact, such as
+documentation or tooling changes. An example would be [#5209][] which updated how we perform a release of
+cert-manager but didn't have any realistic chance of having a runtime impact.
```
munnerz, line 180: I think it's great to write down/clarify all this 😄

Personally I am more conservative wrt making changes to the tooling surrounding making a release, as I think these _can_ affect runtime of cert-manager (e.g. if we make a change to release tooling that means say, the `acmesolver` image is not updated, then this could quietly break an end-users renewal).

I think there _is_ definitely room for some discretion here (as noted in #5209), but in general I think we should try and avoid any changes as much as possible (i.e. develop tooling changes/features & deprecation cycles in such a way that enables us to not have to modify old release branches too)

content/docs/installation/supported-releases.md
```diff
@@ -167,12 +166,30 @@ Kubernetes API or the command line flags. We avoid making breaking changes
 where possible, and where they're required we'll give as much notice as
 possible.
 
+<h4 id="other-backports">Other back-ports</h4>
+
+We aim to be extremely conservative in what we back-port. That applies especially for anything
```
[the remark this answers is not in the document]
SgtCoDFish: Thanks, agreed! :grin:

content/docs/installation/supported-releases.md
SgtCoDFish: I proposed the following rule of thumb which I think is a good mix of pragmatism and conservatism:

> "Does this back-port improve cert-manager, bearing in mind that we really value stability for already-released versions?"

What do you propose instead?

munnerz: Summarising it in one sentence, I'd say "The stability of already-released versions & their following patch versions is valued above all else. Only the most essential patches needed to ensure the user's security & reliability should be merged." (i.e. security fixes & regressions is an example).

> "Does this back-port improve cert-manager, bearing in mind that we really value stability for already-released versions?"

Regardless on what definition that we go for here, I actually think this statement right now _doesn't_ cover the rclone changes mentioned previously in the paragraph, as from an end-user perspective it has not been improved & it doesn't improve/increase the stability of the release either. (so I definitely think _something_ needs to change here :))

munnerz: To make absolutely clear here - I am not calling into question whether that rclone PR should have been backported. I am simply highlighting that we give that PR as a reference in this text, and then the rule of thumb at the bottom introduces ambiguity as to whether or not it is covered by it, as it seems definition of an "improvement" differs/it's quite a broad term.

SgtCoDFish:
> What about Helm charts and kube manifests? Should we allow backports for tooling changes that also produces new version of charts and manifests? It could potentially be a side-effect, no?

I think with anything like this it's gotta be a case-by-case approach - if we had to backport a `helm` update or change a chart to fix a security issue that'd be totally reasonable, but I guess generally we'd probably avoid it in most cases.

If it'd come under "changes which are unlikely to have a runtime impact" it might be worth considering but it'd have to be case-by-case

jahrlin:
Agreed. Thanks for this excellent addition!

/lgtm

PR #1011, opened 2022-06-16
Certificate re-issuance FAQ

irbekrm:
This PR:

- adds an FAQ for when will certs be re-issued
- adds a warning to the section in backup & restore docs about potential side effects of there being no `CertficateRequest`s

This was triggered by user question https://github.com/cert-manager/cert-manager/issues/5146

### DIFF
--- a/content/docs/faq/README.md
+++ b/content/docs/faq/README.md
@@ -22,20 +22,23 @@ This is a feature in cert-manager starting in `v0.16` using the `cmctl` CLI. Mor
 
 To determine if a certificate needs to be re-issued, cert-manager looks at the the spec of `Certificate` resource and latest `CertificateRequest`s as well as the data in `Secret` containing the X.509 certificate.
 
-Issuance process will always get triggered if:
-- the `Secret` named on `Certificate`'s spec, does not exist, is missing private key or certificate data or contains corrupt data
-- private key stored in the `Secret` does not match private key spec on `Certificate`
-- public key of the issued certificate does not match the private key stored in the `Secret`
-- cert-manager issuer annotations on the `Secret` do not match the issuer specified on `Certificate`
-- DNS names, IP addresses, URLS or email addresses on the issued certificate do not match those on the `Certificate` spec
-- certificate needs to be renewed
+The issuance process will always get triggered if the:
 
-Additionally if the latest `CertificateRequest` for the `Certificate` is found, cert-manager will also re-issue if:
+- `Secret` named on `Certificate`'s spec, does not exist, is missing private key or certificate data or contains corrupt data
+- private key stored in the `Secret` does not match the private key spec on `Certificate`
+- public key of the issued certificate does not match the private key stored in the `Secret`
+- cert-manager issuer annotations on the `Secret` do not match the issuer specified on the `Certificate`
+- DNS names, IP addresses, URLS or email addresses on the issued certificate do not match those on the `Certificate` spec
+- certificate needs to be renewed (because it has expired or the renewal time is now or in the past)
+- certificate has been marked for renewal manually [using `cmctl`](../usage/cmctl.md#renew)
+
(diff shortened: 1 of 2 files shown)


content/docs/faq/README.md
```diff
@@ -18,6 +18,37 @@ face:
...
+
+To determine if a certificate needs to be re-issued, cert-manager looks at the the spec of `Certificate` resource and latest `CertificateRequest`s as well as the data in `Secret` containing the X.509 certificate.
+
+Issuance process will always get triggered if:
+- the `Secret` named on `Certificate`'s spec, does not exist, is missing private key or certificate data or contains corrupt data
+- private key stored in the `Secret` does not match private key spec on `Certificate`
+- public key of the issued certificate does not match the private key stored in the `Secret`
+- cert-manager issuer annotations on the `Secret` do not match the issuer specified on `Certificate`
+- DNS names, IP addresses, URLS or email addresses on the issued certificate do not match those on the `Certificate` spec
+- certificate needs to be renewed
```
SgtCoDFish, line 31: nitpick (non-blocking): I think with a slight reword this would read a little more naturally, but it's totally fine as is!

suggestion (non-blocking): Also suggesting a little extra! We mention `cmctl renew` below but I wonder if it might be worth also mentioning it here since I feel like people are more likely to read bullet points than prose!

```suggestion
The issuance process will always get triggered if the:

- `Secret` named on `Certificate`'s spec, does not exist, is missing private key or certificate data or contains corrupt data
- private key stored in the `Secret` does not match the private key spec on `Certificate`
- public key of the issued certificate does not match the private key stored in the `Secret`
- cert-manager issuer annotations on the `Secret` do not match the issuer specified on the `Certificate`
- DNS names, IP addresses, URLS or email addresses on the issued certificate do not match those on the `Certificate` spec
- certificate needs to be renewed (because it's going to expire)
- certificate has been marked for renewal manually [using `cmctl`](../usage/cmctl.md#renew)
```

content/docs/faq/README.md
```diff
@@ -18,6 +18,37 @@ face:
...
+
+Additionally if the latest `CertificateRequest` for the `Certificate` is found, cert-manager will also re-issue if:
+- the common name on the CSR found on the `CertificateRequest` does not match that on the `Certificate` spec
+- the subject fields on the CSR found on the `CertificateRequest` do not match the subject fields of the `Certificate` spec
+- the duration on the `CertificateRequest` does not match the duration on the `Certificate` spec
+- `isCA` field value on the `Certificate` spec does not match that on the `CertificateRequest`
+- DNS names, IP addresses, URLS or email addresses on the `CertificateRequest` spec do not match those on the `Certificate` spec
+- key usages on the `CertificateRequest` spec do not match those on the `Certificate` spec
```
SgtCoDFish, line 39: nitpick (non-blocking): as above, minor rewording to make it read naturally:

```suggestion
Additionally, if the latest `CertificateRequest` for the `Certificate` is found, cert-manager will also re-issue if:

- the common name on the CSR found on the `CertificateRequest` does not match that on the `Certificate` spec
- the subject fields on the CSR found on the `CertificateRequest` do not match the subject fields of the `Certificate` spec
- the duration on the `CertificateRequest` does not match the duration on the `Certificate` spec
- `isCA` field value on the `Certificate` spec does not match that on the `CertificateRequest`
- the DNS names, IP addresses, URLS or email addresses on the `CertificateRequest` spec do not match those on the `Certificate` spec
- key usages on the `CertificateRequest` spec do not match those on the `Certificate` spec
```

SgtCoDFish (approved):
/lgtm
/hold
/approve

I have a few suggestions but they're minor; this is a big improvement and I'd happily merge as-is!

irbekrm:
Thanks @SgtCoDFish for the good suggestions!
I've commited all except for changing `because it's going to expire` -> `because it has expired or the renewal time is now or in the past`

SgtCoDFish (approved):
/lgtm
/unhold

Thanks, looks great :grin:

PR #1019, opened 2022-06-21
Update next docs

irbekrm:
This PR updates next-docs as discussed earlier in our standup so folks who want to contribute documentation for features that will only be included in the next release have an up-to-date place where to PR that against (namely ./content/next-docs)

I've copied the Readme and every subdirectory of `./content/docs/` except for `reference` which I think makes no sense to copy as we'll generate it at the time of release.

I havent' actually read through these changes (but cannot see why as at this point we aren't evaluating content)


SgtCoDFish (approved):
/lgtm
/approve

I'm assuming this is all kosher! We're surely gonna merge at least a few PRs which only add to docs and not next docs between now and the release of 1.9, but that's not the end of the world; should be easy to diff docs and next-docs and see if there's been divergence!

maelvls:
I noticed that there is a clash in the paths that are declared in [`content/next-docs/manifest.json`](https://github.com/cert-manager/website/blob/b441b2d3b82e14d57b22ad041fffea7fe5ffd7ae/content/next-docs/manifest.json) and in [`content/docs/manifest.json`](https://github.com/cert-manager/website/blob/b441b2d3b82e14d57b22ad041fffea7fe5ffd7ae/content/docs/manifest.json). Both manifests define paths starting with `/docs`.

A symptom of this came up when Richard removed `/docs/faq/webhook` in #1064, the file `sitemap-0.xml` which still showed `/docs/faq/webhook` even though the page disappeared from content/docs/faq/webhook.

SgtCoDFish:
> I noticed that there is a clash in the paths that are declared in [content/next-docs/manifest.json](https://github.com/cert-manager/website/blob/b441b2d3b82e14d57b22ad041fffea7fe5ffd7ae/content/next-docs/manifest.json) and in [content/docs/manifest.json](https://github.com/cert-manager/website/blob/b441b2d3b82e14d57b22ad041fffea7fe5ffd7ae/content/docs/manifest.json). Both manifests define paths starting with /docs.

Yeah you're right, the next-docs manifest should say "next-docs" and not "docs"

PR #1022, opened 2022-06-28
Modifications after trying the istio-csr tutorial

davidsbond:
This commit makes two small tweaks to the istio-csr tutorial on the website.

The first is that the final clean up command will fail, following the instructions verbatim
will produce an error, so it has been updated to just `kind delete cluster`, which successfully
deletes the kind cluster. 

```
$ kind delete cluster kind
ERROR: unknown command "kind" for "kind delete cluster"
```

The second change is to add that the `isitoctl install` command will prompt for user input
to confirm the choice of profile. The user should confirm the choice. 

Signed-off-by: David Bond <davidsbond@users.noreply.github.com>

PR #1023, opened 2022-06-28
Updates developer documentation to give examples of using make

SgtCoDFish:
This replaces a _lot_ of older documentation which referenced bazel, most of which will no longer work. A lot of the docs were incorrect even before we started making the switch to make.

Sorry that this is a large PR; I ended up changing more than I expected because so much of the language was out of date, broken or just wrong. Our DCO signoff page for example still asked that copyright notices mention Jetstack, even though the notices were updated a long time ago!

**Important**: This PR is smaller than it looks! Since all the changes are duplicated in both docs and next-docs, the lines changed looks twice as big as it actually is!

### DIFF
--- a/content/docs/contributing/building.md
+++ b/content/docs/contributing/building.md
@@ -20,9 +20,11 @@ You should install the following tools before you start developing cert-manager:
 
 - [git](https://git-scm.com/)
 - [curl](https://curl.se/)
-- GNU Coreutils (available via [homebrew](https://formulae.brew.sh/formula/coreutils) for macOS)
-- `jq` (available in package managers and in homebrew)
+- [GNU make](https://www.gnu.org/software/make/manual/make.html), `v3.82` or newer
+- GNU Coreutils (usually already installed on Linux, available via [homebrew](https://formulae.brew.sh/formula/coreutils) for macOS)
+- `jq` (available in Linux package managers and in [homebrew](https://formulae.brew.sh/formula/jq))
 - `docker` (or `podman`, see [Container Engines](#container-engines) below)
+- `Go` (optional; see [Go Versions](#go-versions) below)
 
 ## Getting Started
 
@@ -31,15 +33,39 @@ developing cert-manager. We'll also provide an overview on this page of some of 
 
 ### Go Versions
 
-cert-manager can be built with the version of Go you've installed locally on your system. Alternatively, you can "vendor" Go to ensure you use the
-same version that's built with in CI.
+cert-manager defaults to using whatever version of Go you've installed locally on your system. If you want to use your system Go, that's totally fine.
 
-To check which version of Go is _currently_ being used, run: `make which-go`.
+Alternatively, make can provision and "vendor" Go specifically for cert-manager, helping to ensure you use the same version that's used in CI and to
+make it easier to get started developing.
 
-To start using a vendored Go, run: `make vendor-go`. You only need to run `vendor-go` once and it'll be "sticky", being used for all future make invocations
-in your local checkout.
+To start using a vendored Go, run: `make vendor-go`.
+
+You only need to run `vendor-go` once and it'll be "sticky", being used for all future make invocations in your local checkout.
 
 To return to using your system version of go, run: `make unvendor-go`.
+
+To check which version of Go is _currently_ being used, run: `make which-go`, which prints the version number of Go and the path to the Go binary.
+
+```console
+# Use a vendored version of go
+$ make vendor-go
+cd _bin/tools/ && ln -f -s ../downloaded/tools/_go-1.XY.Z-linux-amd64/goroot .
+cd _bin/tools/ && ln -f -s ../downloaded/tools/_go-1.XY.Z-linux-amd64/goroot/bin/go .
+
+# A path to go inside the cert-manager directory indicates that a vendored Go is being used
+$ make which-go
+go version go1.XY.Z linux/amd64
+go binary used for above version information: /home/user/workspace/cert-manager/_bin/tools/go
+
+# Go back to the system Go
+$ make unvendor-go
+rm -rf _bin/tools/go _bin/tools/goroot
+
+# The binary is now "go" which should be found in $PATH
+$ make which-go
+go version go1.AB.C linux/amd64
+go binary used for above version information: go
+```
 
 ### Parallelism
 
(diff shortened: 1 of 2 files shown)


content/next-docs/tutorials/istio-csr/istio-csr.md
```diff
@@ -136,6 +136,14 @@ istioctl install -f istio-install-config.yaml
 # istioctl install --set profile=openshift -f istio-install-config.yaml
 ```
 
+You will be prompted for input to confirm your choice of Istio profile:
```
SgtCoDFish, line 139: note: This is just #1022 but copied into next-docs - I didn't spot it when I reviewed that PR, so I'm fixing it here :grin:

content/docs/contributing/building.md
```diff
@@ -3,123 +3,211 @@ title: Building cert-manager
 description: 'cert-manager contributing guide: Building cert-manager'
 ---
 
-cert-manager makes use of [Bazel](https://bazel.build/) to build the project. 
-Bazel manages all developer dependencies, Helm chart building, Docker images and the code itself. 
-We try to use it as much as possible.
-We currently use Bazel `v4.2.1`. The minimum supported version is `v4.0.0`.
+cert-manager is built and tested using [make](https://www.gnu.org/software/make/manual/make.html), with a focus on using the standard Go tooling
```
irbekrm, line 6: Question: does it matter what version of make we use (i.e in CI)? Is it worth mentioning it here?

content/docs/contributing/building.md
```diff
@@ -3,123 +3,211 @@ title: Building cert-manager
...
+You should install the following tools before you start developing cert-manager:
 
-`bazel build [...]` will build and compile code for you e.g. `bazel build //cmd/ctl` will build our CLI. 
-`bazel test [...]` will run any tests for a given package
-`bazel run [...]` is only used to run certain scripts not the compiled code (unlike Go). e.g. `bazel run //hack/bin:helm` will download and run Helm.
+- [git](https://git-scm.com/)
+- [curl](https://curl.se/)
+- GNU Coreutils (available via [homebrew](https://formulae.brew.sh/formula/coreutils) for macOS)
+- `jq` (available in package managers and in homebrew)
+- `docker` (or `podman`, see [Container Engines](#container-engines) below)
```
irbekrm, line 25: Suggestion: if by default all targets use system's Go I'd add Go here as well- it feels more straightforward

content/docs/contributing/building.md
```diff
@@ -3,123 +3,211 @@ title: Building cert-manager
...
-> For example `bazel test //pkg/...` will test all tests in all packages.
+cert-manager can be built with the version of Go you've installed locally on your system. Alternatively, you can "vendor" Go to ensure you use the
+same version that's built with in CI.
 
-## Help so much Bazel!
+To check which version of Go is _currently_ being used, run: `make which-go`.
 
-No worries we have a lot of helper scripts for you!
-Need to set up a local cluster and install cert-manager in it? Take a look at [our kind documentation](./kind.md).
+To start using a vendored Go, run: `make vendor-go`. You only need to run `vendor-go` once and it'll be "sticky", being used for all future make invocations
```
irbekrm, line 39: This seems a little confusing tbh, how would I remember if I ran this comand already and how can I be sure that it's really 'unvendored'? But I guess this is more a question of whether we want to do it this way not a documentation issue.

Suggestion: maybe state clearly that by default system's Go is used and add all the `vendor/unvendor` as optional steps so this is more straightforward to read?

content/docs/contributing/building.md
```diff
@@ -3,123 +3,211 @@ title: Building cert-manager
...
+It's possible to limit the number of threads Go uses we'd generally recommend doing so when using Make parallelism.
+
+The best values to use will depend on your system, but we've had success using around half of the available number of cores for Make and limiting Go to between
+2 and 4 threads per core.
+
+For example, using an 8-core machine:
+
+```bash
+# Run 4 make targets in parallel, and limit each `go build` to 2 threads.
+make GOMAXPROCS=2 -j4 release-artifacts
```
irbekrm, line 61: This is useful 👍🏼

irbekrm (changes requested):
Thanks @SgtCoDFish !

I've read through the doc and generally this looks good to me- I believe that this is a part that will be changed and improved as we go, so no need to try to get this perfect first time round.

I've added a couple nits, but generally happy to lgtm

content/docs/contributing/building.md
SgtCoDFish: Good shout, yeah, I'll add :+1:

content/docs/contributing/building.md
SgtCoDFish: It _basically_ doesn't matter - I've lightly tested our build system going as far back as CentOS 7 (released in 2014) and it works fine with Make 3.82 there. I'll add 3.82 as a minimum, but realistically everyone running Linux or macOS today who's doing K8s development will have a new enough version for everything to work just fine!

content/docs/contributing/building.md
SgtCoDFish: I'll try rewording this to make it clearer!

I added `make which-go` to address this point of possible confusion specifically, and that's documented just above here; I'll add an example showing how I expected vendoring and unvendoring to work!

content/docs/contributing/building.md
```diff
@@ -31,16 +33,40 @@ developing cert-manager. We'll also provide an overview on this page of some of
...
 
-To start using a vendored Go, run: `make vendor-go`. You only need to run `vendor-go` once and it'll be "sticky", being used for all future make invocations
-in your local checkout.
+To start using a vendored Go, run: `make vendor-go`.
+
+You only need to run `vendor-go` once and it'll be "sticky", being used for all future make invocations in your local checkout.
 
 To return to using your system version of go, run: `make unvendor-go`.
 
+To check which version of Go is _currently_ being used, run: `make which-go`, which prints the version number of Go and the path to the Go binary.
```
irbekrm, line 47: This is great, thanks!

irbekrm (approved):
Thanks for adding the extra info for what go version is used @SgtCoDFish!

I think this is good to be merged now- folks will benefit from knowing/being able to use make

/lgtm

PR #1030, opened 2022-07-07
add prerequisite for non-failing testgrid to release process

jahrlin:
As per discussion in cert-manager standup on 2022-07-07

Should it say something like "for the last X days"?


jakexks (approved):
/lgtm

PR #1033, opened 2022-07-12
Replace instances of jetstack/cert-manager with cert-manager/cert-manager

SgtCoDFish:
This was done with the following command (note that it focuses on the repo location and not just `jetstack/cert-manager`):

```bash
grep -rl "https://github.com/jetstack/cert-manager" content | \
grep -v importing.md | \
xargs sed -i 's/https:\/\/github.com\/jetstack\/cert-manager/https:\/\/github.com\/cert-manager\/cert-manager/g'
```

We can't replace "jetstack/cert-manager" on its own since that would also hit helm chart install instructions and container locations. This should get through the brunt of the references to the old repo location.

We skip "importing.md" because it contains the old repo name on purpose, see here: https://github.com/cert-manager/website/blob/6af723d1f209d878ded49c13cfd0b307b8c068e2/content/docs/contributing/importing.md?plain=1#L33

Part of https://github.com/cert-manager/cert-manager/issues/4824


wallrj (approved):
/lgtm

PR #1038, opened 2022-07-18
Adds docs for projects/trust Namespace Selector feature

JoshVanL:
[Code PR](https://github.com/cert-manager/trust/pull/37)

/hold 
Until code PR is merged an in a release.

### DIFF
--- a/content/docs/projects/trust.md
+++ b/content/docs/projects/trust.md
@@ -48,13 +48,26 @@ spec:
       -----END CERTIFICATE-----
   target:
     # Data synced to the ConfigMap `my-org.com` at the key `root-certs.pem` in
-    # every namespace.
+    # every namespace that has the label "linkerd.io/inject=enabled".
     configMap:
       key: "root-certs.pem"
+    namespaceSelector:
+      matchLabels:
+        linkerd.io/inject: "enabled"
 ```
 
 Bundle currently supports the source types `configMap`, `secret` and `inLine`,
 and target type `configMap`.
+
+#### Namespace Selector
+
+The target `namespaceSelector` can be used for scoping which Namespaces targets
+are synced to, supporting the field `matchLabels`. Please see
+[here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors)
+for more information and how label selectors are configured.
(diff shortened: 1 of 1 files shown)


content/docs/projects/trust.md
```diff
@@ -48,14 +48,27 @@ spec:
...
 
+#### Namespace Selector
+
+The target `namespaceSelector` can be used for scoping which Namespaces targets
+are synced to, supporting the field `matchLabels`. Please see
+[here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors)
+for more information and how label selectors are configured.
+
+If `namespaceSelector` is empty, a bundle target will be synced to all
+Namespaces.
```
SgtCoDFish, line 70: suggestion (non-blocking): a rewording which maybe reads a little smoother and voids a link called "here"; what do you think?

```suggestion
The `namespaceSelector` field in a target reduces the set of Kubernetes namespaces to which that target is synced.

Label selectors are detailed in the [Kubernetes documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors).

If `namespaceSelector` is empty, a bundle target will be synced to all namespaces.
```

content/docs/projects/trust.md
```diff
@@ -48,14 +48,27 @@ spec:
       -----END CERTIFICATE-----
   target:
     # Data synced to the ConfigMap `my-org.com` at the key `root-certs.pem` in
-    # every namespace.
+    # every namespace that has the label "linkerd.io/inject=enabled".
```
SgtCoDFish, line 51: nitpick: it might be worth adding this to `content/next-docs/projects/trust.md` as well, but that only matters if this PR goes in before 1.9 and even then it'll be pretty simple for the releaser to fix!

SgtCoDFish (approved):
/lgtm
/approve
/hold

Looks good, thanks for being so diligent with documenting this!

Holding so we can choose to merge alongside the code PR and because I made a totally optional suggestion.

SgtCoDFish:
/unhold

The feature itself has merged now; no real reason to delay this going in!

PR #1039, opened 2022-07-18
Link to the Season-of-Docs documentation survey

wallrj:
Mehak,

I've update the survey banner with your documentation survey.

PTAL

/cc @maelvls @mehak151


SgtCoDFish:
> Is Richard Collins done with his user survey by the way?

I think we can assume it's done now, yeah!

PR #1047, opened 2022-07-26
Updates after the release of v1.9.1

SgtCoDFish:
Should only be merged after 1.9.1 has finished releasing


irbekrm (approved):
Thanks for making the release steps more clear!

/lgtm

PR #1049, opened 2022-08-01
docs: Update Vault issuer docs for k8s 1.24

hawksight:
In response to the issue: https://github.com/cert-manager/cert-manager/issues/4144
I made [a comment](https://github.com/cert-manager/cert-manager/issues/4144#issuecomment-1181815631) about how to workaround the change. 
This PR is to attempt to address the k8s 1.24 change in the cert-manager documentation


hawksight:
I've tested this on kind with k8s 1.24:
```
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.0", GitCommit:"4ce5a8954017644c5420bae81d72b09b735c21f0", GitTreeState:"clean", BuildDate:"2022-05-19T15:39:43Z", GoVersion:"go1.18.1", Compiler:"gc", Platform:"linux/amd64"}
```

This pattern also matches the note on the [hashicrop resource](https://learn.hashicorp.com/tutorials/vault/kubernetes-cert-manager?in=vault/kubernetes#kubernetes-1-24-only) too.

content/docs/configuration/vault.md
```diff
@@ -154,38 +154,105 @@ Account Tokens [here](https://www.vaultproject.io/docs/auth/kubernetes.html).
...
+      vault:
+        path: pki_int/sign/example-dot-com
+        server: https://vault.local
+        caBundle: <base64 encoded caBundle PEM file>
+        auth:
+          kubernetes:
+            role: my-app-1
+            mountPath: /v1/auth/kubernetes
+            secretRef:
+              name: my-service-account-token-hvwsb
```
irbekrm, line 207: Nit: make this a placeholder for secret name like the caBundle on line 201

irbekrm (approved):
Thanks @hawksight this makes perfect sense, we actually made the same change in our e2e tests where we need the service account `Secret`s and it works just fine https://github.com/cert-manager/cert-manager/commit/ac956abb0c9a33a536b5746a1e4884ce3d9c7465

I've added a very minor nit, apart from that happy to lgtm, I think this will be very useful for folks. This should work fine with scripts/automation too which creates SA and Secret at the same time (tested it by hand).

/hold
/lgtm

hawksight:
Just a question for maintainers whether the headings used for the k8s examples makes sense when you look at the navigation bar to the right. Screenshot of what I mean:
Screenshot from 2022-08-08 16-07-05

I'm not sure the side bar makes it apparent they are sub sections of "Authenticating with Kubernetes Service Accounts". Perhaps I should replace the heading text with just bold or some other formatting so it doesn't appear in the side bar?

Or beyond a certain heading level, does the site ignore the heading?

hawksight:
Thanks @irbekrm

> IMO the header could stay as is in this PR

Could you readd the `/lgtm` label please? (since I changes the headers to remove `<=` and `>=` symbols)

irbekrm:
Thanks for working on this @hawksight !

/lgtm
/hold cancel

hawksight:
@irbekrm it appeared to be some spelling issues causing the failure. I hope I've corrected them in [23f2d19](https://github.com/cert-manager/website/pull/1049/commits/23f2d190725386d4f473f899867e1f91b2fd3334) if you could re-label this PR again. Re:

> New changes are detected. LGTM label has been removed.

Is there a pre-commit in this repo btw?

irbekrm:
> Is there a pre-commit in this repo btw?

No, there isn't..
You're thinking to run the verify script locally before committing? Yeah that could be good. One minor issue is that the verify script requires various node things to be installed though- wouldn't want to make it a requirement for external contributors who are just submitting some small fix. However, perhaps it's possible to somehow package the dependencies into an image for example

hawksight:
> You're thinking to run the verify script locally before committing? Yeah that could be good. One minor issue is that the verify script requires various node things to be installed though- wouldn't want to make it a requirement for external contributors who are just submitting some small fix. However, perhaps it's possible to somehow package the dependencies into an image for example

Pre-commit is opt-in in that you have to have run the setup task for a repository otherwise it takes no effect and acts like normal. So that should help, that most don't need to do anything and can commit / PR as normal.

Yer it doesn't have to be everything, but we could just keep it to simple things, like spell checking. There are some built in actions for some simpler stuff. Mainly I was just thinking of ways to avoid the cycle of, review, label then it fails for a spelling or formatting issue.... and then it needs reviewed and labelled again.

Also thank you for seeing this through! :)

PR #1052, opened 2022-08-04
render the side menu when browsing without javascript

jahrlin:
This PR makes it so that the page renders the left menu when browsing without javascript.

The page will be very long because there are many navigation items in the menu, but this makes the website usable at least.

What do you think?

See example screenshot below.

image


wallrj (approved):
Thanks @jahrlin 

It works well for me.

Sorry for delayed review.

/lgtm

PR #1056, opened 2022-08-12
Improve the layout of the left-hand menu

wallrj:
…and one less for the content

On displays `>=1280px` the left-hand menu was too narrow to display the nested menu items clearly,
On smaller displays the [responsive CSS](https://tailwindcss.com/docs/responsive-design) actually made the menu larger.
So we've widened it by 1 column on displays `>=1280px`  and reduced the width of the content by 1 column to compensate.
This makes the menu much easier to read on laptop and desktop computer screens.

We fixed an inconsistency in the vertical spacing between menu items with sub-menus and those without.

And finally, we moved the version selector to the bottom of the side-bar to avoid distracting the reader.
Screen Shot 2022-08-12 at 14 52 44

### DIFF
--- a/components/docs/Sidebar/Dropdown.jsx
+++ b/components/docs/Sidebar/Dropdown.jsx
@@ -15,7 +15,7 @@ export default function Dropdown({
   })
 
   return (
-    <div className="mt-4">
+    <div className="pt-2 pb-2">
       <button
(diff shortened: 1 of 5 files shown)


jahrlin (approved):
Thanks @wallrj, looks and feels much better!

/lgtm
/approve

PR #1057, opened 2022-08-13
Improve the table-of-contents 

wallrj:
**Preview**: https://deploy-preview-1057--cert-manager-website.netlify.app/docs/tutorials/getting-started-with-cert-manager-on-google-kubernetes-engine-using-lets-encrypt-for-ingress-ssl/

* Made it so that the TOC items do not wrap and overflow the container column (which I think looks neater, given that the right hand column is so narrow).
* Made it possible to filter out the headings above level 2 (including the level-3 headings made the menu look too noisy)
* Removed some (I think) unnecessary HTML wrapper elements and CSS classes.

New:
image

Current:
image

/cc @PatrickHeneise

### DIFF
--- a/components/docs/Toc.jsx
+++ b/components/docs/Toc.jsx
@@ -11,38 +11,30 @@ export default function Toc({ contents }) {
 
   return (
     contents.length > 0 && (
-      <div className="sticky top-4">
-        <div className="flex flex-col flex-grow rounded-lg">
-          <div className="flex items-center flex-shrink-0 px-2 pb-8">
-            <div className="flex-grow flex flex-col">
-              <nav className="flex-1 px-2 space-y-1">
-                <h4 className="text-blue-900 text-xs uppercase font-bold mt-6">
-                  On this page
-                </h4>
-                <ul>
-                  {contents.map((item) => {
-                    return (
-                      <TocMenuItem
-                        key={item.slug}
-                        slug={item.slug}
-                        raw={item.raw}
-                        text={item.text}
-                        isActive={
+        <nav className="sticky top-4">
+          <h4 className="text-blue-900 text-xs uppercase font-bold mb-2">
+            On this page
+          </h4>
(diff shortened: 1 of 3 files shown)


wallrj:
I'm also inclined to remove the react-scroll component, since it doesn't seem to work properly anyway: https://github.com/wallrj/website/compare/toc-improvements...wallrj:website:remove-react-scroll

* Doesn't currently hightlight the in-view section, because we don't set the active-class
* Even then, it doesn't highlight the heading consistently as you scroll up and down
* The back - forward button navigation doesn't always seem to work.

I can remove it in a followup PR if others agree.
Or replace it with something simpler which does at least reliably highlight the TOC entry for the in-view section. Maybe https://github.com/makotot/GhostUI

SgtCoDFish:
I don't currently have the bandwidth to review cert-manager PRs, sorry. I've removed myself as a reviewer :+1:

jahrlin:
> I'm also inclined to remove the react-scroll component, since it doesn't seem to work properly anyway: [wallrj/website@toc-improvements...wallrj:website:remove-react-scroll](https://github.com/wallrj/website/compare/toc-improvements...wallrj:website:remove-react-scroll)
> 
> * Doesn't currently hightlight the in-view section, because we don't set the active-class
> * Even then, it doesn't highlight the heading consistently as you scroll up and down
> * The back - forward button navigation doesn't always seem to work.
> 
> I can remove it in a followup PR if others agree. Or replace it with something simpler which does at least reliably highlight the TOC entry for the in-view section. Maybe https://github.com/makotot/GhostUI

I'm a big +1 to remove it, stuff like that interferes with native behaviour and is an accessibility problem.

jahrlin (approved):
Nice, this is great. Tested in a few different viewports and I really like it.

Much cleaner visually and much cleaner code.

/lgtm
/approve

PR #1058, opened 2022-08-15
Remove react-scroll because it doesn't work

wallrj:
In https://github.com/cert-manager/website/pull/1057#issuecomment-1214753222 @wallrj  wrote:

I'm inclined to remove the [react-scroll component](https://github.com/fisshy/react-scroll), since it doesn't seem to work properly.

* Doesn't currently hightlight TOC menu item for the the in-view section, because we don't set the active-class
* Even then, it doesn't highlight the heading consistently as you scroll up and down
* Sporadically updates the location bar hash tag as you scroll down ( but never when I scroll up in Firefox).
* The back - forward button navigation doesn't always seem to work.
* Confusing in so far as  when I highlight the current location bar in order to copy-paste the URL I find that a hastag has been added to the URL simply because I scrolled down the document. I don't want to copy a link to the end of the document just because I happen to have read to the bottom of the document.

/cc @PatrickHeneise


jahrlin (approved):
Thanks @wallrj!

/lgtm
/approve

PR #1064, opened 2022-08-26
Add a dedicated Troubleshooting section

wallrj:
This is another chunk of the changes that we explored in https://github.com/cert-manager/website/pull/1048.

Here we move the troubleshooting material out of the FAQ section and into a top-level troubleshooting section.

We've also tried to improve the page titles and descriptions of those pages that we've moved.

/cc @maelvls @mehak151 @jsoref


content/docs/concepts/webhook.md
```diff
@@ -1,32 +1,34 @@
 ---
-title: Webhook
-description: 'cert-manager core concepts: Webhook'
+title: All About the cert-manager Webhook
+description: |
+    Learn about the webhhook component of cert-manager, which validates, converts and sets default values for the cert-manager custom resources
```
jsoref, line 4:
```suggestion
    Learn about the webhook component of cert-manager, which validates, converts and sets default values for the cert-manager custom resources
```

content/docs/concepts/webhook.md
```diff
@@ -1,32 +1,34 @@
...
 ---
 
-cert-manager makes use of extending the Kubernetes API server using a Webhook
-server to provide [dynamic admission
-control](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)
-over cert-manager resources. This means that cert-manager benefits from most of
-the same behavior that core Kubernetes resource have. The webhook has three
-main functions:
+cert-manager extends the Kubernetes API using Custom Resource Definitions
+and it installs a webhook which has three main functions:
```
jsoref, line 8: I'm not a fan of `and it` here.

One approach would be:
```suggestion
cert-manager extends the Kubernetes API using Custom Resource Definitions.
It installs a webhook which has three main functions:
```

Note that `functions` can have two meanings, one `a programming function` and the other roughly `purpose/task`. I think you're using it in the latter meaning, but since we're programmers first, I think people are more likely to parse it in the former. That said, I can't think of a word I like better here for the latter meaning.

content/docs/troubleshooting/README.md
```diff
@@ -1,8 +1,20 @@
 ---
 title: Troubleshooting
-description: 'cert-manager FAQ: Troubleshooting'
+description: |
+    Learn how to debug common problems with cert-manager
 ---
 
+In this section, you will learn troubleshooting techniques that will help you find the root cause if your Certificate fails to be issued or renewed.
```
jsoref, line 7: I can't quite explain why I want to make this change:
```suggestion
In this section, you will learn troubleshooting techniques that will help you find the root cause when your Certificate fails to be issued or renewed.
```

(I'm not attached to it--I'd be more attached if I could explain my rationale.)

The same change would apply in other instances, but only consider it/them if someone else can justify the change in general.

content/docs/troubleshooting/webhook.md
```diff
@@ -0,0 +1,56 @@
...
+    Learn how to diagnose problems with the cert-manager webhook, such as connection errors or invalid or missing TLS certificate errors
+---
+
+Learn how to diagnose problems with the cert-manager webhook, such as connection errors or invalid or missing TLS certificate errors.
+
+## Overview
+
+You may encounter errors when creating cert-manager resources indicating that the webhook is unreachable.
+In this case, it is advised to first check the [compatibility](../installation/compatibility.md) of your environment
+and take necessary action outlined there.
```
jsoref, line 13: Possibly:
```suggestion
and take necessary actions outlined there.
```

content/docs/troubleshooting/webhook.md
```diff
@@ -0,0 +1,56 @@
...
+Then refer to the [known webhook problems](../concepts/webhook.md#known-problems-and-solutions).
+And finally, if you are still having problems follow the steps below.
+
+## Check the webhook TLS certificates
+
+The Kubernetes API server will load the CA content from the webhook configuration and use that to verify the serving certificate presented by the webhook server, when the TLS connection is established.
```
jsoref, line 19:
```suggestion
The Kubernetes API server will load the CA content from the webhook configuration and use that to verify the serving certificate presented by the webhook server when the TLS connection is established.
```

content/docs/concepts/webhook.md
wallrj: Duties? I'll leave it as "functions" for now.

content/docs/troubleshooting/README.md
wallrj: We didn't want to imply that it would *inevitably* fail :-) I'll leave it for now.

wallrj (commented):
Thanks @jsoref . I've applied most of your suggestions.

maelvls:
Thanks @jsoref for the review!
The changes look great.
/lgtm

PR #1072, opened 2022-09-08
Adds documentation for the new `caBundleSecretRef` feature on the Vault Issuer

JoshVanL:
Ref: https://github.com/cert-manager/cert-manager/pull/5387

PR is against next-docs so can be merged straight away.

### DIFF
--- a/content/next-docs/configuration/vault.md
+++ b/content/next-docs/configuration/vault.md
@@ -48,7 +48,7 @@ spec:
       ...
 ```
 
-The CA Bundle can also be optionally defined by referencing a Secret. The Secret
+The CA Bundle can also optionally be defined by referencing a Secret. The Secret
 should exist in the same Namespace as that of the Issuer, or in the [`Cluster
 Resource Namespace` (default cert-manager)][] when using a ClusterIssuer.
 
--- a/content/next-docs/configuration/vault.md
+++ b/content/next-docs/configuration/vault.md
@@ -25,7 +25,8 @@ namely the server, path, and CA bundle:
   *must* use the `sign` endpoint.
 - CA bundle denotes an optional field containing a base64 encoded string of the
   Certificate Authority to trust the Vault connection. This is typically
-  _always_ required when using an `https` URL.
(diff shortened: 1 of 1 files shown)


content/next-docs/configuration/vault.md
```diff
@@ -47,6 +48,33 @@ spec:
       ...
 ```
 
+The CA Bundle can also be optionally defined by referencing a Secret. The Secret
```
jsoref, line 51:
```suggestion
The CA Bundle can also optionally be defined by referencing a Secret. The Secret
```

wallrj (changes requested):
Thanks @JoshVanL 

1. I'm not sure if the `next-docs` folder works. There's no preview on Netlify and I think that might be because I broke it while implementing a nicer version selector button.
2. In any case, how will this be merged into the `docs/` folder upon the next release? Since there have been many changes in that folder which will not have been duplicated to `next-docs/`.
  3. Perhaps instead there should be a staging branch for the next documentation release and we should merge this into that branch and eventually merge that into the main branch at the time of the release.
4. Why did we use a Secret for the CA Bundle? The CA certs are public. I worry that using a Secret ref encourages people to mount the Secret that may also contain the private key and serving certificate, which may or may not have been created by cert-manager....but in any case is  bad practice which is now documented here: https://cert-manager.io/docs/concepts/certificate/

> When configuring a client to connect to a TLS server with a serving certificate that is signed by a private CA, you will need to provide the client with the CA certificate in order for it to verify the server. ca.crt will likely contain the certificate you need to trust, but do not mount the same Secret as the server to access ca.crt.

5. We're currently talking about the name of the `trust` project, and one thing I learned is that that widely accepted term for these files containing one or more PEM encoded CA certs is `truststore`...So would it be more consistent to call this field `trustStoreRef` ?

I'd be happy to merge this if there are answers to all those points.

wallrj:
The `issuer.vault.spec.caBundleSecretRef` docs are missing 

I searched cert-manager.io for   caBundleSecretRef and found only the release note:
 * https://cert-manager.io/docs/releases/release-notes/release-notes-1.10/#feature-1

PR #1073, opened 2022-09-08
Minor edits

jsoref:
One change fixes this strange rendering:
image

VSCode seems to expect something like this to happen, although I don't know markdown well enough to understand why.

Some changes to make things more consistent.


maelvls:
This strange rendering is so weird... Both the VSCode higheliger and `@next/mdx` (the markdown parser we are using) are understanding it as an element of an ordered list:
```md
10250. 
```

One workaround is to escape the `.`:

```md
10250\. 
```

What do you think?

jsoref:
I'm quite happy w/ using backticks, as seen in https://deploy-preview-1073--cert-manager-website.netlify.app/docs/troubleshooting/webhook/:
image

But, if you'd prefer to not decorate port numbers and instead use a backslash against the `.`, I can live with that. I'm not attached to any particular change. Anything that fixes the current rendering is an improvement.

maelvls:
I actually like the fact that these port numbers "stand out" thanks to the backticks. Let's go with that.

/lgtm

wallrj (approved):
/approve
/ok-to-test

PR #1074, opened 2022-09-08
Fix the open graph tag "og:title"

maelvls:
The `og:title` HTLM meta tag is important because that's the first thing people see when they share a cert-manager.io link in Twitter or on Slack. Today, all the pages have the same `og:title`: "cert-manager - Documentation".

The title of the page that we set in each front matter is discarded. This change fixes that.

To reproduce the before and after commands below, first run:

```
./scripts/server-netlify
```

**Before:**

```sh
$ curl -sL localhost:8888/docs/tutorials/getting-started-with-cert-manager-on-google-kubernetes-engine-using-lets-encrypt-for-ingress-ssl | htmlq meta | grep og:title
<meta content="cert-manager - Documentation" property="og:title">
```

**After:**

```sh
$ curl -sL localhost:8888/docs/tutorials/getting-started-with-cert-manager-on-google-kubernetes-engine-using-lets-encrypt-for-ingress-ssl | htmlq meta | grep og:title
<meta content="Deploy cert-manager on Google Kubernetes Engine (GKE) and create SSL certificates for Ingress using Let's Encrypt" property="og:title">
```

### DIFF
--- a/pages/[...docs].jsx
+++ b/pages/[...docs].jsx
@@ -31,7 +31,7 @@ const DocumentationPage = ({
         canonical={currentUrl}
         openGraph={{
           url: currentUrl,
-          title: page.title,
+          title: frontmatter.title,
           description: frontmatter.description
         }}
       />


maelvls:
Maybe we should append `— cert-manager documentation` to the titles?

wallrj (approved):
Thanks. I think it looks a lot better.

image


/lgtm

PR #1077, opened 2022-09-12
Mark ADCS issuer as not maintained

tongpu:
According to their README ADCS issuer is not maintained anymore


JoshVanL:
Thanks @tongpu!

Rather than removing the line completely, can we please instead add a `[NOT MAINTAINED]` label on it for prosperity?

tongpu:
@JoshVanL Good point. I've added the label.

PR #1071, opened 2022-09-07
Improved the summary on the docs homepage

wallrj:
**Preview**: https://deploy-preview-1071--cert-manager-website.netlify.app/docs/

The opening paragraph of the `docs/` homepage was a bit waffly, 
so we've tried to get straight to the point and linked to the internal pages about each supported CA rather than sending visitors off to external sites before they've read about the cert-manager integration with each of the CAs.

PR #1078, opened 2022-09-12
Cherry-pick the parts of "@zentered/next-product-docs" that we need

maelvls:
It is becoming hard to do changes to the website due to most of the MDX settings being done in the "@zentered/next-product-docs" npm library. I propose to cherry-pick the files that we need instead.

For context, the library [zentered/next-product-docs](https://github.com/opstrace/next-product-docs/) does three things for us:

1. It gives us the functions `staticPaths` and `staticProps` which are needed when using Next.js. These two functions work by reading `manifest.json`. 
2. It gives us four JSX components: `<Documentation>`, ~`<Interpolate>`~, ~`<Tabs>`~, `<CodeBlock>`, and `<InlineCode>`.
3. It gives us three remark plugins: `links`, `import`, and `images` (the others, such as `header-ids`, were broken).

I am OK with (2) and (3), but (1) is a problem because we can't control which rehype and redoc plugins are used:

- In [this thread](https://github.com/cert-manager/website/pull/1071#discussion_r965883963), accessing the `rehypePlugins` configuration would allow us to fix the issue,
- The code snippets can't be customized (e.g., the font size, the hightlithing), and could be fixed by being able to access the `rehypePlugins` too.

## Why not do the changes upstream?

Making these rehype and redoc changes "upstream" (i.e., the repository https://github.com/opstrace/next-product-docs/) will break Opstrace's website, most probably.

## Why not vendoring the whole library in a `vendor/` folder?

Vendoring in a given folder would have the benefit of letting us upstream changes, as well as benefiting from the changes made upstream. But


package.json
```diff
@@ -48,18 +48,26 @@
...
   },
+  "browser": {
+    "fs": false,
+    "path": false,
+    "child_process": false
+  },
   "dependencies": {
     "@docsearch/react": "^3.0.0",
     "@headlessui/react": "^1.5.0",
     "@zentered/next-product-docs": "^0.0.127",
```
inteon, line 59: why do we still depend on `@zentered/next-product-docs` here?

maelvls: Ah, I forgot to remove that line.

maelvls:
Update: we discussed in this morning's standup that we aren't sure what to do. 

- My approach is to "merge" the Zentered library into our own code. The reasoning is that I think most of this code is meant to be ad-hoc on a per-site manner (e.g. rehype, redoc configurations). The only part that would make sense to have in common is the sidebar feature (that uses the `manifest.json` file).
- Richard suggests to "vendor" the library (i.e., copy verbatim the library in `vendor/`) so that we can upstream the changes later on. Richard made the point that Zentered will continue maintaining and making changes to https://npm.io/package/@zentered/next-product-docs.

This PR is paused so that we can discuss that.

components/docs/CodeBlock.jsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
```
wallrj, line 1: I think all these files have  Apache Licence. 
* https://github.com/zentered/next-product-docs/blob/main/LICENSE

And I think we should add a `(c) Copright Zentered` at the top of each of the imported files with a link to the upstream version on GitHub.

lib/index.jsx
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable react/display-name */
+import React from 'react'
+import { MDXRemote } from 'next-mdx-remote'
+// import { State, Observe } from 'mdx-observable'
+// import Tabs, { Tab } from './components/Tabs.jsx'
+// import Interpolate from './components/Interpolate.jsx'
```
wallrj, line 6: Remove all these commented imports and the target files if they are not used.

lib/remark-plugins/heading-ids/index.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * Forked from https://github.com/imcuttle/remark-heading-id
+ * License MIT
+ * @author imcuttle
+ */
```
wallrj, line 5: I don't know what copyright to add here. Probably both the original author and zentered if they modified the code here.

lib/serialize.js
```diff
@@ -0,0 +1,161 @@
+import { serialize } from 'next-mdx-remote/serialize'
+import matter from 'gray-matter'
+import { marked } from 'marked'
+import GithubSlugger from 'github-slugger'
+import rehypeSlug from 'rehype-slug'
+// import rehypeSanitize from 'rehype-sanitize'
```
wallrj, line 6: Remove this commented line and the target package.

wallrj (changes requested):
Thanks @maelvls 

The new PR title makes it clear what you're doing here.

I think all the copied files should have copyright headers giving credit to Zentered and any other authors.
And there should be a link at the top of  each  copied file to the exact source file in the origin branch, so that we can more easily keep track of what changes have been made upstream.

Remove all the commented imports and remove the source files or package.json entries for those no-longer-imported files.

lib/remark-plugins/links/index.js
```diff
@@ -0,0 +1,103 @@
+import { visit } from 'unist-util-visit'
```
wallrj, line 1: What's the origin and licence of this file?

components/docs/CodeBlock.jsx
```diff
@@ -0,0 +1,71 @@
+// This file originates from https://github.com/zentered/next-product-docs,
```
wallrj, line 1: Sorry to be a pain, but I think it's worth being more precise than that. I.e.

 * https://github.com/zentered/next-product-docs/blob/f4fb801ce08b393466b4af484b42266763edaf43/src/components/CodeBlock.jsx

There's no record in the PR or in the files that they were copied from cert-manager-adjustments branch:
https://github.com/zentered/next-product-docs/tree/feat/cert-manager-adjustments rather than from a tag of the main branch.

Consider adding a little more detail about the precise origin and then I'll be happy to add a lgtm

wallrj (changes requested):
Thanks. There's one file still missing a header and I made a suggestion below.

wallrj (changes requested):
Also there are some build failures.

components/docs/CodeBlock.jsx
maelvls: I changed it to:

```go
// This file originates from https://github.com/zentered/next-product-docs on
// the branch "feat/cert-manager-adjustments" (commit f4fb801), copyright
// Zentered 2022, licensed under the Apache 2.0 license.
```

lib/remark-plugins/links/index.js
maelvls: Oops, I missed this one. I added the license header.

wallrj (changes requested):
One last question. Have you made any modifications to the code in the files imported in this PR?
If so, what were the changes and what UI changes should I be looking for?

For example, I tried copying and pasting a snippet of yaml from the first code block after: 
* https://deploy-preview-1078--cert-manager-website.netlify.app/docs/configuration/selfsigned/
 and the first time I did it I  found that it was not properly indented:

```
$ cat > foo
apiVersion: cert-manager.io/v1
kind: Issuer
metadata:
name: selfsigned-issuer
namespace: sandbox
spec:
selfSigned: {}
^C

```

I refreshed the page and then the snippet was correctly indented. 

Perhaps that's an existing bug.

maelvls:
Except for the `import ... from ...` statement that we removed and some code being moved around (one of them being that the `<Documentation>` component was merged into the `[...docs].jsx` file), I have made no fixes. 

I am not sure why the first "Paste" didn't contain the correct indentation though 😔

PR #1081, opened 2022-09-30
Vault: document the new field "serviceAccountRef"

maelvls:
| Preview: https://deploy-preview-1081--cert-manager-website.netlify.app/docs/configuration/vault/#authenticating-with-a-kubernetes-service-account-without-a-secret |
|:--|

The feature itself is being developed in ~https://github.com/cert-manager/cert-manager/pull/4524~ https://github.com/cert-manager/cert-manager/pull/5502.

### DIFF
--- a/content/docs/configuration/vault.md
+++ b/content/docs/configuration/vault.md
@@ -163,7 +163,7 @@ Kubernetes service account token can be provided in two ways:
 
 With the secretless authentication with a service account, cert-manager creates
 an ephemeral service account token using the TokenRequest API and uses it to
-authenticates with Vault. These tokens are short-lived (10 minutes) and are
+authenticate with Vault. These tokens are short-lived (10 minutes) and are
 never stored to disk.
 
 This is the recommended authentication method because it does not rely on the
--- a/content/docs/configuration/vault.md
+++ b/content/docs/configuration/vault.md
@@ -144,6 +144,8 @@ spec:
           key: token
 ```
 
+<a id="static-service-account-token"></a>
+
 ### Authenticating with Kubernetes Service Accounts
 
 Vault can be configured so that applications can authenticate using Kubernetes
@@ -253,6 +255,101 @@ Kubernetes 1.24 and above.
               name: vault-issuer-token
               key: token
     ```
+
+### Authenticating with a Kubernetes Service Account Without A Secret
+
+The [previous method](#static-service-account-token) of authenticating with
+Vault has a major disadvantage: it relies on a less secure "static" token (by
+"static", we mean a token that does not have an expiry time).
+
+Using the field `serviceAccountRef` instead of `secretRef`, you can let
+cert-manager request ephemeral tokens.
+
+The first step is to create a `ServiceAccount` resource, like in the `secretRef`
+method:
+
+```sh
+kubectl create serviceaccount -n sandbox vault-issuer
+```
+
+Then, you will need to add an RBAC Role so that cert-manager can use the service
+account:
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: vault-issuer
+  namespace: sandbox
+rules:
+  - apiGroups: ['']
+    resources: ['serviceaccounts/token']
+    resourceNames: ['vault-issuer']
+    verbs: ['create']
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: vault-issuer
+  namespace: sandbox
+subjects:
+  - kind: ServiceAccount
+    name: cert-manager
+    namespace: cert-manager
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: vault-issuer
+```
+
+Finally, you can create the Issuer resource:
+
+```yaml
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: vault-issuer
+  namespace: sandbox
+spec:
+  vault:
+    path: pki_int/sign/example-dot-com
+    server: https://vault.local
+    auth:
+      kubernetes:
+        role: my-app-1
+        mountPath: /v1/auth/kubernetes
+        serviceAccountRef:
+          name: vault-issuer
+```
+
+To create the role in Vault, you can use the following command:
+
+```bash
+vault write auth/kubernetes/role/vault-issuer \
+    bound_service_account_names=vault-issuer \
+    bound_service_account_namespaces=sandbox \
+    audience="vault://sandbox/vault-issuer" \
+    policies=vault-issuer \
+    ttl=1m
+```
+
+It is recommended to use a different Vault role each per Issuer or
+ClusterIssuer. The `audience` allows you to restrict the Vault role to a single
+Issuer or ClusterIssuer. The syntax is the following:
+
+```yaml
+"vault://<namespace>/<issuer-name>"   # For an Issuer.
+"vault://<issuer-name>"               # For a ClusterIssuer.
+```
+
+The expiration duration for the Kubernetes tokens that are requested is
+hard-coded to 10 minutes (that's the minimum accepted). The `ttl` field can be
+as short as possible, since cert-manager requests a new token every time it
+needs to talks to Vault.
+
+Although it is not recommended, you can also use the same Vault role for all of
+your Issuers and ClusterIssuers by omitting the `audience` field and re-using
+the same service account.
 
 ## Verifying the issuer Deployment
 
--- a/content/docs/configuration/vault.md
+++ b/content/docs/configuration/vault.md
@@ -230,7 +230,7 @@ spec:
 > **Issuer vs. ClusterIssuer:** With an Issuer resource, you can only refer to a
 > service account located in the same namespace as the Issuer. With a
 > ClusterIssuer, the service account must be located in the namespace that is
-> configured by the flag `--cluster-scoped namespace`.
+> configured by the flag `--cluster-resource-namespace`.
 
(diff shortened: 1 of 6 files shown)


content/docs/configuration/vault.md
```diff
@@ -254,6 +256,101 @@ Kubernetes 1.24 and above.
...
+```
+
+The expiration duration for the Kubernetes tokens that are requested is
+hard-coded to 10 minutes (that's the minimum accepted). The `ttl` field can be
+as short as possible, since cert-manager requests a new token every time it
+needs to talks to Vault.
+
+Although it is not recommended, you can also use the same Vault role for all of
+your Issuers and ClusterIssuers by omitting the `audience` field and re-using
+the same service account.
```
maelvls, line 352: Hmm. I wonder if most people would go with this instead of bothering with `"vault://<namespace>/<issuer-name>"`.

content/docs/configuration/vault.md
```diff
@@ -254,6 +256,101 @@ Kubernetes 1.24 and above.
               key: token
     ```
 
+### Authenticating with a Kubernetes Service Account Without A Secret
+
+The [previous method](#static-service-account-token) of authenticating with
+Vault has a major disadvantage: it relies on a less secure "static" token (by
+"static", we mean a token that does not have an expiry time).
```
wallrj, line 263: Don't start by dismissing the previous method. Why not just put this at the top, as the recommended method, and  start by explaining its advantages.

content/docs/configuration/vault.md
```diff
@@ -254,6 +256,101 @@ Kubernetes 1.24 and above.
...
+
+Using the field `serviceAccountRef` instead of `secretRef`, you can let
+cert-manager request ephemeral tokens.
```
wallrj, line 266: I think you should elaborate more on what it means to "request ephemeral tokens".
For example: https://cert-manager.io/docs/configuration/acme/dns01/azuredns/#managed-identity-using-aad-workload-identity

content/docs/configuration/vault.md
```diff
@@ -254,6 +256,101 @@ Kubernetes 1.24 and above.
...
+
+The first step is to create a `ServiceAccount` resource, like in the `secretRef`
+method:
```
wallrj, line 269: Don't refer to the secretRef method.

content/docs/configuration/vault.md
```diff
@@ -254,6 +256,101 @@ Kubernetes 1.24 and above.
...
+
+```sh
+kubectl create serviceaccount -n sandbox vault-issuer
+```
+
+Then, you will need to add an RBAC Role so that cert-manager can use the service
+account:
```
wallrj, line 276:
```suggestion
Then add an RBAC Role so that cert-manager can get tokens for the ServiceAccount:
```

content/docs/configuration/vault.md
```diff
@@ -254,6 +256,101 @@ Kubernetes 1.24 and above.
...
+  - kind: ServiceAccount
+    name: cert-manager
+    namespace: cert-manager
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: vault-issuer
+```
+
+Finally, you can create the Issuer resource:
```
wallrj, line 305:
```suggestion
Finally, create the Issuer resource:
```

content/docs/configuration/vault.md
```diff
@@ -254,6 +256,101 @@ Kubernetes 1.24 and above.
...
+  vault:
+    path: pki_int/sign/example-dot-com
+    server: https://vault.local
+    auth:
+      kubernetes:
+        role: my-app-1
+        mountPath: /v1/auth/kubernetes
+        serviceAccountRef:
+          name: vault-issuer
+```
```
wallrj, line 323: Add a note about Issuer vs ClusterIssuer and how Issuer can only refer to a ServiceAccount in the same namespace while ClusterIssuer must refer to a ServiceAccount in the `--cluster-scoped` namespace.

wallrj (changes requested):
Nice start to the documentation for this feature, but please add some narrative and some  links to Hashicorp Vault documentation explaining how this style of authentication works.

Shouldn't we also explain how to configure vault for this authentication (both for in-cluster vault and out-of-cluster vault installations) or is that explained elsewhere in this document?

Also please rebase this onto the `release-next` branch, adding a release note to go with it.

content/docs/configuration/vault.md
maelvls: Thanks. I added the following:

> **Issuer vs. ClusterIssuer:** With an Issuer resource, you can only refer to a
> service account located in the same namespace as the Issuer. With a
> ClusterIssuer, the service account must be located in the namespace that is
> configured by the flag `--cluster-resource-namespace`.

content/docs/configuration/vault.md
maelvls: I added a paragraph on "request ephemeral tokens". Hopefully it makes more sense now.

content/docs/configuration/vault.md
maelvls: I move the section at the top. Or rather, I moved it right before the existing service account authentication.

maelvls:
> Shouldn't we also explain how to configure vault for this authentication (both for in-cluster vault and out-of-cluster vault installations)

The only documentation that exists in the page is the `vault write auth...` command that I added. It only focuses on in-cluster installations.

content/docs/configuration/vault.md
```diff
@@ -144,13 +144,125 @@ spec:
...
+- [Authentication with a Static Service Account Token](#static-service-account-token).
+
+#### Secretless Authentication with a Service Account
+
+ℹ️ This feature is available in cert-manager >= v1.12.0.
+
+With the secretless authentication with a service account, cert-manager creates
+an ephemeral service account token using the TokenRequest API and uses it to
+authenticates with Vault. These tokens are short-lived (10 minutes) and are
+never stored to disk.
```
wallrj, line 167:
```suggestion
With the secretless authentication with a service account, cert-manager creates
an ephemeral service account token using the TokenRequest API and uses it to
authenticate with Vault. These tokens are short-lived (10 minutes) and are
never stored to disk.
```

content/docs/configuration/vault.md
```diff
@@ -144,13 +144,125 @@ spec:
...
+        role: my-app-1
+        mountPath: /v1/auth/kubernetes
+        serviceAccountRef:
+          name: vault-issuer
+```
+
+> **Issuer vs. ClusterIssuer:** With an Issuer resource, you can only refer to a
+> service account located in the same namespace as the Issuer. With a
+> ClusterIssuer, the service account must be located in the namespace that is
+> configured by the flag `--cluster-scoped namespace`.
```
wallrj, line 233:
```suggestion
> **Issuer vs. ClusterIssuer:** With an Issuer resource, you can only refer to a
> service account located in the same namespace as the Issuer. With a
> ClusterIssuer, the service account must be located in the namespace that is
> configured by the flag `--cluster-resource-namespace`.
```

See https://deploy-preview-1081--cert-manager-website.netlify.app/docs/cli/controller/

content/docs/configuration/vault.md
```diff
@@ -144,13 +144,125 @@ spec:
...
+```
+
+It is recommended to use a different Vault role each per Issuer or
+ClusterIssuer. The `audience` allows you to restrict the Vault role to a single
+Issuer or ClusterIssuer. The syntax is the following:
+
+```yaml
+"vault://<namespace>/<issuer-name>"   # For an Issuer.
+"vault://<issuer-name>"               # For a ClusterIssuer.
+```
```
wallrj, line 253:
```suggestion
```yaml
"vault://<namespace>/<issuer-name>"   # For an Issuer.
"vault://<cluster-issuer-name>"               # For a ClusterIssuer.
```
```

**Review by wallrj — CHANGES_REQUESTED** [review]

File: content/docs/contributing/release-process.md [hunk 11]
```diff
@@ -578,3 +578,32 @@ Other than the different `cert-manager/release` tag and `cmrel` version, you can
...
+to check that the "base branch" is correct.
+
+|                Description                 |  Base branch   |     Folder     |
+|--------------------------------------------|----------------|----------------|
+| Documentation PR for an existing feature   | `master`       | `content/docs` |
+| Documentation PR for an unreleased feature | `release-next` | `content/docs` |
+
+When creating a documentation PR for an unreleased feature, you will need to
+fast-forward the `release-next` branch to `master` before asking for a review of
+the PR:
```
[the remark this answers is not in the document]
maelvls: I moved these changes to https://github.com/cert-manager/website/pull/1192/files#r1118901920. Let us continue the conversation there.

maelvls:
I addressed the comments and have added a paragraph to the release-notes-1.12.md page. Please take another look.

package.json
```diff
@@ -22,7 +22,7 @@
     "generate:sitemap": "next-sitemap",
     "export": "next export",
     "start": "next start",
-    "check": "concurrently --group --timings npm:check:* # Run all the npm check:* scripts in parallel",
+    "check": "npm exec concurrently -y -- --group --timings npm:check:* # Run all the npm check:* scripts in parallel",
```
wallrj, line 25: What's the reason for this change? 
All the npm installed executables are installed in `.node_modules/.bin` by `npm ci` and are in PATH when these `npm run` scripts are run, so why do you add the `npm exec ... -y` here?

wallrj (changes requested):

package.json
wallrj: If the documentation in https://github.com/cert-manager/website#website-development-tooling is wrong, we should fix that.

wallrj (approved):
Thanks!

/lgtm

PR #1085, opened 2022-10-04
content/projects/csi-driver: Fix PKCS12 attribute keys to correct strings

JoshVanL:
fixes https://github.com/cert-manager/csi-driver/issues/117

The documentation has the old attribute keys which changed during code review.


jahrlin (approved):
/lgtm
/approve

PR #1089, opened 2022-10-15
Update docs to remove subchart warning

tylerauerbeck:
Since it appears that this is supported since v 1.9 and support is documented below the warning on the same page, this removes the warning for users to not use cert-manager as a sub chart.

Resolves https://github.com/cert-manager/cert-manager/issues/5503

### DIFF
--- a/content/docs/installation/helm.md
+++ b/content/docs/installation/helm.md
@@ -7,9 +7,6 @@ description: 'cert-manager installation: Using Helm'
 
 cert-manager provides Helm charts as a first-class method of installation on both Kubernetes and OpenShift.
 
-Be sure never to embed cert-manager as a sub-chart of other Helm charts; cert-manager manages
-non-namespaced resources in your cluster and care must be taken to ensure that it is installed exactly once.
-
 ### Prerequisites
 
 - [Install Helm version 3 or later](https://helm.sh/docs/intro/install/).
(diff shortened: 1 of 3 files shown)


content/docs/installation/helm.md
```diff
@@ -7,9 +7,6 @@ description: 'cert-manager installation: Using Helm'
 
 cert-manager provides Helm charts as a first-class method of installation on both Kubernetes and OpenShift.
 
-Be sure never to embed cert-manager as a sub-chart of other Helm charts; cert-manager manages
-non-namespaced resources in your cluster and care must be taken to ensure that it is installed exactly once.
```
wallrj, line 11: I suggest that we adjust this message and move it above the new sub-charts section e.g.

https://cert-manager.io/docs/installation/helm/#installing-cert-manager-as-subchart

## Installing cert-manager as subchart

> :warning: If you embed cert-manager as a sub-chart care must be taken to ensure that it is installed exactly once. cert-manager manages non-namespaced resources such as ClusterIssuer which will break if there are multiple cert-manager controllers installed. Likewise, the cert-manager webhook and cainjector components must only be installed once.

wallrj (changes requested):
Thanks @tylerauerbeck 

We've removed the `next-docs/` folder. Please rebase and make the change  also in  `v1.11-docs` and `v1.10-docs`.

Ping me for another review when you're done.

/ok-to-test
/approve

Issue #1063, opened 2022-08-25
"Securing Ingresses with Venafi" tutorial contains link to missing manifest


amcginlay:
See [here](https://cert-manager.io/docs/tutorials/venafi/venafi/#installing-ingress-nginx).

The manifest at https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/aws/service-nlb.yaml is no longer present.

I suspect this needs to be repointed to https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/aws/nlb-with-tls-termination/deploy.yaml but this has not been tested

amcginlay:
Furthermore I think https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/mandatory.yaml should become https://github.com/kubernetes/ingress-nginx/blob/main/deploy/static/provider/aws/deploy.yaml

amcginlay:
Also, in `venafi-issuer.yaml` the attribute `spec.venafi.cloud.url` is no longer required and its inclusion as specified causes a failure.
[This](https://cert-manager.io/docs/configuration/venafi/) documentation has it correct.

amcginlay:
Also, the certificate manifest needs a `commonName` so, for example:
```
spec:
  secretName: demo-jetstack-mcginlay.net-tls
  dnsNames:
  - demo.jetstack.mcginlay.net
  commonName: demo.jetstack.mcginlay.net
```

amcginlay:
May I also suggest we rename this tutorial from "Securing Ingresses with Venafi" to "EKS + Ingress + Venafi" so it matches the pattern established by "GKE + Ingress + Let's Encrypt".

Finally, this tutorial should be linked from the [Getting Started](https://cert-manager.io/docs/getting-started/) page, making sure to use appropriate iconography. One could argue that the link to the [Securing NGINX-ingress](https://cert-manager.io/docs/tutorials/acme/nginx-ingress/) tutorial should be removed to emphasize the public cloud use cases.

PR #1090, opened 2022-10-15
Update Venafi ingress example

tylerauerbeck:
Updates the Venafi ingress example per #1063


content/docs/tutorials/venafi/venafi.md
```diff
@@ -346,7 +346,6 @@ spec:
   venafi:
     zone: "Default" # Set this to the Venafi policy zone you want to use
     cloud:
-      url: "https://api.venafi.cloud/v1"
```
JoshVanL, line 349: Should we be removing this?

JoshVanL (changes requested):

PR #1091, opened 2022-10-16
Adding SSA related warnings to the affected versions

sathyanarays:
Signed-off-by: Sathyanarayanan Saravanamuthu <sathyanarays@.vmware.com>

## Description
- Added notes for versions 1.7, 1.8, 1.9 and 1.10 to enable SSA in the cluster
- Added a line to mandate SSA enabled while upgrading from 1.6 to 1.7
- - I haven't added this line to other upgrades (1.7 -> 1.8,...). These are not required. For example, if any customer is running version 1.7 successfully, that implicitly means SSA is enabled. So, no special instructions are required.

### DIFF
--- a/content/docs/installation/supported-releases.md
+++ b/content/docs/installation/supported-releases.md
@@ -16,23 +16,25 @@ release every two months.
 
 | Release  | Release Date |   End of Life   | [Supported Kubernetes versions][s] | [Supported OpenShift versions][s] |
 |----------|:------------:|:---------------:|:----------------------------------:|:---------------------------------:|
-| [1.10][] | Oct 17, 2022 | Release of 1.12 |            1.20 → 1.25             |             4.7 → 4.11            |
-| [1.9][]  | Jul 22, 2022 | Release of 1.11 |            1.20 → 1.24             |             4.7 → 4.11            |
+| [1.10][] | Oct 17, 2022 | Release of 1.12 |            1.20 → 1.25*            |            4.7 → 4.11             |
+| [1.9][]  | Jul 22, 2022 | Release of 1.11 |            1.20 → 1.24*            |            4.7 → 4.11             |
 
+\*ServerSideApply should be enabled in the cluster
 ## Upcoming releases
 
 | Release  |         Release Date          |  End of life   | [Supported Kubernetes versions][s] | [Supported OpenShift versions][s] |
 |----------|:-----------------------------:|:--------------:|:----------------------------------:|:---------------------------------:|
-| [1.11][] |   ~Late Dec 22/Early Jan 23   | Mid June, 2023 |            1.20 → 1.25             |             4.7 → 4.11            |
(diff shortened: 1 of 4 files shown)


content/docs/installation/supported-releases.md
```diff
@@ -16,22 +16,24 @@ release every two months.
...
 ## Upcoming releases
 
 | Release  |     Release Date    |    End of life     | [Supported Kubernetes versions][s] | [Supported OpenShift versions][s] |
 |----------|:-------------------:|:------------------:|:----------------------------------:|:---------------------------------:|
-| [1.10][] | ~Mid/Late Sep, 2022 | Mid January, 2023  |             1.20 → 1.24            |             4.7 → 4.11            |
+| [1.10][] | ~Mid/Late Sep, 2022 | Mid January, 2023  |            1.20 → 1.24*            |             4.7 → 4.11            |
 
 Dates in the future are uncertain and might change.
 
+\*ServerSideApply should be enabled in the cluster
```
JoshVanL, line 31:
```suggestion
\*ServerSideApply must be enabled in the cluster.

```

content/docs/installation/supported-releases.md
```diff
@@ -70,6 +72,7 @@ and release notes on [cert-manager.io](https://cert-manager.io/docs/release-note
 
 We also maintain detailed [upgrade instructions](https://cert-manager.io/docs/installation/upgrading/).
 
+\*ServerSideApply should be enabled in the cluster
```
JoshVanL, line 75:
```suggestion
\*ServerSideApply must be enabled in the cluster.

```

content/docs/installation/upgrading/upgrading-1.6-1.7.md
```diff
@@ -14,6 +14,7 @@ Please read [Migrating Deprecated API Resources] for full instructions.
 If you are currently using HTTP-01 challenges or the Ingress shim annotations, please read the [Ingress class compatibility](./ingress-class-compatibility.md)
 notes to see if your Ingress controller has any known issues with the migration to Ingress v1.
 
+If you have disabled the ServerSideApply feature gate, ensure to enabled them.
```
JoshVanL, line 17:
```suggestion
If running Kubernetes versions before `v1.22`, the [`ServerSideApply`](https://kubernetes.io/docs/reference/using-api/server-side-apply/) feature gate _must_ be enabled. This beta feature is enabled by default on supported versions before `v1.22`.
```

JoshVanL (changes requested):
Thanks @sathyanarays!

I think we should also add that the feature gate is required for the v1.8 and v1.9 upgrade docs for completeness.

content/docs/installation/supported-releases.md
sathyanarays: Done!

sathyanarays:
> Thanks @sathyanarays!
> 
> I think we should also add that the feature gate is required for the v1.8 and v1.9 upgrade docs for completeness.

@JoshVanL made the changes as suggested!

PR #1094, opened 2022-10-18
Venafi: mention why username & password is not supported on TPP  >= 19.2

maelvls:
The cert-manager documentation makes people think that username and password authentication has been dropped entirely in Venafi TPP 19.2. In reality, the username and password will always be supported as an authentication method.

The issue comes from cert-manager itself: cert-manager has not moved yet to the newer `/vedauth/authorize/oauth` because vcert itself doesn't support it yet. vcert [still relies](https://github.com/Venafi/vcert/blob/b89f255872d0253f67902cd9d5f270386a8d9ad6/pkg/venafi/tpp/tpp.go#L343) on the endpoint `/vedsdk/authorize`.

The `/vedsdk/authorize` endpoint itself works fine until TPP 22.1. In TPP 22.2, anyone willing to continue using this endpoint will  have to ask customer service for a license key, which they will then enter in the TPP UI.

I thus propose to remove entirely the phrasing that suggests that username and password is deprecated.

### DIFF
--- a/content/docs/configuration/venafi.md
+++ b/content/docs/configuration/venafi.md
@@ -189,10 +189,10 @@ $ kubectl create secret generic \
 
 #### Username / Password Authentication
 
-NOTE: username / password authentication is deprecated and should only be used when connecting to TPP < 19.2.
-It requires the username and password of a TPP user to be stored in the Kubernetes cluster
-and it does not allow scoped access to the API.
-This means that if these credentials are leaked an attacker may gain long term access to the TPP API and web UI.
+NOTE: cert-manager uses a deprecated TPP endpoint for authenticating with a username
+and password. cert-manager has not been updated to use the newer authentication
+endpoint. Because of this limitation in cert-manager, you should only use username
+and password with to TPP < 19.2.
 
 ```bash
 $ kubectl create secret generic \


maelvls:
@tr1ck3r Hi! Would you be able to review this change? Thanks!

content/docs/configuration/venafi.md
```diff
@@ -137,13 +137,13 @@ In order to set up a Venafi Trust Protection Platform `Issuer`, you must first
 create a Kubernetes `Secret` resource containing your Venafi TPP API
 credentials.
 
-NOTE: For TPP >= 19.2 use Access Token Authentication
-and for older versions of TPP, use username / password authentication.
+NOTE: Starting with TPP 22.2, the username and password authentication will
```
SpectralHiss, line 140: Should note that in 22.3 API Key authentication is completely removed!

But using user/pass doesn't mean necessarily using API Key no? unless that's how it's implemented in cert-manager

SpectralHiss (changes requested):

content/docs/configuration/venafi.md
hawksight: I have the same question as @SpectralHiss. I was under the impression we didn't use API keys and cert-manager used the user / pass to hit the Oauth API.

antstacks: Yeah when I read the initial description here, thats the thought that came to my mind. Saying user/pass doesn't mean you go to the api key. You still use a user pass to hit the oAuth token endpoint, but its the actual auth method that has changed. Simply put the endpoint of api key: "vedsdk/authorize" is no longer accessible, instead the endpoint of Oauth: "vedauth/authorize/oauth" is the only available option in 22.3

wallrj (approved):
@maelvls Let's merge this and then I'll do another PR to add some further clarifications.

/lgtm
/approve

PR #1098, opened 2022-11-02
Rename trust to trust-manager

inteon:
In https://github.com/cert-manager/trust/pull/64 the name trust was changed to trust-manager. This PR renames all occurrences of trust to trust-manager.


wallrj (approved):
Thanks @inteon 

Looks good, but I suppose the `helm install jetstack/trust-manager` instruction won't work until we publish a new release, so I'll add a hold too.

/lgtm
/hold

PR #1100, opened 2022-11-05
Fix links in Getting started with GKE page

foo-dogsquared:
Just noticed these links that goes 404 so I fixed it.

Signed-off-by: Gabriel Arazas <foo.dogsquared@gmail.com>


JoshVanL:
Thanks!

/lgtm
/approve
/ok-to-test

PR #1105, opened 2022-11-08
Fix the heading levels so that they show up in the right-hand menu

wallrj:
## Old

image

## New

image


SgtCoDFish (approved):
/lgtm
/approve

Thanks, good improvement!

PR #1108, opened 2022-11-10
Add a single API Key deprecation block with more information about the affected TPP versons

wallrj:
This builds on @maelvls 's content from #1094 and hopefully makes it clearer that this incompatibility is only related to the way that cert-manager Venafi Issuer does username / password auth and also makes it clearer that the API Key feature is permanently removed in TPP 22.3.

/cc @maelvls

### DIFF
--- a/content/docs/configuration/venafi.md
+++ b/content/docs/configuration/venafi.md
@@ -137,11 +137,6 @@ In order to set up a Venafi Trust Protection Platform `Issuer`, you must first
 create a Kubernetes `Secret` resource containing your Venafi TPP API
 credentials.
 
-NOTE: Starting with TPP 22.2, the username and password authentication will
-require you to ask Venafi's customer support for a license key to continue
-using usernames and passwords in cert-manager. This is due to cert-manager
-relying on the old Web SDK authentication.
-
 ### Access Token Authentication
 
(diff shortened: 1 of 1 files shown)


content/docs/configuration/venafi.md
```diff
@@ -189,11 +184,19 @@ $ kubectl create secret generic \
...
+> ⚠️ When you supply a Venafi TPP username and password,
+> cert-manager uses an older authentication method which is called "API Keys",
+> which has been deprecated since Venafi TPP `19.2`.
+>
+> Beginning in Venafi TPP `22.2`, "API Keys" are disabled by default.
+> You will need to contact Venafi customer support for a special licence key which will allow you to re-enable the "API Keys" feature,
+> so that you can continue to use username and password authentication with cert-manager.
+>
+> In Venafi TPP `22.3`, the "API Keys" feature will be permanently removed,
+> and you will need to use access-token authentication instead.
```
wallrj, line 196: In https://github.com/cert-manager/website/pull/1094#discussion_r1000266914 @SpectralHiss wrote:
> Should note that in 22.3 API Key authentication is completely removed!
> But using user/pass doesn't mean necessarily using API Key no? unless that's how it's implemented in cert-manager


I've tried to address Houssem's comment in this modified version of the warning.

SgtCoDFish (approved):
/lgtm
/approve

Will re-review when spelling checker is placated!

SgtCoDFish:
/lgtm

Thanks for doing this!

PR #1109, opened 2022-11-10
Add twitter / mastodon links

SgtCoDFish:
Includes a temporary banner as a call-to-action, and adds homepage links for both sites.

Also adds an icon for the "view docs" button we have on the homepage!


wallrj (approved):
Thanks.

I clicked the new buttons and they worked for me in Firefox.

/approve
/lgtm

PR #1111, opened 2022-11-15
Release process fixes and clarifications

wallrj:
**Preview**: https://deploy-preview-1111--cert-manager-website.netlify.app/docs/contributing/release-process/#process-for-releasing-a-version

* Remove the `git tag -s` argument, because not all release managers have a suitable gpg key and the gpg requirement is not documented or justified.
* Add a note about optional patch pre-releases, which allow us to provide a patch release to a small number of volunteer community testers before it is released generally.

There has been a discussion about these changes in the following thread in Slack:
 * https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1668521866894379

### DIFF
--- a/content/docs/contributing/release-process.md
+++ b/content/docs/contributing/release-process.md
@@ -138,10 +138,10 @@ page if a step is missing or if it is outdated.
     | initial beta release               | `v1.3.0-beta.0`    |
     | subsequent beta release            | `v1.3.0-beta.1`    |
     | final release                      | `v1.3.0`           |
-    | (optional) patch pre-release[^1]   | `v1.3.1-rc.0`      |
+    | (optional) patch pre-release[^1]   | `v1.3.1-beta.0`    |
     | patch release (or "point release") | `v1.3.1`           |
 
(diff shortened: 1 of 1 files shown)


content/docs/contributing/release-process.md
```diff
@@ -206,7 +209,7 @@ page if a step is missing or if it is outdated.
     3. Create the tag for the new release locally and push it upstream:
 
        ```bash
-       git tag -s -m"v1.8.0-beta.0" v1.8.0-beta.0
+       git tag -m"v1.8.0-beta.0" v1.8.0-beta.0
```
SgtCoDFish, line 212: note: Having `-s` here was a mixture of an oversight and optimism from me.

Oversight: I assumed we'd all have keys set up just for fun.
Optimism: I figured in the future we'd have some way of signing tags using a specific cert-manager key (perhaps using `cmrel`).

I think it's reasonable to remove the requirement to sign tags here, given we sign our actual releases using a cert-manager key which we publish and all signed tags so far have been done with personal developer keys.

SgtCoDFish (commented):
I have a question but as I mention in the review I'd merge as-is so I don't block things needlessly. Thanks for raising this!

SgtCoDFish (approved):
/lgtm
/approve

:+1:

PR #1112, opened 2022-11-16
Add information about OpenShift 4.12

wallrj:
As part of the release of cert-manager v1.11.0-alpha.0 I've checked the supported Kubernetes versions page and noticed the following: 

* We weren't yet mentioning  OpenShift 4.12 (now added)
* [OpenShift 4.7  reached EOL in in august 2022](https://access.redhat.com/support/policy/updates/openshift/#dates), before the release of cert-manager 1.10 so I've removed OpenShift 4.7 from the support table for cert-manager 1.10 and 1.11.

I considered adding Kubernetes 1.26 to the table, since it should be released soon, but perhaps we should only do that once we can test 1.26 with Kind.

### DIFF
--- a/content/docs/installation/supported-releases.md
+++ b/content/docs/installation/supported-releases.md
@@ -17,7 +17,7 @@ and other world events.
 
 | Release  | Release Date |   End of Life   | [Supported Kubernetes versions][s] | [Supported OpenShift versions][s] |
 |----------|:------------:|:---------------:|:----------------------------------:|:---------------------------------:|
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm
/approve

> I considered adding Kubernetes 1.26 to the table, since it should be released soon, but perhaps we should only do that once we can test 1.26 with Kind.

I agree we should do it when it's tested! I don't think it matters much but that'd be my preference.

PR #1117, opened 2022-11-22
Updates cm + Gateway usage docs

irbekrm:
This PR updates Gateway API docs:
- removes a note about TLS blocks in Gateway vs HTTPRoute CR as the TLS block has now been removed from HTTPRoute resources and TLS is only configured via Gateway https://gateway-api.sigs.k8s.io/guides/tls/
- Points users at the Gateway's installation bundles in installation instructions
-  Adds that cert-manager 1.8+ will work with v1beta1 of Gateway API

Once we merge https://github.com/cert-manager/cert-manager/pull/5583 we should also update docs for 1.11 with a note that 1.11 requires that v1beta1 of Gateway API is installed

### DIFF
--- a/content/docs/configuration/acme/http01/README.md
+++ b/content/docs/configuration/acme/http01/README.md
@@ -180,13 +180,13 @@ improvements over the Ingress API.
 
 <div className="info">
 
-📌  This feature requires the installation of the Gateway API CRDs and passing a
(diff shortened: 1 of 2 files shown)


SgtCoDFish (approved):
/lgtm
/approve

Thanks :grin:

PR #1118, opened 2022-11-23
Work around a bug which causes code comments to be parsed as headings

wallrj:
The comments inside fenced code blocks were appearing as headings in the side menu.
Adding a blank line above each fenced code block seems to work around the bug.

image

vs

image


SgtCoDFish (approved):
/lgtm
/approve

Not ideal that we have to do this but seems a simple measure to take to improve things. Thanks!

PR #1122, opened 2022-11-30
Adds a warning about Gateway APi cross-namespace secrets

irbekrm:
Gateway API cross-namespace secrets don't work with gateway-shim see https://github.com/cert-manager/cert-manager/issues/5610

This PR updates docs to add a warning about that.


SgtCoDFish (approved):
/lgtm
/approve

Seems to be an improvement, thanks!

PR #1127, opened 2022-12-05
change stringData to data

cdenneen:
Fixes https://github.com/cert-manager/cert-manager/issues/5617

Signed-off-by: Chris Denneen <cdenneen@gmail.com>

### DIFF
--- a/content/docs/configuration/acme/dns01/akamai.md
+++ b/content/docs/configuration/acme/dns01/akamai.md
@@ -17,7 +17,7 @@ kind: Secret
 metadata:
   name: akamai-secret
 type: Opaque
-stringData:
+data:
   clientSecret: use_akamai_client_secret
   accessToken: use_akamai_access_token
   clientToken: use_akamai_client_token


content/docs/configuration/acme/dns01/akamai.md
```diff
@@ -17,7 +17,7 @@ kind: Secret
 metadata:
   name: akamai-secret
 type: Opaque
-stringData:
+data:
```
wallrj, line 20: I don't agree with this change. 
`stringData` is the correct way to supply these credentials when creating the Secret because it otherwise the user would have to manually base64 encode the values.
See:

 * https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data

wallrj (changes requested):

content/docs/configuration/acme/dns01/akamai.md
cdenneen: Ok makes sense

PR #1130, opened 2022-12-14
Expand Google Workload Identity section with a note about Ambient Credentials

ar-qun:
Signed-off-by: Alex <17101722+ar-qun@users.noreply.github.com>

At least for me, the Workload Identity didn't start working until Ambient Credentials where enabled. I added a small section to cover that dependency.

### DIFF
--- a/content/docs/configuration/acme/dns01/google.md
+++ b/content/docs/configuration/acme/dns01/google.md
@@ -149,6 +149,10 @@ detailed instructions on how to enable workload identity in your GKE cluster.
 The instructions in the following sections assume you are deploying cert-manager
 to a GKE cluster with workload identity already enabled.
 
+### Enable Ambient Credential Usage
+
+'Ambient Credentials' are credentials drawn from the environment, metadata services, or local files which are not explicitly configured in the ClusterIssuer API object. When this flag is enabled Cert-Manager will access the GKE Metadata server for credentials. By default this is enabled for ClusterIssuer resources but is disabled for Issuer resources. To enable it for Issuer resources set the `--issuer-ambient-credentials` flag.
+
 ### Link KSA to GSA in GCP
 
 The cert-manager component that needs to modify DNS records is the pod created


content/docs/configuration/acme/dns01/google.md
```diff
@@ -149,6 +149,10 @@ detailed instructions on how to enable workload identity in your GKE cluster.
 The instructions in the following sections assume you are deploying cert-manager
 to a GKE cluster with workload identity already enabled.
 
+### Enable Ambient Credential Usage
+
+'Ambient Credentials' are credentials drawn from the environment, metadata services, or local files which are not explicitly configured in the ClusterIssuer API object. When this flag is enabled Cert-Manager will access the GKE Metadata server for credentials. By default this is enabled for ClusterIssuer resources but is disabled for Issuer resources. To enable it for Issuer resources set the `--issuer-ambient-credentials` flag.
```
wallrj, line 154: Thanks for adding this extra context.
In a future PR we should add a much clearer explanation of the security implications of enabling ambient credentials for Issuer resources.

The Azure DNS docs already have something which we should repeat everywhere ambient credentials are used.

 https://release-next--cert-manager-website.netlify.app/docs/configuration/acme/dns01/azuredns/#%EF%B8%8F-using-ambient-credentials-with-clusterissuer-and-issuer-resources

> :warning: This authentication method is an example of what cert-manager calls 'ambient credentials'. Ambient credentials are enabled by default for ClusterIssuer resources, but disabled by default for Issuer resources. This is to prevent unprivileged users, who have permission to create Issuer resources, from issuing certificates using credentials that cert-manager incidentally has access to. ClusterIssuer resources are cluster scoped (not namespaced) and only platform administrators should be granted permission to create them.
> 
> It is possible (but not recommended) to enable this authentication mechanism for Issuer resources, by setting the --issuer-ambient-credentials flag on the cert-manager controller to true.
>

wallrj (approved):
/ok-to-test
/approve
/lgtm

PR #1131, opened 2022-12-15
Update supported releases with latest data

SgtCoDFish:
Adds k8s 1.26 as discussed in standup on 2022-12-15

Also: 

- Removes older k8s versions from future releases
- Adds predicted support + release dates for cert-manager 1.12
- Updates cloud vendor supported versions list


irbekrm (approved):
Thanks @SgtCoDFish !

/lgtm

PR #1133, opened 2022-12-21
Various updates + tweaks

SgtCoDFish:
- Fixes the version selector to sort descending by release number rather than the current sorting which is ascending and is broken in any case. Pictures below.
- NPM told me to update caniuse-lite, so I did
- Fixes a currently broken link on supported releases
- Adds 1.10 release notes to next docs (this was missed previously. Doesn't matter since next-docs aren't appearing on the site ATM anyway and the manifest for next docs is broken since it refers to `docs` and not `next-docs` - this is just a cleanup thing)

:warning: Also adds some docs to next-docs for the new ACME caBundle support!

## Old Version Picker

Old version picker

## New Version Picker

New version picker

### DIFF
--- a/components/docs/VersionSelect.jsx
+++ b/components/docs/VersionSelect.jsx
@@ -3,6 +3,8 @@ import { useState } from 'react'
 import SidebarLink from './Sidebar/SidebarLink'
 import Icon from 'components/Icon'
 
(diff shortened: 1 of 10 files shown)


wallrj (approved):
/lgtm

PR #1135, opened 2023-01-06
Link to next release docs

wallrj:
I've added a link to the Netlify preview of the release-next branch so that contributors can more easily see the results of merging their next release documentation.

The process will be that we get all release notes, upgrade notes, installation links etc updated in the release-next branch  before we do the final release and then merge that branch into master after we have done the final release.

We might even update the release-next branch with installation versions for each beta release, so that users can more easily do pre-release testing.

I assume (hope) that the netlify preview site will not appear search engine listings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ There are several folders in `content/` and which one you need depends on what y
 - Something which applies to the current version of cert-manager? <br />
   Add it to `docs/` and possibly to the specific version of cert-manager that's latest (e.g. `v1.8-docs/`)
 - Something which only applies to the next major version of cert-manager? <br />
-  Add it to `next-docs`
+  Add it to `docs/` but branch from the [`release-next` branch](https://github.com/cert-manager/website/tree/release-next) and merge the PR into that branch.
 - Something which isn't "versioned", e.g. a page under "contributing", release notes or our supported-releases page? <br />
   Add it to `docs/`, which is the only place such pages should appear
 - Something which applies only to versions of cert-manager which have already been released? <br />


SgtCoDFish:
I really like the idea of linking to the Netlify preview! Simple solution to the problem. Traffic levels to next docs should presumably be low in any case and it should be easy to keep up to date, in theory.

> I assume (hope) that the netlify preview site will not appear search engine listings.

This is the only real blocker I might have with this PR - I'm not sure if this is true. I'm not really familiar enough with SEO to say confidently, but I think that the site _will_ appear in search engine listings if we link to it like this.

Looking at https://release-next--cert-manager-website.netlify.app/robots.txt it looks like the site _would_ be indexed.

Also, the HTML on the index page + docs page of the preview explicitly asks to be indexed:

```html
<head>
...
<meta name="robots" content="index,follow"/>
...
</head>
```

The response headers indicate that the page should not be cached, though:

>  cache-control: public, max-age=0, must-revalidate

(I also wonder if this would be a case for [`rel=nofollow`](https://en.wikipedia.org/wiki/Nofollow), but again I don't really know how to be confident about that)

I feel like if we're going to rely on a preview in "production" like this, then it would be good to be sure that it's [not crawled](https://developers.google.com/search/docs/crawling-indexing/block-indexing). But I imagine it'll be a pain to get a different robots.txt / meta tag served on a preview.

Long story short: I'm concerned that this will be indexed unless I'm missing something that'll stop that. I don't think we'd want it to be indexed. I'm happy to merge anyway I think, but I'll add a hold in case you want to investigate further.

/approve
/lgtm
/hold

wallrj:
The deploy-preview sites already have the noindex header, but branch builds do not, so I'll try and add it according to these instructions:

 * https://docs.netlify.com/routing/headers/#custom-headers-for-different-branch-or-deploy-contexts

* deploy-preview
```terminal

$ curl --head -i https://deploy-preview-1135--cert-manager-website.netlify.app/
HTTP/2 200 
age: 0
cache-control: public, max-age=0, must-revalidate
content-type: text/html; charset=UTF-8
date: Fri, 06 Jan 2023 15:11:05 GMT
etag: "9768dd9514ea9f1665c3ea8ac7127bd7-ssl"
server: Netlify
strict-transport-security: max-age=31536000; includeSubDomains; preload
x-nf-request-id: 01GP3T5CRCGCDV30JHB5SZPCNM
x-robots-tag: noindex
content-length: 90955
```

vs

* branch-deploy
```
$ curl --head -i https://release-next--cert-manager-website.netlify.app 
HTTP/2 200 
age: 52
cache-control: public, max-age=0, must-revalidate
content-type: text/html; charset=UTF-8
date: Fri, 06 Jan 2023 15:09:49 GMT
etag: "0183a3a8da5a1b1c8658bedb256a6597-ssl"
server: Netlify
strict-transport-security: max-age=31536000; includeSubDomains; preload
x-nf-request-id: 01GP3T4P29Q622KMH7CFR8634Q
content-length: 90955

```

wallrj:
Thanks for reviewing that branch. The noindex headers are now appearing, so I think it's safe to merge this PR now.

```
$ curl --head -i https://release-next--cert-manager-website.netlify.app 
HTTP/2 200 
age: 0
cache-control: public, max-age=0, must-revalidate
content-type: text/html; charset=UTF-8
date: Fri, 06 Jan 2023 16:04:41 GMT
etag: "3db0e5296370dec005cb3901ac3ba678-ssl"
server: Netlify
strict-transport-security: max-age=31536000; includeSubDomains; preload
x-nf-request-id: 01GP3X7HTFCFRZDTK2H4VXZ1PQ
x-robots-tag: noindex
content-length: 90955

```

/hold cancel

PR #1145, opened 2023-01-11
Remove next-docs and add redirect

SgtCoDFish:
@wallrj made some great changes whereby we use the release-next branch to add changes which should only appear in the next version of our site. This is a huge improvement over the "next-docs" system we previously had, which always got out of date.

This PR removes the next-docs folder and any references to it, and also adds a blanket redirect from any old next-docs pages (which are currently broken anyway on the live site) to the new release-next branch preview.


wallrj (approved):
/lgtm

PR #1102, opened 2022-11-06
relativelyrehan: outside click for navigation menu

relativelyrehan:
- Add file useOutsideClick.js for handling outside click on the navigation menu on mobile screen
- FIxed [this](https://github.com/cert-manager/website/issues/1082)

https://user-images.githubusercontent.com/48210765/200192517-a5985d08-6ab6-475f-9946-8f265b08d250.mov

PR #1152, opened 2023-01-12
add outside alert handler

relativelyrehan:
Signed-off-by: relativelyrehan <rehan18alam@gmail.com>

@wallrj Hi, I was getting some signoff related issue in the old PR #1102 and hence closed that PR, this will contain the same code with Signoff update.

Ping, if you face any issues, or want any change

### DIFF
--- a/lib/useOutsideClick.js
+++ b/lib/useOutsideClick.js
@@ -1,0 +1,27 @@
+import { useEffect } from 'react'
+
+/*
+  Handles outside click events on the menu component
+  This function is used to close the menu when a user clicks outside of it
+*/
+
+function useOutsideAlerter(ref, setShow = null, callBack = null) {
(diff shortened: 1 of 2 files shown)


lib/useOutsideClick.js
```diff
@@ -0,0 +1,27 @@
+import { useEffect } from 'react'
+
+/*
+  Handles outside click events on the menu component
+  This function is used to close the menu when a user clicks outside of it
+*/
```
wallrj, line 6: Thanks for that explanation.

wallrj (approved):
Thanks @relativelyrehan  

I tested it using Firefox responsive design mode and observed that the main menu now disappears  when I *click* outside it.

If there are any other improvements we can make to the site, please me know because I'm not a react / nextjs expert.

Issue #974, opened 2022-05-06
Investigate styled 404 page


SgtCoDFish:
Suggested by @wallrj

We used to have a broken but styled 404 page for the site. Now we have a default netlify page. We should get a working 404 styled page!

prafull904434:
sir,may i work with this issue can you assign to me

SgtCoDFish:
Sorry @prafull904434 it looks like this ticket was actually addresed by #1153 ! I don't think there's anything to do here so I'll close this. Thanks!

PR #1153, opened 2023-01-12
add 404 page

relativelyrehan:
Signed-off-by: relativelyrehan <rehan18alam@gmail.com>

Solve #974 

This 404 page can be accessed by going to any unavailable route

### Desktop

Screenshot 2023-01-13 at 1 53 40 AM

### Mobile
Screenshot 2023-01-13 at 1 55 18 AM


wallrj (changes requested):
Thanks @relativelyrehan !

It looks great, but the 404 in the 404 image seems to be cropped. 

And do you think we need to make some other change to make this work for clients without Javascript. 

When I visit https://deploy-preview-1153--cert-manager-website.netlify.app/docs/install/verify/ ([a known broken link)](https://github.com/cert-manager/website/pull/1154) it displays the Netlify default 404 page instead.

wallrj:
See https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling

relativelyrehan:
@wallrj that 404 image has that design, in which the 404 seems to be cropped. That is intentional design to indicate the idea of a broken page. 

For users without JS, I think it's not an issue as on build all of this will be a static HTML with image url and homepage url, both of them are static sources.

wallrj:
> @wallrj that 404 image has that design, in which the 404 seems to be cropped. That is intentional design to indicate the idea of a broken page.

It's an interesting idea, but my first reaction was that the image is broken. 
Please use a non cropped image and then I'll be happy to merge this.

relativelyrehan:
@wallrj I will have a look at the documentation for netlify, because there it doesn't seems to work.

I have one question though, are we using netlify for the production deploy too?

wallrj:
It might be worth investigating that in a separate PR though. What you've got in this branch is nice and self contained and is a clear improvement which *I think* should work for people who have JavaScript enabled and who are simply clicking around the site.

pages/404.jsx
```diff
@@ -0,0 +1,23 @@
...
+                src={'/images/cert-manager-404-illustration.svg'}
+                height={360}
+                width={625}
+                alt="404 - Not found"
+            />
+            <h1 className="text-2xl mb-5 lg:mb-4 lg:text-4xl max-w-3xl">Oops, We can&apos;t seem to find
+                the page you are looking for.</h1>
+            <p className="text-base mb-5 lg:mb-14 lg:text-xl max-w-3xl">
+                The page you are looking for might have been removed, had its name changed, or is temporarily unavailable.
+            </p>
```
SgtCoDFish, line 17: suggestion:

```suggestion
            <p className="text-base mb-5 lg:mb-14 lg:text-xl max-w-3xl">
                The page you are looking for might have been removed, had its name changed, or be temporarily unavailable.
            </p>
```

wallrj (approved):
@relativelyrehan It's been pointed out to me that I probably shouldn't be nitpicking on your 404 logo design. This is a great improvement over the default Netlify page.  
Thanks for sorting it out.

/approve
/lgtm

relativelyrehan:
@wallrj no problem. Thanks for the merge. I also wanted to check with you if there is any open position for a partime paid frontend developer in your org.

PR #1154, opened 2023-01-13
fix: dir name is installation not install

jackroberts-gh:
Links are broken in the docs due to linking to "install" dir rather than "installation".


wallrj (approved):
Thanks for spotting that! We ought to run a link checker in our CI :roll_eyes: 

/ok-to-test
/approve
/lgtm

PR #1156, opened 2023-01-13
Upgrade all JavaScript dependencies

wallrj:
**Preview**: https://deploy-preview-1156--cert-manager-website.netlify.app/

I've upgraded to nextjs 13 and also upgraded all the other dependencies in `package.json`.
This required some (mostly machine generated) changes to the code.

See comments below for more context.


components/Image.jsx
```diff
@@ -1,4 +1,4 @@
-import Image from 'next/image'
+import Image from "next/legacy/image";
```
wallrj, line 1: This import was changed automatically by `npx @next/codemod next-image-to-legacy-image .`
    
See https://nextjs.org/docs/messages/next-image-upgrade-to-13

components/docs/VersionSelect.jsx
```diff
@@ -6,11 +6,9 @@ import Icon from 'components/Icon'
...
 
 function labelFromVersion(version) {
-    return (
-        version === 'docs'
-            ? 'latest'
-            : version.replace(/-docs$/, '').replace(/^v/, '')
-    );
+    return version === 'docs'
+        ? 'latest'
+        : version.replace(/-docs$/, '').replace(/^v/, '');
```
wallrj, line 11: This is a side effect of running `npx @next/codemod new-link .` but it seems innocuous.

lib/remark-plugins/heading-ids/index.js
```diff
@@ -31,5 +31,5 @@ export default function headingIds() {
         }
       }
     })
-  }
+  };
```
wallrj, line 34: Another side effect of one of the code rewriters (above)

next.config.js
```diff
@@ -34,5 +31,8 @@ module.exports = {
   },
   eslint: {
     ignoreDuringBuilds: true
+  },
+  images: {
+    unoptimized: true
   }
```
wallrj, line 37: The configuration we had [was a work around for `next export`](https://github.com/vercel/next.js/issues/27231#issuecomment-888414710), which no longer works with next 13

So I removed it which results in us using the default image loading mechanism. See
    
    https://nextjs.org/docs/messages/next-image-missing-loader

But that is not compatible with next export either, so I added the `unoptimized: true` configuration in response to the cli error message:

```
> next export

info  - Loaded env from /home/richard/projects/cert-manager/website/.env
info  - using build directory: /home/richard/projects/cert-manager/website/.next
info  - Copying "static build" directory
info  - No "exportPathMap" found in "/home/richard/projects/cert-manager/website/next.config.js". Generating map from "./pages"
Error: Image Optimization using Next.js' default loader is not compatible with `next export`.
  Possible solutions:
    - Use `next start` to run a server, which includes the Image Optimization API.
    - Configure `images.unoptimized = true` in `next.config.js` to disable the Image Optimization API.
  Read more: https://nextjs.org/docs/messages/export-image-api
    at /home/richard/projects/cert-manager/website/node_modules/next/dist/export/index.js:153:23
    at async Span.traceAsyncFn (/home/richard/projects/cert-manager/website/node_modules/next/dist/trace/trace.js:79:20)


```

package.json
```diff
@@ -54,55 +54,55 @@
...
+    "prettier": "2.8.2",
+    "remark-cli": "11.0.0",
+    "remark-lint": "9.1.1",
+    "remark-preset-lint-consistent": "5.1.1",
+    "remark-preset-lint-recommended": "6.1.2",
+    "remark-toc": "8.0.1",
+    "sass": "1.57.1",
+    "stylelint": "14.16.1",
+    "stylelint-config-standard": "29.0.0",
+    "windicss-webpack-plugin": "1.7.7"
```
wallrj, line 106: The dependency changes in this file were  mostly performed using `ncu -u --removeRange` as recommended in https://stackoverflow.com/a/16074029

But then I manually removed the `^` prefix from some of the untouched dependencies, for consistency.

package.json
```diff
@@ -54,55 +54,55 @@
...
-    "sass": "^1.50.0",
-    "unist-util-visit": "^4.1.0"
+    "@docsearch/react": "3.3.2",
+    "@headlessui/react": "1.7.7",
+    "@next/mdx": "13.1.2",
+    "classnames": "2.3.2",
+    "compare-versions": "6.0.0-rc.1",
+    "gray-matter": "4.0.3",
+    "next": "13.1.2",
+    "next-mdx-remote": "4.2.0",
```
wallrj, line 64: This particular dependency was initially upgraded to 4.2.1 by ncu, but that wasn't compatible with next 13 so I downgraded it to 4.2.0 . See:
 * https://github.com/hashicorp/next-mdx-remote/issues/307#issuecomment-1376580829

wallrj (commented):
In addition to the remarks below, I ran `npm shrinkwrap` which resulted in the renaming of [package-lock.json → npm-shrinkwrap.json](https://github.com/cert-manager/website/pull/1156/files#diff-44a7c69871c5958e657226d5552c9451606a778233fb824f68530d0cfd3f8994)

This seems to be best practice for npm applications (versus libraries) but I may have misunderstood:
 * https://docs.npmjs.com/cli/v9/configuring-npm/npm-shrinkwrap-json/

inteon:
> In addition to the remarks below, I ran `npm shrinkwrap` which resulted in the renaming of [package-lock.json → npm-shrinkwrap.json](https://github.com/cert-manager/website/pull/1156/files#diff-44a7c69871c5958e657226d5552c9451606a778233fb824f68530d0cfd3f8994)
> 
> This seems to be best practice for npm applications (versus libraries) but I may have misunderstood:
> 
> * https://docs.npmjs.com/cli/v9/configuring-npm/npm-shrinkwrap-json/

I think npm-shrinkwrap.json is only useful when you are talking about libraries:  
https://medium.com/@hossam.hilal0/package-json-vs-package-lock-json-vs-npm-shrinkwrap-json-33fcddc1521a
```
Thus:
If you are not publishing your package to npm, the choice between these two files
is of little consequence. You may wish to use package-lock.json because it is the default
and its name is clearer to npm beginners; alternatively, you may wish to use npm-shrinkwrap.json
for backwards compatibility with npm 2-4 if it is difficult for you to ensure everyone
on your development team is on npm 5+. (Note that npm 5 was released on 25th May 2017;
backwards compatibility will become less and less important the further we get from that
date, as most people will eventually upgrade.)
```

components/Image.jsx
inteon: can we run npx `@next/codemod next-image-experimental .` too?

components/Button.jsx
```diff
@@ -16,14 +16,14 @@ export default function Button({
...
+    (<Link
+      href={href}
+      target={target || '_self'}
+      className={`font-montserrat font-bold uppercase text-sm leading-20px btn-primary rounded-5px ${styles} ${className}`}>
+
+      <span className={"block " + iconWClass}>{icon && <Icon name={icon} />}</span>
+      <span>{caption}</span>
+
+    </Link>)
+  );
```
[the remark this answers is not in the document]
inteon: we probably can remove these extra `(` and `)` here.

components/Image.jsx
wallrj: No, someone can do that in another PR. The documentation says:

> Please note this second codemod is experimental and only covers static usage (such as <Image src={img} layout="responsive" />) but not dynamic usage (such as <Image {...props} />).

wallrj:
> I think npm-shrinkwrap.json is only useful when you are talking about libraries:
> https://medium.com/@hossam.hilal0/package-json-vs-package-lock-json-vs-npm-shrinkwrap-json-33fcddc1521a

Agreed. I've revered that change.

PTAL

inteon:
Thank you for upgrading!
/lgtm
/approve

PR #1160, opened 2023-01-16
Wide-ranging updates to trust-manager docs for v0.4.0

SgtCoDFish:
Includes an updated tutorial, defaultCAs docs, details about avoiding gotchas, and more!

Should only be merged after trust-manager v0.4.0 is released

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -32,6 +32,7 @@ ajvn
 illrill
 davidsbond
 fvlaicu
+facto
 4molybdenum2
 jetstack-bot
 codegen
@@ -89,6 +90,7 @@ CSR
 CSRs
 CVE
 CVEs
+Certbot
 CertificateRequest
 CertificateRequestPolicy
 CertificateRequestPolicies
--- a/content/docs/projects/trust-manager/README.md
+++ b/content/docs/projects/trust-manager/README.md
@@ -269,6 +269,8 @@ defaultPackageImage:
   tag: XYZ
 ```
 
+These versions of the default package image tags are derived directly from the version of the `ca-certificates` package in Debian.
+
 Finally, apply back the changes, being sure to manually specify the version of trust-manager which is installed, to avoid
 also updating the trust-manager controller at the same as the default CA package:
 
@@ -283,11 +285,14 @@ echo $TRUST_MANAGER_VER
 helm upgrade -f values.yaml -n cert-manager trust-manager jetstack/trust-manager --version $TRUST_MANAGER_VER
 ```
 
+If an incorrect tag is used, your deployment will fail and you'll likely need to use `helm rollback` to get back
+to a working state.
+
 ## Preparing for Production
 
-TLS is a complicated beast and there are plenty of ways to use TLS certificates in ways which will cause problems
-down the line. We'd like to document a few of these potential gotchas here to help prepare users for running
-trust-manager safely in production.
+TLS can be complicated and there are many ways to misuse TLS certificates.
+
+Here are some potential gotchas here to be aware of before running trust-manager in production.
 
 If you're planning on running trust-manager in production and you're using more than just the default CA package,
 we **strongly** advise you to read and understand this section. It could save you from causing an outage later.
--- a/content/docs/projects/trust-manager/README.md
+++ b/content/docs/projects/trust-manager/README.md
@@ -3,44 +3,64 @@ title: trust-manager
 description: 'Distributing Trust Bundles in Kubernetes'
 ---
 
-## Distributing Trust Bundles in Kubernetes
+trust-manager is the easiest way to manage trust bundles in Kubernetes and OpenShift clusters.
 
-trust-manager is an operator for distributing trust bundles across a Kubernetes cluster.
-trust-manager is designed to complement
-[cert-manager](https://github.com/cert-manager/cert-manager) by enabling services to
-trust X.509 certificates signed by Issuers, as well as external CAs which may
-not be known to cert-manager at all.
+It orchestrates bundles of trusted X.509 certificates which are primarily used for validating
+certificates during a TLS handshake but can be used in other situations, too.
+
+## Overview
+
+trust-manager is a small Kubernetes operator which aims to help reduce the overhead of managing
+TLS trust bundles in your clusters.
+
+It adds the `Bundle` custom Kubernetes resource (CRD) which can read input from various sources
+and combine the resultant certificates into a bundle ready to be used by your applications.
+
+trust-manager ensures that it's both quick and easy to keep your trusted certificates up-to-date
+and enables cluster administrators to easily automate providing a secure bundle without having
+to worry about rebuilding containers to update trust stores.
+
+It's designed to complement cert-manager and works well when consuming CA certificates from a
+cert-manager `Issuer` or `ClusterIssuer` but can be used entirely independently from cert-manager
+if needed.
 
 ## Usage
 
-trust ships with a single cluster scoped `Bundle` resource. A Bundle represents
-a set of data that should be distributed and made available across the cluster.
-There are no constraints on what data can be distributed.
+trust-manager is intentionally simple, and adds one new Kubernetes `CustomResourceDefintion`: `Bundle`.
 
-The Bundle gathers and appends trust data from a number of `sources` located in
-the trust namespace (where the trust controller is deployed), and syncs them to
-a `target` in every namespace.
+A `Bundle` represents a set of PEM-encoded X.509 certificates that should be distributed and made
+available across the cluster. `Bundle`s are cluster scoped.
 
-A typical Bundle looks like the following:
+Users specify a list of `sources`, which trust-manager will query and concatenate certificate data from.
+The only other required field is the `target`, which describes how and where the resulting bundle will
+be written.
+
+An example `Bundle` might look like this:
 
 ```yaml
 apiVersion: trust.cert-manager.io/v1alpha1
 kind: Bundle
 metadata:
-  name: my-org.com
+  name: my-org.com  # The bundle name will also be used for the target
 spec:
   sources:
-  # A Secret in the trust namespace created via a cert-manager Certificate
+  # Include a bundle of publicly trusted certificates which can be
+  # used to validate most TLS certificates on the internet, such as
+  # those issued by Let's Encrypt, Google, Amazon and others.
+  - useDefaultCAs: true
+
+  # A Secret in the trust-manager namespace
   - secret:
       name: "my-db-tls"
       key: "ca.crt"
-  # A ConfigMap in the trust namespace
+
+  # A ConfigMap in the trust-manager namespace
   - configMap:
       name: "my-org.net"
       key: "root-certs.pem"
-  # An In Line
+
+  # A manually specified string
   - inLine: |
-      # my-org.com CA
       -----BEGIN CERTIFICATE-----
       MIIC5zCCAc+gAwIBAgIBADANBgkqhkiG9w0BAQsFADAVMRMwEQYDVQQDEwprdWJl
       ....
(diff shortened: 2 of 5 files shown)


.spelling
```diff
@@ -32,6 +32,7 @@ ajvn
 illrill
 davidsbond
 fvlaicu
+facto
```
SgtCoDFish, line 35: note: as in "de facto".

SgtCoDFish:
/hold

Only to be merged around the release of trust-manager v0.4.0

content/docs/projects/trust-manager.md
```diff
@@ -3,44 +3,58 @@ title: trust-manager
 description: 'Distributing Trust Bundles in Kubernetes'
 ---
 
-## Distributing Trust Bundles in Kubernetes
+trust-manager is the easiest way to manage TLS trust bundles in Kubernetes and OpenShift clusters.
```
inteon, line 6: `TLS trust bundles` these trusted `X.509` certs are not only for TLS right, you could also use them for document signing etc.?

content/docs/projects/trust-manager.md
```diff
@@ -3,44 +3,58 @@ title: trust-manager
...
+It's designed to complement cert-manager and works well when consuming CA certificates from a
+cert-manager `Issuer` or `ClusterIssuer` but can be used entirely independently from cert-manager
+if needed.
 
 ## Usage
 
-trust ships with a single cluster scoped `Bundle` resource. A Bundle represents
-a set of data that should be distributed and made available across the cluster.
-There are no constraints on what data can be distributed.
+trust-manager ships with a single cluster scoped `Bundle` resource. A `Bundle` represents
```
inteon, line 26: `a single cluster scoped Bundle resource` -> this confuses me. you can have multiple resources, it just ships with one CRD or one resource type right?

content/docs/projects/trust-manager.md
```diff
@@ -3,44 +3,58 @@ title: trust-manager
...
+a set of PEM-encoded certificates that should be distributed and made available across the cluster.
 
-The Bundle gathers and appends trust data from a number of `sources` located in
-the trust namespace (where the trust controller is deployed), and syncs them to
-a `target` in every namespace.
+The `Bundle` gathers and concatenates certificate data from a number of `sources` located in
```
inteon, line 29: Here 'The `Bundle`' also confuses me, see comment above.

content/docs/projects/trust-manager.md
SgtCoDFish: Yeah that's true - I'm trying to get the keyword in there because "trust bundles" feels incredibly abstract and the number of people who'll use this for anything other than TLS is probably a rounding error.

X.509 is definitely (currently) more accurate... but maybe I'll just change to "trust bundles" for now so we don't have to change it later if we add e.g. JWK support

content/docs/projects/trust-manager.md
SgtCoDFish: Agree, the wording is clumsy and confusing and needs more than just the update from trust -> trust-manager. I'll change.

content/docs/projects/trust-manager.md
```diff
@@ -56,79 +70,299 @@ spec:
...
 ```
 
-Bundle currently supports the source types `configMap`, `secret` and `inLine`,
-and target type `configMap`.
+`Bundle` resources currently support several source types:
+
+- `configMap` - a `ConfigMap` resource in the trust-manager namespace
+- `secret` - a `Secret` resource in the trust-manager namespace
+- `inLine` - a manually specified string containing at least one certificate
+- `useDefaultCAs` - usually, a bundle of publicly trusted certificates
```
inteon, line 78: can we be a bit more specific here and explain where more info can be found?

SgtCoDFish: Yeah that'd be an improvement; I think realistically, we should have API reference docs for trust-manager like we have for cert-manager because honestly at this point I don't really think we have a good place where "more info can be found" :joy:

I'll take a quick stab at generating API docs - maybe it's achievable for this PR.

content/docs/projects/trust-manager.md
```diff
@@ -56,79 +70,299 @@ spec:
...
+```bash
+git clone --single-branch https://github.com/cert-manager/trust-manager trust-manager
+cd trust-manager
+make demo
+# kubeconfig is in ./bin/kubeconfig.yaml
+# kind cluster is called "trust"
+```
+
+The demo installation uses Helm, and roughly matches what you'd get by installing trust-manager into your own
+cluster using Helm - although it uses locally-built images rather than the ones we provide publicly.
```
inteon, line 113: Do we want to explain this here or do we move this to the README.md file in the trust-manager project.

content/docs/projects/trust-manager.md
```diff
@@ -56,79 +70,299 @@ spec:
...
+TRUST_MANAGER_VER=$(helm list --filter "^trust-manager$" -n cert-manager -ojson | jq -r ".[0].app_version")
+
+# Check the version makes sense
+echo $TRUST_MANAGER_VER
+
+# Run the upgrade
+helm upgrade -f values.yaml -n cert-manager trust-manager jetstack/trust-manager --version $TRUST_MANAGER_VER
 ```
+
+## Avoiding Gotchas
```
inteon, line 297: Can we make this a more important section, like "read this before starting" or "Read this before running in production"? To me "Gotchas" seems like something I won't read because I know what I'm doing.

content/docs/projects/trust-manager.md
```diff
@@ -56,79 +70,299 @@ spec:
...
+down the line. We'd like to document a few of these potential gotchas here.
+
+If you're planning on running trust-manager in production and you're using more than just the default CA package,
+we **strongly** advise you to read and understand this section. It could save you from causing an outage later.
+
+> ℹ️ These gotchas aren't specific to trust-manager and you could run into any of them with any method of managing TLS trust!
+
+### Bundling Intermediates
+
+A very common anti-pattern in TLS is that servers fail to send their chain along with their leaf certificate. The correct
```
inteon, line 309: can we add some references here?

content/docs/projects/trust-manager.md
```diff
@@ -56,79 +70,299 @@ spec:
...
+fix for this is that the server should be updated to start sending its chain, but often a fix that _seems_ to work is
+for the client to add the intermediate to its trust store.
+
+This is dangerous; it means that the intermediate cannot be safely rotated without all trust stores which contain
```
inteon, line 313: can we add some icons/ bold text here?

inteon: and make sure it is very clear that this is a serious problem.

content/docs/projects/trust-manager.md
```diff
@@ -56,79 +70,299 @@ spec:
...
 ```
 
-#### Quick Start Example
+### Manual Installation
+
+We strongly recommend that you install trust-manager using Helm and we don't currently support manually installed
+versions of trust-manager. This is so that we can focus on continuing to improve trust-manager with the resources
+we currently have available.
+
+A longer term goal of ours is to add first-class support for installation without requiring Helm or cert-manager.
```
inteon, line 135: can we remove this sentence, we probably don't want to add this in the docs. I think this will only lead to users expecting more and waiting for this to be completed. If we don't add this, they might be more inclined to contribute something like this themselves.

content/docs/projects/trust-manager.md
SgtCoDFish: I rewrote this; my "references" would be "bitter personal experience" so I don't really have any useful links.

content/docs/projects/trust-manager.md
SgtCoDFish: I think there's value to having this here as sort of a "quick and easy way to experiment with trust-manager" - but agreed it's probs more useful in the README!

I raised https://github.com/cert-manager/trust-manager/pull/103

content/docs/projects/trust-manager.md
SgtCoDFish: I've tried rewriting a little, what do you think?

content/docs/projects/trust-manager.md
inteon: Thx, I approved that PR.

content/docs/projects/trust-manager/README.md
```diff
@@ -56,79 +76,293 @@ spec:
...
-default              example-bundle   1      2m18s
-kube-node-lease      example-bundle   1      2m18s
-kube-public          example-bundle   1      2m18s
-kube-system          example-bundle   1      2m18s
-local-path-storage   example-bundle   1      2m18s
+helm get values -n cert-manager trust-manager -oyaml > values.yaml
+```
+
+Next, if `defaultPackageImage.tag` is already set in `values.yaml`, update it. Otherwise, add it.
+You can find the available tags [on `quay.io`](https://quay.io/repository/jetstack/cert-manager-package-debian?tab=tags&tag=latest).
```
wallrj, line 263: I looked at that page and was confused about why the tag is a date in 2021 but the  image appears to be last modified only 18 days ago.

image


I suggest explaining the significance of the date in these tags.

content/docs/projects/trust-manager/README.md
```diff
@@ -56,79 +76,293 @@ spec:
...
-kubectl get cm -A --field-selector=metadata.name=example-bundle
-NAMESPACE            NAME             DATA   AGE
-cert-manager         example-bundle   1      2m18s
-default              example-bundle   1      2m18s
-kube-node-lease      example-bundle   1      2m18s
-kube-public          example-bundle   1      2m18s
-kube-system          example-bundle   1      2m18s
-local-path-storage   example-bundle   1      2m18s
+helm get values -n cert-manager trust-manager -oyaml > values.yaml
+```
```
wallrj, line 260: The content of `values.yaml` will be `null` if the chart was originally installed without any custom values.
Maybe add a note about that.

It might be simpler to pass the `--reuse-values` flag instead and supply `--set defaultPackageImage.tag=XYZ` to override the last such value if it was supplied.

content/docs/projects/trust-manager/README.md
```diff
@@ -56,79 +76,293 @@ spec:
...
-ABC
-hello world!
+# Get the currently installed version. You could do this manually if you find that easier.
+TRUST_MANAGER_VER=$(helm list --filter "^trust-manager$" -n cert-manager -ojson | jq -r ".[0].app_version")
+
+# Check the version makes sense
+echo $TRUST_MANAGER_VER
+
+# Run the upgrade
+helm upgrade -f values.yaml -n cert-manager trust-manager jetstack/trust-manager --version $TRUST_MANAGER_VER
```
wallrj, line 283: Maybe add a note to say that if you specify an incorrect `defaultPackageImage.tag`, the new deployment will fail, and you will need to use `helm rollback`.

content/docs/projects/trust-manager/README.md
```diff
@@ -56,79 +76,293 @@ spec:
...
 ```
+
+## Preparing for Production
+
+TLS is a complicated beast and there are plenty of ways to use TLS certificates in ways which will cause problems
+down the line. We'd like to document a few of these potential gotchas here to help prepare users for running
+trust-manager safely in production.
```
wallrj, line 290: I think "complicated beast" might confuse non-uk readers. Maybe instead.
```suggestion
TLS is complicated and there are many ways to misuse TLS certificates. 
Here are some potential gotchas to be aware of before running
trust-manager in production.
```

content/docs/projects/trust-manager/README.md
```diff
@@ -56,79 +76,293 @@ spec:
...
+which contain it being updated first.
+
+Intermediates in this case become _de facto_ root certificates, which completely defeats the point of having
+intermediate certificates in the first place.
+
+Avoid using intermediates in any trust store wherever possible unless you're absolutely certain they should be included.
+An example of where it might be OK would be cross signing, which is not likely to be required in the general case.
+
+It would be better to copy just the root certificate to a new `ConfigMap` and use that as a source rather than trusting
+an intermediate.
```
wallrj, line 323: I find this section confusing because 

> For most users and applications `fullchain.pem` is the only correct choice.

contradicts

> Avoid using intermediates in any trust store wherever possible unless you're absolutely certain they should be included.

If the section is meant to be about what file users should instruct trust-manager to copy to the trust store.

Given the CA certificates:
 *  `Example Corp Root > Cloud Infrastructure > Linkerd`
 *  `Example Corp Root > VPN > Sales`

I would not want my linkerd client to trust a peer that presented a certificate signed by `Example Corp Root > VPN > Sales`.
So I would configure trust-manager to copy the intermediate CA  `Example Corp Root > Cloud Infrastructure > Linkerd` into the ConfigMap `linkerd/linkerd-identity-trust-roots`.

wallrj (changes requested):
A few questions and suggestions.

content/docs/projects/trust-manager/README.md
SgtCoDFish: Good idea, yeah. I'm struggling for time today but I'll try to get this done!

content/docs/projects/trust-manager/README.md
SgtCoDFish: That does seem simpler but I think I'll struggle to test that today to ensure that I'm writing sense :thinking:

Would we be OK to keep the current version of this - with the warts - and then come back to it?

content/docs/projects/trust-manager/README.md
SgtCoDFish: > I find this section confusing because ...

That makes sense, yeah. I can see how that would be confusing.

>  So I would configure trust-manager to copy the intermediate CA Example Corp Root > Cloud Infrastructure > Linkerd into the ConfigMap linkerd/linkerd-identity-trust-roots.

This kinda thing feels like an antipattern to me. It's _so_ easy to get this wrong and it turns every single intermediate into a de facto root certificate. It would be much better to have two roots, "Corp Root Infra" and "Corp Root VPN", in this scenario.

I think orgs would be well advised not to create PKI structures like this if at all possible. Certainly I wouldn't actively recommend it, although I acknowledge that this pattern exists today.

wallrj (approved):
Thanks for responding to my earlier comments.
The new content is great, but I do think some of the gotchas need refining in future PRs.

/lgtm

SgtCoDFish:
>  The new content is great, but I do think some of the gotchas need refining in future PRs.

Definitely agree! This is def a starting point not the end goal!

SgtCoDFish:
/unhold

The release is now done, so this can go live!

PR #1163, opened 2023-01-18
Use a grid layout on the 404 page consistent with other pages

wallrj:
image

vs

image


wallrj:
@relativelyrehan Please take a look at this and tell me if you think it's an improvement.

PR #1165, opened 2023-01-18
Execute next lint via npm run lint script

wallrj:
There's already an `npm run lint` "script" which in turn runs `next lint` which runs `eslint`. See
 * https://nextjs.org/docs/basic-features/eslint

Pushed to origin because new GitHub actions workflows don't seem to run from my fork.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -1,0 +1,15 @@
+name: check
+on:
+  push:
+    branches: [master]
+  pull_request:
+jobs:
+  lint:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+          cache: npm
+      - run: npm run lint


components/docs/VersionSelect.jsx
```diff
@@ -18,9 +17,11 @@ export default function VersionSelect({
...
   const [selectedVersion, setSelectedVersion] = useState(version)
 
-  versions = versions.sort(function(first, second){
-	  return compareVersions(labelFromVersion(first), labelFromVersion(second))
-  }).reverse()
+  versions = versions
+    .sort(function (first, second) {
+      return compareVersions(labelFromVersion(first), labelFromVersion(second))
+    })
+    .reverse()
```
wallrj, line 24: This is the formatting applied by `prettier` using the `.prettierrc` file which already exists in this repo.

lib/remark-plugins/heading-ids/index.js
```diff
@@ -16,7 +16,7 @@ export default function headingIds() {
 
         if (matched) {
           const id = matched[1]
-          if (!!id.length) {
+          if (id.length) {
```
wallrj, line 19: I think this has the same meaning, assuming that 0 is falsy and >0 is truthy

Issue #1168, opened 2023-01-20
Rendering issues for generated API docs


SgtCoDFish:
We run [gen-crd-api-reference-docs](https://github.com/ahmetb/gen-crd-api-reference-docs) in "markdown" mode, where it will process doc comments as markdown, but it's buggy and incorrect for some links.

An example is [this line](https://github.com/cert-manager/cert-manager/blob/1038ca44948702ff7c2063aec367be2b61831608/pkg/apis/certmanager/v1/types_issuer.go#L192) which is converted to:

```text
<p>Server is the connection address for the Vault server, e.g: &ldquo;<a href="https://vault.example.com:8200&quot;">https://vault.example.com:8200&rdquo;</a>.</p>
```

That invalid link ends up getting parsed a second time (when we render the file using next.js) and we end up creating a second `<a>` inside the first `<a>` which is even more invalid. That then confuses React which expects only valid HTML, and it prints lots of errors (but renders OK, I think).

A possible solution would be to parse all of the docs comments as HTML, but that's not possible because some of them [are invalid HTML](https://github.com/cert-manager/cert-manager/blob/1038ca44948702ff7c2063aec367be2b61831608/pkg/apis/acme/v1/types_challenge.go#L83-L89) and in any case a lot of them use Markdown so it'd be tedious to change it for all CRDs on all branches.

---

I created an example using just blackfriday which illustrates the error:

```go
package main

import (
	"fmt"

	"github.com/russross/blackfriday/v2"
)

func main() {
	fmt.Println(string(blackfriday.Run([]byte(`for example: "https://example.com"`))))
}
```

This produces this output:

```text
<p>for example: &ldquo;<a href="https://example.com&quot;">https://example.com&rdquo;</a></p>
```

On both the version of blackfriday used by gen-crd-api-reference-docs and on the latest version

SgtCoDFish:
Note: In #1170 we add trust-manager API doc generation with `crdoc` (https://github.com/fybrik/crdoc/) which doesn't seem to have the same problems with invalid HTML generation. It might not be a universal solution for everything but it might be better to use that for cert-manager, too.

PR #1170, opened 2023-01-23
Trust manager API doc generation

SgtCoDFish:
This uses crdoc instead of the tool we use for cert-manager. crdoc might be a candidate for use instead of our current tool (at least in part due to issues with  #1168)

This is a prerequisite for #1160 , which will be updated to include docs generated from trust-manager 0.4.0 (when it's tagged)

Preview: https://deploy-preview-1170--cert-manager-website.netlify.app/docs/projects/trust-manager-api-ref/


content/docs/manifest.json
```diff
@@ -344,6 +344,10 @@
             {
               "title": "trust-manager",
               "path": "/docs/projects/trust-manager.md"
+            },
+            {
+              "title": "trust-manager API Docs",
+              "path": "/docs/projects/trust-manager-api-ref.md"
```
SgtCoDFish, line 350: note: I couldn't decide between /projects and /reference

scripts/bin/crdoc
```diff
@@ -0,0 +1,61 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# This script downloads a crdoc binary for local use if such a binary doesn't already exist.
```
SgtCoDFish, line 21: note: this is copied from the htmltest script in the same directory. It could obviously be improved (multi arch support, etc) but this works well.

content/docs/manifest.json
inteon: we might need a API ref page for approval-policy too

inteon: I would like to have it all in a separate subsection (eg /docs/projects/reference or /docs/reference/projects)

wallrj: I suppose it depends whether we are trying to make all these projects appear as an integrated suite of software or whether we'd rather keep each project more self container.
I think for now, I'd prefer it in `/docs/projects/trust-manager/api-reference.md`. My 2c

SgtCoDFish: >  I think for now, I'd prefer it in /docs/projects/trust-manager/api-reference.md. My 2c

Totally reasonable, yeah. I'll go for making trust-manager its own section, then, as in:

```text
Projects
-> approver-policy
-> ...
-> trust-manager
---> Introduction
---> API Reference Docs
```

(I think it's a valuable long-term goal for trust-manager to be able to run entirely independently of cert-manager, to the point where it could live in its own independent docs site like `https://trust-manager.cert-manager.io`. I think a lot of the other subprojects won't reasonably exist independently of cert-manager, but trust-manager is useful even if you never install cert-manager)

SgtCoDFish:
Thanks both for the reviews / comments!

PR #1177, opened 2023-02-10
Remove unused verify-release script

wallrj:
This script is no longer necessary. 
It was called by the [prow CI job which we have since removed](https://github.com/jetstack/testing/pull/789). 

I've left  a single `verify` script which executes `npm run check`.


inteon (approved):
Thx for simplifying our dev tools
/approve
/lgtm

PR #1179, opened 2023-02-10
Only generate API and CLI docs on release

wallrj:
There's no point running the slow `./scripts/gendocs/generate` script on every build, because the files are already committed to the repository.
We should run that script when necessary and commit the results, which is what I've documented in this PR.

This should speed up the deployment and it should make it clear what is being generated, 
vs the current situation where problems with the generated output may go unnoticed unless a developer meticulously checks the CLI and API pages that are published.

image

```
$ time ./scripts/build-release
...

Export successful. Files written to /home/richard/projects/cert-manager/website/out

real    0m57.096s
user    4m48.119s
sys     0m27.888s
```

vs

```
$ time ./scripts/build-release
...
Export successful. Files written to /home/richard/projects/cert-manager/website/out

real    4m58.418s
user    14m51.916s
sys     2m13.773s
```


inteon:
@wallrj is it possible to still check that gendocs are not out-of-date?

wallrj:
> @wallrj is it possible to still check that gendocs are not out-of-date in CI?

It is possible, but they never did that. They just regenerated the docs and threw away the results.
https://github.com/cert-manager/website/blob/9f40b713605095d1a0a969ab5d5bda10f8785896/scripts/verify-release

I could add that check as a periodic action, but in another PR.

inteon (approved):
Based on what I understand:
- we don't want to fail unrelated PRs because a new API version was released
- we don't have to build the API docs for every PR, since they are not supposed to change (except from when we manually build the API docs)

I'll approve this PR.
From now on:
1. we will have to be more diligent about making sure we manually generate the API docs (there will be no non-related PRs that fail automatically)
2. we will not check anymore that the committed API docs are actually the result of running the generator tool

/hold
@wallrj please unhold if you agree with the points above or explain where I'm wrong so I can update the items above & revisit my approval

wallrj:
Thanks @inteon here are my answers:

>     we don't want to fail unrelated PRs because a new API version was released

That is true, but not related to this PR. 
None of the scripts which I deleted or modified in this PR (or in #1177) would have failed if the API version changed.

>     we will not check anymore that the committed API docs are actually the result of running the generator tool

"anymore" implies that this PR (or #1177) removed such a check, but such a check never existed.
The scripts have never checked whether the committed API docs were out of sync with the APIs.
They simply regenerated them and published them on the website without committing the changes (if there were any).

>     we don't have to build the API docs for every PR, since they are not supposed to change 
>  (except from when we manually build the API docs)

That is true.

>     we will have to be more diligent about making sure we manually generate the API docs 

That is true. And in this PR I've made it part of the release process that the API and CLI docs should be regenerated,
so that they get updated and committed at release time.

> (there will be no non-related PRs that fail automatically)

That was never the case.

inteon:
@wallrj thank you for the clarifying that I was incorrectly assuming that we checked for outdated docs already.

PR #1181, opened 2023-02-13
CSS: smaller font for code blocks, 1rem for normal font

maelvls:
The code blocks were unreadable. Their font size was set to 1.25 em, i.e., the same font size as the normal text content. Another problem of setting the same font size as the normal text is that monospaced fonts usually have a bigger footprint (`monospace` is a notable example). With all that, I propose to set the monospaced font to 0.8 em.

I also propose to reduce the font size to a "normal" font size of 1 em, like any other documentation website would do. Examples: Cilium is at 1 em, Linkerd is at 1 em.

| Before | After |
|--|--|
| Screenshot from 2023-02-13 12-52-19 | Screenshot from 2023-02-13 12-52-06 |

### DIFF
--- a/styles/docs.scss
+++ b/styles/docs.scss
@@ -27,7 +27,6 @@
     li {
       @apply mb-0;
       @apply mt-0;
-      @apply text-lg;
     }
 
     .docs-copy-btn {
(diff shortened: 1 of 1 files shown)


wallrj (approved):
Thanks. I agree that it is easier to read the code blocks with the smaller font.
I preferred the larger text, but I agree that it makes sense for the normal text to be at 1em.
`em` is relative to the font size of the parent element, right? 

/lgtm

PR #1182, opened 2023-02-13
Configure the netlify dev server to auto-reload redirects

wallrj:
Remove the alternative server scripts, which probably just confuse contributors.

`server-release` is also no longer needed  because we have documented that the CLI and API docs should be updated during the release process and it was  so slow as to be unusable anyway.


scripts/update-node_modules
maelvls, line 1: It looks like this script is still being sourced in various places:

```
rg /scripts/update-node_modules                           
scripts/index
32:"${REPO_ROOT}/scripts/update-node_modules"

scripts/gendocs/generate-new-import-path-docs
46:"${REPO_ROOT}/scripts/update-node_modules"

scripts/gendocs/generate-old-import-path-docs
51:"${REPO_ROOT}/scripts/update-node_modules"
```

wallrj: * https://github.com/cert-manager/website/pull/1200

PR #1183, opened 2023-02-13
Docs: Clarify zeroSSL setup instructions

chen-anders:
I went through a bit of confusion trying to setup ZeroSSL with certmanager.io this morning when I went through the docs. 

This PR updates the ZeroSSL setup instructions to clarify what values go where, so the next person doesn't end up spending a good hour wondering why they keep getting `Failed to register ACME account: 400 urn:ietf:params:acme:error:malformed: [External Account Binding] The Key Identifier was not recognized`

### DIFF
--- a/content/docs/tutorials/zerossl/zerossl.md
+++ b/content/docs/tutorials/zerossl/zerossl.md
@@ -63,14 +63,14 @@ Once you will get your credentials first step is to create seed with secrets. Th
 $ kubectl create secret generic \
        zero-ssl-eabsecret \
        --namespace=cert-manager \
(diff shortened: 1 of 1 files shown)


chen-anders:
> `Please note!` \
> EAB credentials are one-use only. Create additional pair for different environments.

Is this still accurate given the following?

> Since March 2022 all EAB credentials are reusable. One set of EAB credentials should be enough for most use cases. EAB credentials are limited to a maximum per user/per day.

maelvls:
It looks like you are correct. Can you remove the warning? Thanks!

maelvls:
Thank you so much!

/lgtm
/approve

PR #1191, opened 2023-02-22
Updates uninstall doc with backup and restore guide link

irbekrm:
Adds backup and restore guide link to uninstall doc as that seems to be a good place for it, it's hard to find otherwise.
I'm undecided if maybe the whole page should be moved under ./installation it's not really a tutorial as such.


SgtCoDFish (approved):
/lgtm
/approve

I love little improvements like this!

PR #1192, opened 2023-02-27
release-process: explain how and when to merge master into release-next

maelvls:
These changes were extracted from https://github.com/cert-manager/website/pull/1081#discussion_r1114395964.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,27 @@ Sometimes you'll want to add a comment which is only for documentation maintaine
 
 Use `{/* my comment */}` rather than the HTML-style comments you'd normally use for Markdown files. Other comment types will cause errors.
 
+### Task: Merging `master` into `release-next`
+
+In rare occasions, when writing documentation for an unreleased feature, you may
+notice that some recent changes in `master` aren't present in `release-next`. If
+that is a problem, you will want to update `release-next` branch with the latest
+changes from `master`. To update `release-next` with the changes made to
+`master`, follow these steps:
+
+1. Create a PR to merge `master` into `release-next` using [this magic
+   link][ff-release-next].
+2. If you see the label `dco-signoff: no`, add a comment on the PR with:
+
+   ```text
+    /override dco
+    ```
+
+   It is necessary because some the merge commits have been written by the bot
+   and do not have a DCO signoff.
+
+[ff-release-next]: https://github.com/cert-manager/website/compare/release-next...master?quick_pull=1&title=%5Brelease-next%5D+Merge+master+into+release-next&body=%3C%21--%0A%0AThe+command+%22%2Foverride+dco%22+is+necessary+because+some+the+merge+commits%0Ahave+been+written+by+the+bot+and+do+not+have+a+DCO+signoff.%0A%0A--%3E%0A%0A%2Foverride+dco
+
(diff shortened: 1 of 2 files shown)


content/docs/contributing/release-process.md
```diff
@@ -578,3 +578,32 @@ Other than the different `cert-manager/release` tag and `cmrel` version, you can
...
+to check that the "base branch" is correct.
+
+|                Description                 |  Base branch   |     Folder     |
+|--------------------------------------------|----------------|----------------|
+| Documentation PR for an existing feature   | `master`       | `content/docs` |
+| Documentation PR for an unreleased feature | `release-next` | `content/docs` |
+
+When creating a documentation PR for an unreleased feature, you will need to
+fast-forward the `release-next` branch to `master` before asking for a review of
+the PR:
```
maelvls, line 594: Copied from https://github.com/cert-manager/website/pull/1081#discussion_r1114395964:

> I do not agree with this.
> It may be desirable to bring release-next up to date before submitting your release-next PR,
> but usually only if the new feature you are documenting might conflict with some changes which have meanwhile been merged into the live documentation.
> 
> In any case, release-next doesn't have to be "fast-forwarded"...just merged....unless there's a way to create a PR which really does fast forward the commits from one branch to another. Is that possible?
> 
> This release process documentation should be in a separate PR.

maelvls: I agree. I think I misunderstood the implications of not having `release-next` up-to-date. Even in my case, I didn't have to update `release-next` to match `master`.

- I will rephrase to make it an optional step only needed if the changes you intend to make to the documentation relies on changes that were made recently to master and aren't in release-next.
- I will use the word "merge master into release-next" instead of "fast-forward release-next to master". I agree with you, it can't be fast-forwarding and merge commits are needed.

maelvls:
@wallrj I have removed most of what I had written since I realized that it was already explained in the README. The only thing I kept (and moved to the README) is how to merge master into release-next, and why that may be needed.

README.md
```diff
@@ -104,6 +104,27 @@ Sometimes you'll want to add a comment which is only for documentation maintaine
...
+1. Create a PR to merge `master` into `release-next` using [this magic
+   link][ff-release-next].
+2. If you see the label `dco-signoff: no`, add a comment on the PR with:
+
+   ```text
+    /override dco
+    ```
+
+   It is necessary because some the merge commits have been written by the bot
+   and do not have a DCO signoff.
```
wallrj, line 124: I wonder if it's possible to configure the bot to use --signoff when merging?

wallrj (approved):
Thanks @maelvls 

/lgtm

README.md
maelvls: I think @inteon investigated this idea ([message](https://github.com/jetstack/testing/pull/872#issuecomment-1506580069)):

> From what I found it is not possible to add DCO signoff in the commit message. However, I found that the DCO check should actually check if the committer is a bot and not complain in that case, this is not well-documented and needs further investigation. For now, I found that just merging in case all other checks are OK works fine.

inteon: That was specific for that generic autobumper tool.
I don't know what tool is used here.

PR #1195, opened 2023-03-03
Fix description of what happens to denied CertificateRequest

irbekrm:
When a in-tree issuer's CR gets denied, there is actually a race condition between issuing controller and certificate requests controller, the CR might or might not get retried.


content/docs/projects/approver-policy.md
```diff
@@ -72,13 +72,11 @@ selector.
 least one policy is appropriate for the request but none of those permit the
 request, the request is denied.**
 
-A denied CertificateRequest is considered to be permanently failed. If it was
-created for a Certificate resource, the issuance will be retried with
-[exponential
-backoff](../faq/README.md#what-happens-if-issuance-fails-will-it-be-retried)
-like all other permanent issuance failures. A CertificateRequest that is neither
-approved nor denied (because no matching policy was found) will not be further
-processed by cert-manager until it gets either approved or denied.
```
wallrj, line 81: > A CertificateRequest that is neither
> approved nor denied (because no matching policy was found) will not be further
> processed by cert-manager until it gets either approved or denied.

I think that is useful information and suggest we keep it.

content/docs/projects/approver-policy.md
```diff
@@ -72,13 +72,11 @@ selector.
...
+A denied CertificateRequest is considered to be permanently failed. Currently if
+a CertificateRequest that was created for a Certificate resource gets denied,
+issuance might or might not be retried with [exponential
+backoff](../faq/README.md#what-happens-if-issuance-fails-will-it-be-retried) (a
+new CertificateRequest created after a delay).
```
wallrj, line 79: Maybe add a bit more detail about when it might and when it might not get retried.
I can't remember what we found out.

wallrj (changes requested):

content/docs/projects/approver-policy.md
irbekrm: Currently it depends on which controller comes first in the race condition. However, I will update this PR and point it at next docs since https://github.com/cert-manager/cert-manager/pull/5887 got merged that makes the behaviour predictable.

content/docs/concepts/certificaterequest.md
```diff
@@ -60,16 +60,37 @@ manage the logic and life cycle of `CertificateRequests`.
...
+This condition should only be set by an approver.
+
+### Approved
+| Approved | Reason  | Condition Meaning                                                                                                                                                                                                                               |
+| ----- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| True | \<name of the approver\> | The `CertificateRequest` was approved by the approver. This `CertificateRequest` is approved and can be issued by the issuer.
+
+This condition should only be set by an approver.
+
+### InvalidRequest
```
irbekrm, line 89: This condition exists, but I am not aware of any issuer that actually sets it https://github.com/cert-manager/cert-manager/blob/master/pkg/apis/certmanager/v1/types_certificaterequest.go#L199

irbekrm:
Thank you for taking a look @wallrj I have updated the PR PTAL

https://deploy-preview-1195--cert-manager-website.netlify.app/docs/faq/#what-happens-if-issuance-fails-will-it-be-retried
https://deploy-preview-1195--cert-manager-website.netlify.app/docs/concepts/certificaterequest/#conditions
https://deploy-preview-1195--cert-manager-website.netlify.app/docs/contributing/external-issuers/#approval

^ these three paths are touched by this PR

I think it actually makes sense to merge this into master not next docs as the behaviour doesn't necessarily change, it just becomes consistent

content/docs/concepts/certificaterequest.md
```diff
@@ -60,16 +60,37 @@ manage the logic and life cycle of `CertificateRequests`.
...
 | False | Pending | The `CertificateRequest` is currently pending, waiting for some other operation to take place. This could be that the `Issuer` does not exist yet or the `Issuer` is in the process of issuing a certificate.                                   |
-| False | Failed  | The certificate has failed to be issued - either the returned certificate failed to be decoded or an instance of the referenced issuer used for signing failed. No further action will be taken on the `CertificateRequest` by it's controller. |
-| True  | Issued  | A signed certificate has been successfully issued by the referenced `Issuer`.                                                                                                                                                                   |
+| False | Failed  | The certificate has failed to be issued - either the returned certificate failed to be decoded or an instance of the referenced issuer used for signing failed. No further action will be taken on the `CertificateRequest` by it's controller and it can be considered terminally failed. |
```
wallrj, line 70:
```suggestion
| False | Failed  | The certificate has failed to be issued - either the returned certificate failed to be decoded or an instance of the referenced issuer used for signing failed. No further action will be taken on the `CertificateRequest` by its controller and it can be considered terminally failed. |
```

wallrj (changes requested):
Thanks @irbekrm 

I think all the new content makes sense.
Just spotted one typo.

content/docs/concepts/certificaterequest.md
irbekrm: Thank you @wallrj I keep automatically typing one instead of the other for these two. Should be fixed now.

wallrj (approved):
/lgtm

PR #1197, opened 2023-03-12
doc about new option default-cleanup-policy

RomanenkoDenys:
This is doc part of https://github.com/cert-manager/cert-manager/pull/5158


content/docs/usage/certificate.md
```diff
@@ -267,11 +267,54 @@ associated with compromised keys.
...
+currently relying on that certificate, but the certificate will no longer be renewed.
 The `Secret` needs to be manually deleted if it is no longer needed.
 
-If you would prefer the `Secret` to be deleted automatically when the `Certificate` is deleted, you need to configure your installation to pass the `--enable-certificate-owner-ref` flag to the controller.
+If you would prefer the `Secret` to be deleted automatically when the `Certificate`
+is deleted, you need to configure your installation to pass the 
+`--default-secret-cleanup-policy=OnDelete` flag to the controller.
+
+Also `Secret` deletion can be configured per-certificate, by setting
+`Certificate` `spec.cleanupPolicy` field.
```
maelvls, line 281: I suggest rewording this paragraph so that the `cleanupPolicy: OnDelete` is presented first.

```suggestion
If you would prefer the `Secret` to be deleted automatically when the `Certificate`
is deleted, you will need to set `cleanupPolicy: OnDelete` on the Certificate resource. Alternatively, you can add the flag `--default-secret-cleanup-policy=OnDelete` to the cert-manager controller pod in case you want all Secret resources to be cleaned up by default.
```

content/docs/usage/certificate.md
```diff
@@ -267,11 +267,54 @@ associated with compromised keys.
...
+
+#### `cleanupPolicy`
+
```
maelvls, line 284:
```suggestion
**FEATURE STATE**: This feature is available since cert-manager 1.12.

```

content/docs/usage/certificate.md
```diff
@@ -267,11 +267,54 @@ associated with compromised keys.
...
+kind: Certificate
+metadata:
+  name: my-cert
+spec:
+  ...
+  secretName: my-cert-tls
+  cleanupPolicy: OnDelete
+
+# Results in:
+
```
maelvls, line 304:
~~~suggestion
```

Results in:

```yaml
~~~

**Comment by maelvls** [comment]
Thanks for this PR! I left a couple of minor comments.

I changed and the base branch to `release-next` and rebased on top of it.

**Comment by RomanenkoDenys** [comment]
> Thanks for this PR! I left a couple of minor comments.
> 
> I changed and the base branch to `release-next` and rebased on top of it.

Thank you, i apply all suggestions !

**Comment by RomanenkoDenys** [comment]
> Updating the API reference documentation is quite tricky because the [`scripts/gendocs/generate`](https://github.com/maelvls/website/blob/9cec242e29d6cd5fcacd27e709f569cdb858f633/scripts/gendocs/generate-new-import-path-docs#L168) script hardcodes the "target" branch from which the docs will be generated. You can skip updating the API reference page in this PR.

I cannot update api reference in this PR

File: content/docs/usage/certificate.md [hunk 4]
```diff
@@ -267,11 +267,53 @@ associated with compromised keys.
...
+#### `cleanupPolicy`
+**FEATURE STATE**: This feature is available since cert-manager 1.12.
+
+`cleanupPolicy` is when this field is set to `OnDelete`, the owner reference
+is always created on the Secret resource and the secret will be automatically
+removed when the certificate resource is deleted. When this field is set to `Never`,
+the owner reference is never created on the Secret resource and the secret will not
+be automatically removed when the certificate resource is deleted.
+If the value of this field is unset this field "inherits" the value of
+the flag `--default-secret-cleanup-policy`.
```
maelvls, line 288:
```suggestion
The field `cleanupPolicy` can be used on the Certificate resource to configure
whether cert-manager should remove the leftover Secret resource when the 
Certificate is deleted.

When this field is set to `OnDelete`, the Secret resource will automatically
be removed when the Certificate is deleted. That works by adding an owner 
reference on the Secret resource.

When this field is set to `Never`, the Secret resource is preserved when
the Certificate is deleted.

When this field is unset, the value of the flag `--default-secret-cleanup-policy`
is used.
```

maelvls (commented):
Thank you for adding the documentation. Apart from two suggestion, it looks good to go.

Maybe it would require a bit more context as to why someone would use that feature, but the current documentation is already good.

I'll LGTM after the suggestion is reviewed.

erikgb:
This looks like an improvement of our docs. @RomanenkoDenys, are you able to target the master branch and rebase the PR?

PR #1199, opened 2023-03-22
Webhook troubleshooting: advise people to set `timeoutSeconds` to 30 seconds

maelvls:
The message `context deadline exceeded` is entirely useless when trying to debug it, and one of the main reasons this message appear is that our own Helm chart configures the "context" timeout to 10 seconds in [values.yaml](https://github.com/cert-manager/cert-manager/blob/0c071f8d2fbe0a568469f1bbf4eb7710ee0a2d0b/deploy/charts/cert-manager/values.yaml#L247).

### DIFF
--- a/content/docs/troubleshooting/webhook.md
+++ b/content/docs/troubleshooting/webhook.md
@@ -613,7 +613,7 @@ $ kubectl get mutatingwebhookconfigurations,validatingwebhookconfigurations cert
 10 10
 ```
 
-THis means that both webhooks are configured with a context timeout of 10
+This means that both webhooks are configured with a context timeout of 10
 seconds. To configure them to 30 seconds, run:
 
 ```bash
--- a/content/docs/troubleshooting/webhook.md
+++ b/content/docs/troubleshooting/webhook.md
@@ -600,41 +600,59 @@ receiving the HTTP response.
 > above error messages is a timeout that the API server decides when calling the
 > webhook. It is often set to 10 or 30 seconds.
 
+
+The first step to debug this issue is to make sure the `timeoutSeconds` field on
+the cert-manager mutating and validating webhook configurations we are
+configured to 30 seconds (the maximum value). By default, it is set to 10
+seconds, meaning that `context deadline exceeded` will potentially hide the
+other timeout messages. To check the value of the `timeoutSeconds` field, run:
+
+```console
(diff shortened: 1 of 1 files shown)


content/docs/troubleshooting/webhook.md
```diff
@@ -600,41 +600,59 @@ receiving the HTTP response.
 > above error messages is a timeout that the API server decides when calling the
 > webhook. It is often set to 10 or 30 seconds.
 
+
+The first step to debug this issue is to make sure the `timeoutSeconds` field on
+the cert-manager mutating and validating webhook configurations we are
+configured to 30 seconds (the maximum value). By default, it is set to 10
+seconds, meaning that `context deadline exceeded` will potentially hide the
+other timeout messages. To check the value of the `timeoutSeconds` field, run:
```
wallrj, line 608:
```suggestion
The first step to debug this issue is to make sure the `timeoutSeconds` field on
the cert-manager mutating and validating webhook configurations are
configured to 30 seconds (the maximum value). By default, it is set to 10
seconds, meaning that `context deadline exceeded` will potentially hide the
other timeout messages. To check the value of the `timeoutSeconds` field, run:
```

content/docs/troubleshooting/webhook.md
```diff
@@ -600,41 +600,59 @@ receiving the HTTP response.
...
+
+```console
+$ kubectl get mutatingwebhookconfigurations,validatingwebhookconfigurations cert-manager-webhook \
+  -ojsonpath='{.items[*].webhooks[*].timeoutSeconds}'
+10 10
+```
+
+THis means that both webhooks are configured with a context timeout of 10
+seconds. To configure them to 30 seconds, run:
```
wallrj, line 617:
```suggestion
This means that both webhooks are configured with a context timeout of 10
seconds. To configure them to 30 seconds, run:
```

wallrj (changes requested):
I noticed some typos.

I suggest changing the default timeout in the cert-manager Helm chart.
Is there any reason not to do that? 

This documentation can still be merged for the benefit of users of older versions of cert-manager.

wallrj (approved):
I applied the fixes for typos.

As mentioned earlier, in 1.14 (due for release on Jan 31 2024)  the default will be 30seconds, so this documentation may need updating again.

 * https://github.com/cert-manager/cert-manager/pull/6488
* https://github.com/cert-manager/cert-manager/releases/tag/v1.14.0-alpha.1 

/lgtm
/approve

PR #1201, opened 2023-03-22
Faster ./scripts/gendocs/generate script

maelvls:
The script `./scripts/gendocs/generate` is extremely slow when I run it from my home due to my slow internet connection. It re-downloads every Go module even though I already have them in `~/go/pkg/mod`. Also, it does a full clone of cert-manager/cert-manager.

**Before:** 2 minutes 33 seconds

**After:** 7 seconds


SgtCoDFish (approved):
/lgtm
/approve

See no reason not to merge this - easy to roll back if anything breaks! Thanks :grin:

PR #1202, opened 2023-03-22
Add section about client cert authentication for vault

joshmue:
Documentation for https://github.com/cert-manager/cert-manager/pull/4330

Things I was not sure about whether to do:
- Add a block with an example `kubernetes.io/tls` Secret
- Include other optional fields `mountPath` and `name` in example

### DIFF
--- a/content/docs/configuration/vault.md
+++ b/content/docs/configuration/vault.md
@@ -254,6 +254,43 @@ Kubernetes 1.24 and above.
               key: token
     ```
 
+### Authenticating with TLS Client Certificate
+
+Another way to authenticate against Vault is by presenting a client certificate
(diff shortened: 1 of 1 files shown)


content/docs/configuration/vault.md
```diff
@@ -254,6 +254,43 @@ Kubernetes 1.24 and above.
               key: token
     ```
 
+### Authenticating with TLS Client Certificate
+
```
maelvls, line 258: I suggest adding:

```suggestion
**FEATURE STATE**: This feature is available since cert-manager 1.12.
```

maelvls:
It looks good! Could you base this PR on the `cert-manager:release-next` branch? Thanks!

PR #1203, opened 2023-03-22
Clarify ingress-shim annotations

irbekrm:
Fixes https://github.com/cert-manager/cert-manager/issues/5872- this PR should clarify that the `cert-manager.io/cluster-issuer` ingress-shim annotation is a shortcut to allow users to refer to cert-manager.io ClusterIssuer, _not_ to refer to any cluster scoped issuer of any type.

The usage might seem slightly confusign for a user of an external cluster-scoped issuer type. However, there is not much we can do- we cannot remove the annotation and because for an external type group and kind needs to be specified anyway, there is no way how `cert-manager.io/cluster-issuer` could be used for external types in the same way as for internal types without adding extra annotations for cluster-issuer group and kind- which does not seem to bring much value.
So I think documenting this and keeping the existing state is the best option for us.

/cc @Hamidhasan

### DIFF
--- a/content/docs/usage/ingress.md
+++ b/content/docs/usage/ingress.md
@@ -49,13 +49,25 @@ spec:
 You can specify the following annotations on Ingress resources in order to
 trigger Certificate resources to be automatically created:
 
-- `cert-manager.io/issuer`:  the name of an Issuer to acquire the certificate
-  required for this Ingress. The Issuer *must* be in the same namespace as the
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm
/approve

Big improvement, thanks!

Hamidhasan:
Thanks @irbekrm for the super quick turnaround on this!

PR #1204, opened 2023-03-26
Update references to ingress objects in the nginx-ingress tutorial

yardenshoham:
From `ingress.extensions` to `ingress.networking.k8s.io`. The objects referenced in the page are not of type `ingress.extensions`


irbekrm (approved):
Thank you for the PR @yardenshoham !

This seems to be a leftover from older times, makes sense to fix 👍🏼 

/ok-to-test
/lgtm
/approve

PR #1205, opened 2023-03-27
Installation note about gke autopilot

sdarwin:


JoshVanL:
Thanks for the improvement @sdarwin, makes sense to me!

/lgtm
/approve
/ok-to-tes

sdarwin:
@JoshVanL thanks for the merge!  A further enhancement to that snippet could be to define which sections of the manifest need to be updated.  Being able to use "kubectl apply" instead of "helm" is appealing but might require more instruction, in the case of autopilot.

PR #1207, opened 2023-03-30
Updated info about issue-temporary-certificate

sdarwin:
The gist of commit 4232d0c64cb215dc487ac5ec4fe36a0e17713b25 , which remains in the documentation, is that 4 years ago all issues surrounding temporary certificates on GKE were "solved", and therefore you don't need "cert-manager.io/issue-temporary-certificate" any more.

That seems to be wrong.

Based on testing, the absence of a certificate at the beginning of the process will cause errors such as this:

"""
45s                   Warning   Sync                                Ingress/www-example                                                    Error syncing to GCP: error running load balancer syncing routine: error initializing translator env: secrets "www-example-tls" not found
"""

[The cert-manager GKE tutorial](https://cert-manager.io/docs/tutorials/getting-started-with-cert-manager-on-google-kubernetes-engine-using-lets-encrypt-for-ingress-ssl/) recommends manually creating a temporary certificate which further corroborates the idea that it's necessary. You can't launch an ingress that refers to a secret without the secret existing.

Therefore, it seems the clearest version of the text is to revert commit 4232d0c64cb215dc487ac5ec4fe36a0e17713b25 .   

Next, 

Also based on testing, something else is occurring which I don't completely understand.

"cert-manager.io/issue-temporary-certificate": "true" was not sufficient to generate a temporary certificate on GKE.

"acme.cert-manager.io/http01-edit-in-place": "true" must also be specified.

Consider the meaning of variable assignment `issue-temporary-certificate = true`.  It should issue a temporary certificate, right? And yet it didn't.  

Additionally, on the [Ingress page](https://cert-manager.io/docs/usage/ingress/) "issue-temporary-certificate" doesn't have it's own entry in the list of Supported Annotations. It would be good to add an entry in order to clarify the annotation.  

Yet another option (which may not be right) is if you were to say "issue-temporary-certificate is obsolete and should tend to be removed from the docs". In which case the current pull request ought to be changed and only discuss "http01-edit-in-place".

Any feedback is welcome, I could modify the wording further.

### DIFF
--- a/content/docs/usage/certificate.md
+++ b/content/docs/usage/certificate.md
@@ -136,24 +136,24 @@ documentation](../reference/api-docs.md#cert-manager.io/v1.KeyUsage).
 
 <h2 id="temporary-certificates-whilst-issuing">Temporary Certificates while Issuing</h2>
 
-On old GKE versions (`1.10.7-gke.1` and below), when requesting certificates
-[using the ingress-shim](./ingress.md) alongside the
-[`ingress-gce`](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress)
-ingress controller, `ingress-gce`
-[required](https://github.com/kubernetes/ingress-gce/pull/388) a temporary
-certificate must be present while waiting for the issuance of a signed
-certificate. Note that this issue was
-[solved](https://github.com/cert-manager/cert-manager/issues/606#issuecomment-424397233)
-in `1.10.7-gke.2`.
+When requesting certificates [using the ingress-shim](./ingress.md), the
+component `ingress-gce`, if used, requires that a temporary certificate is
+present while waiting for the issuance of a signed certificate when serving. To
+facilitate this, if the following annotation:
 
-```yaml
-# Required for GKE 1.10.7-gke.1 and below.
-cert-manager.io/issue-temporary-certificate": "true"
-```
+ ```yaml
+ cert-manager.io/issue-temporary-certificate": "true"
+ ```
 
-That made sure that a temporary self-signed certificate was present in the
-`Secret`. The self-signed certificate was replaced with the signed certificate
-later on.
+is present on the certificate, a self-signed temporary certificate will be
+present on the `Secret` until it is overwritten once the signed certificate has
+been issued.
+
+Adding the following annotation on an ingress will automatically set "issue-temporary-certificate" on the certificate:
+
+ ```yaml
+ acme.cert-manager.io/http01-edit-in-place: "true"
+ ```
 
 <h2 id="rotation-private-key">Rotation of the private key</h2>
 


wallrj (changes requested):
Thanks @sdarwin for digging in to this.

I wrote the [GKE tutorial including the empty Secret hack](https://cert-manager.io/docs/tutorials/getting-started-with-cert-manager-on-google-kubernetes-engine-using-lets-encrypt-for-ingress-ssl/#8-re-configure-the-ingress-for-ssl) and I was not aware of this annotation (or had forgotten about it).

Would you mind updating the tutorial too, if that hack is not actually necessary?

And add some links here and there to the ingress-gce repo and to a specific bug in that repo (create one if necessary) about how the fix [Fire warning event instead of hard failing if TLS certificate is not present](https://github.com/kubernetes/ingress-gce/pull/388#top) appears to have regressed. (if that is what has happened).

sdarwin:
> I wrote the GKE tutorial including the empty Secret hack

That is great! The hack describes the issue very well and what's required. I would consider leaving it in the tutorial as an either/or scenario. Either manually create an empty secret or use an annotation, so the user understands the problem better.

> appears to have regressed

From the perspective of ingress-gce this may not be a regression. Notice the quote in the discussion thread from a ingress-gce developer:

"""
Would be possible for the cert-manager to detect this and generate a self-signed cert to unblock creation of the Ingress and then do the update to final cert from let's encrypt. I think this is the behavior of ingress-nginx (self-signed cert)

It feels wrong to proceed with the LB configuration in a situation when the configuration is not valid.
"""

If they are going to keep stubbornly returning to their position that the LB is misconfigured without having a certificate, so be it. Cert-manager must be able to handle with that situation.  

Then, at that point, tracking the history of the decisions (back and forth again) by linking to various github issues only muddles the documentation. Because the final conclusion is that cert-manager must provision a temporary certificate. 

> Would you mind updating the tutorial too

Yes, I could update the tutorial. A difficulty though, is that I would prefer to update the tutorial by adding this annotation:

```yaml
 cert-manager.io/issue-temporary-certificate": "true"
```

however that won't work. Why is "acme.cert-manager.io/http01-edit-in-place" not merely optional? Maybe there is an important technical reason. Or, a bug.

sdarwin:
To add more details about the problem, I was following the GKE Tutorial which includes this section:

```
    solvers:
    - http01:
        ingress:
          name: web-ingress
```

That means the Issuer is being instructed to use the ingress "web-ingress".

- Isn't edit-in-place sort of implied if I have told it to use "web-ingress" specifically as the solver ingress?
- For the sake of argument, lets say its not implied. http01-edit-in-place is false. Then it's supposed to create a different LB. But that never happens.
- presumably "issue-temporary-certificate" by itself doesn't say anything about one ingress or another, one LB or another.  Only that it should "issue-temporary-certificate" if one is missing. But that does not happen.

content/docs/usage/certificate.md
```diff
@@ -136,24 +136,24 @@ documentation](../reference/api-docs.md#cert-manager.io/v1.KeyUsage).
...
+been issued.
+
+In newer versions of cert-manager, this should be accompanied by another annotation:
 
-That made sure that a temporary self-signed certificate was present in the
-`Secret`. The self-signed certificate was replaced with the signed certificate
-later on.
+ ```yaml
+ acme.cert-manager.io/http01-edit-in-place: "true"
+ ```
```
wallrj, line 156: I don't think this is right. My understanding is that the `acme.cert-manager.io/http01-edit-in-place: "true"` annotation should be applied to the Ingress resource, not the Certificate resource.

And when applied to the Ingress resource, it results in cert-manager ingress-shim also adding the annotation `cert-manager.io/issue-temporary-certificate": "true"` to the linked Certificate resource:

https://github.com/cert-manager/cert-manager/blob/fd9c01fa8ce02decf22f917535af26ec9468f988/pkg/controller/certificate-shim/sync.go#L594-L611

And I don't really see the point of it anyway, because you can name the target Ingress in `Issuer.Spec.ACME.HTTP01.Ingress.Name`...I suppose it avoids you having to keep the Ingress name in sync in two places.

I might also be misunderstanding how all this works.

wallrj (changes requested):

sdarwin:
> should be applied to the Ingress resource, not the Certificate resource.

You are right!  I have totally confused these two things.  The Ingress versus the Certificate.

wallrj:
There are more docs here:

https://github.com/cert-manager/cert-manager/blob/fd9c01fa8ce02decf22f917535af26ec9468f988/pkg/apis/acme/v1/types.go#L20-L39

But I think this one is wrong:
> 	// IngressEditInPlaceAnnotationKey is used to toggle the use of ingressClass instead
>	// of ingress on the created Certificate resource
>	IngressEditInPlaceAnnotationKey = "acme.cert-manager.io/http01-edit-in-place"

wallrj (approved):
Thanks @sdarwin 

I can see that you've reverted back to the previous wording and added a note about http01-edit-in-place.

As mentioned in the comments, I think there are various other improvements we can make to the API docs and to the tutorial, but those shouldn't hold this up.

/lgtm
/hold until I've had a chance to run this past @maelvls who has much more recent experience with the ingress-shim than me.

sdarwin:
@wallrj sounds good.   
About half my comments in this thread were at least partially mistaken then, because I was mixing up Certificate and Ingress annotations.  Thanks for clarifying.
Another thought, to what extent could "issue-temporary-certificate" and "edit-in-place" by made more automatic, such as defaults?  But, that would require more investigation.

wallrj:
@sdarwin I checked with @maelvls and he sees no problem with this change. I will try and make the other improvements to the GKE tutorial.

/unhold

PR #1211, opened 2023-04-14
We settled on 26 april 2023 for the release of 1.12

maelvls:


content/docs/installation/supported-releases.md
```diff
@@ -28,7 +28,7 @@ cert-manager expects that ServerSideApply is enabled in the cluster for all vers
 
 | Release  |  Release Date  |  End of Life   | [Supported Kubernetes versions][s]  | [Supported OpenShift versions][s] |
 |----------|:--------------:|:--------------:|:-----------------------------------:|:---------------------------------:|
-| [1.12][] | End of April, 2023  | End of August, 2023 |            1.22 → 1.27              |            4.9 → 4.13             |
+| [1.12][] | Apr 26, 2023  | End of August, 2023 |            1.22 → 1.27              |            4.9 → 4.13             |
```
SgtCoDFish, line 31: suggestion: this suggests that we'll support 1.27 on release but I really doubt we will. I think we _wanted_ to support it, but we've not tested it yet and probably won't be able to before the release. It might work but we won't have tested.

Should we say "1.22 -> 1.26" for now, and then later we can add support for 1.27 and backport to cert-manager 1.12 after release?

maelvls: Yes, that sounds sensible. FYI @irbekrm

PR #1214, opened 2023-04-18
Fix selector.issuerRef example to use map instead of list

wallrj:
Fixes: https://github.com/cert-manager/approver-policy/issues/217

### DIFF
--- a/content/docs/projects/approver-policy.md
+++ b/content/docs/projects/approver-policy.md
@@ -298,7 +298,7 @@ spec:
   ...
   selector:
     issuerRef:
-    - name: "my-ca"
+      name: "my-ca"
       kind: "*Issuer"
       group: "cert-manager.io"
 ```


content/docs/projects/approver-policy.md
```diff
@@ -298,7 +298,7 @@ spec:
   ...
   selector:
     issuerRef:
-    - name: "my-ca"
+      name: "my-ca"
       kind: "*Issuer"
       group: "cert-manager.io"
```
wallrj, line 303: We don't yet publish the API docs for open-source approver-policy, but in the [Jetstack documentation](https://platform.jetstack.io/documentation/reference/approver-policy-enterprise/kubernetes-api#certificaterequestpolicyspecselector), 
you can see that this issuerRef should be an object rather than a list

JoshVanL:
Thanks!
As you say, we should probably be exporting generated docs 

/lgtm

PR #1216, opened 2023-04-26
Wrap the cert-manager re-configuration instructions for easier reading

wallrj:
**Preview**: https://deploy-preview-1216--cert-manager-website.netlify.app/docs/projects/approver-policy/

In the existing documentation , the `helm upgrade` command also installs and the command is one long line which is difficult to read.

The documentation assumes that  if the user has already installed cert-manager, that they used the `--set installCRDs=true` flag,
which is not the recommended helm install method so I added a note about that.

I considered recommending `helm upgrade --reuse-values` to retain what ever values the user originally supplied when they installed cert-manager, but that's not safe either, because `extraArgs` may or may not have already been customized.
Instead I recommended that the user check existing values before reconfiguring.

### DIFF
--- a/content/docs/projects/approver-policy/README.md
+++ b/content/docs/projects/approver-policy/README.md
@@ -8,22 +8,55 @@ approver-policy is a cert-manager
 that will approve or deny CertificateRequests based on policies defined in
 the `CertificateRequestPolicy` custom resource.
 
-## Installation
+## Prerequisites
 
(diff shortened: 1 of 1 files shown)


irbekrm (approved):
Thanks Richard, I have read through the changes and I think this is an improvement although I regret that we spend so much time on Helm specific stuff. I guess at some point it'd be good to have a user survey to understand what are the most common installation mechanisms and what the trend is to see whether this is reasonable.

/lgtm

PR #1217, opened 2023-04-28
Document the liveness probes in cert-manager components

wallrj:
**Preview:** https://deploy-preview-1217--cert-manager-website.netlify.app/docs/installation/best-practice/

Added documentation for the next release to compliment https://github.com/cert-manager/cert-manager/pull/5962
 * https://github.com/cert-manager/cert-manager/pull/5962

### DIFF
--- a/content/docs/installation/best-practice.md
+++ b/content/docs/installation/best-practice.md
@@ -41,11 +41,19 @@ and the [timings and thresholds can be configured using Helm values](https://git
 
 ### controller
 
-> ℹ️ The cert-manager controller liveness probe was introduced in cert-manager `v1.12.0`.
+> ℹ️ The cert-manager controller liveness probe was introduced in cert-manager release `1.12`.
 
 The cert-manager controller has a liveness probe, but it is **disabled by default**.
 You can enable it using the Helm chart value `livenessProbe.enabled=true`,
 but first read the background information below.
+
+> 📢 The controller liveness probe is a new feature in cert-manager release 1.12
+> and it is disabled by default, as a precaution, in case it causes problems in the field.
+> [Please get in touch](../contributing/README.md)
+> and tell is if you have enabled the controller liveness probe in production
+> and tell us whether you would like it to be turned on by default.
+> Tell us about any circumstances where the controller has become stuck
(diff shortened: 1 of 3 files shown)


content/docs/installation/best-practice.md
```diff
@@ -1,19 +1,102 @@
...
+
+### Background Information
+
+The cert-manager `controller` process and the `cainjector` process,
+both use the Kubernetes [leader election library](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection),
+to ensure that only one replica of each process can be active at any one time.
+The Kubernetes control-plane components also use this library.
+
+The leader election code runs in a loop in a separate thread (go routine).
+If it initially wins the leader election race and if it later fails to renew its leader election lease, it exits.
```
irbekrm, line 80: This is interesting, I actually didn't realize that the behaviour was like that. I imagined that if a replica loses leader election (because another replica managed to become a leader) it would just become a standby and keep attempting to become a leader. I think that's how it works in raft.
However I do see that the way we do it is also how c/r does it https://github.com/kubernetes-sigs/controller-runtime/blob/cd65cb25d314f40a329a688f4714fe3282589e97/pkg/manager/internal.go#L646 so perhaps it makes sense to exit with an error because a leader election cannot be lost in a normal flow (there's no way how leadership could shift unless there is a fatal errror). Something to look into for myself 👀

irbekrm (changes requested):
Thanks Richard, this makes sense and is a great write-up, thanks for adding all those links.

Could we also mention that we might want to enable the controller liveness probe by default in future if we see that there are actual edge cases (not cert-manager bugs) where this would be useful and can we add a note asking users to report back if they have discovered such?

wallrj:
> Could we also mention that we might want to enable the controller liveness probe by default in future if we see that there are actual edge cases (not cert-manager bugs) where this would be useful and can we add a note asking users to report back if they have discovered such?

Done.

irbekrm (approved):
/lgtm

irbekrm (approved):
Thank you Richard

/lgtm

PR #1218, opened 2023-05-02
Add note about linux/arm64 support

SgtCoDFish:
Follows on from https://github.com/cert-manager/cert-manager/pull/5814


inteon (approved):
Thanks
/lgtm
/approve

PR #1220, opened 2023-05-05
Release notes for 1.12

maelvls:
TBD:

- [x] Change all `@name` to `[@name](https://github.com/name)`
- [x] Change all `#1234` to `[#1234](https://github.com/cert-manager/cert-manager/pulls/1234)`
- [x] Fill in the section "Support for JSON logging"
- [x] Fill in the section "Lower memory footprint" (tbd @irbekrm)
- [x] Add a section "Decrease of the attack surface" or something like that to talk about each binary being its own Go module with much less dependencies for each binary (ask @SgtCoDFish)


content/docs/release-notes/release-notes-1.12.md
```diff
@@ -3,8 +3,102 @@ title: Release 1.12
...
+file, providing us with more flexibility to react to CVEs.
+
+In the past, we have been unable to offer security patches due to unsupported
+dependencies. An example of this was seen with Helm: in cert-manager 1.10, we
+were unable to fix a CVE reported in Helm because Helm only offers security
+patches for its latest minor version.
+
+While this doesn't decrease the dependency attack surface for any of the
+binaries, it does allow us to react more quickly and effectively when a
+vulnerability is reported.
```
SgtCoDFish, line 100: suggestion: Would it be worth coming at this from a slightly different angle?

```markdown
### Reduced Transitive Dependencies in cert-manager

cert-manager 1.12 separates some of its functionality into separate Go modules,
with the aim of decreasing the amount of time and toil for maintainers to patch
vulnerable dependencies.

A side effect of this is that it's now easier to import cert-manager in Go,
in terms of transitive dependencies that might show up in your `go.mod` files or
potential version conflicts between cert-manager and your other dependencies.

The caveat here is that we still only recommend importing cert-manager in
[very specific circumstances](../contributing/importing.md), and the module changes
mean that if you imported some paths (specifically under `cmd` or some paths under `test`)
you might see broken imports when you try to upgrade.

If you experience a break as part of this, we're sorry and we'd be interested to chat about it.
The vast majority of projects using cert-manager should notice no impact, and there should
be no runtime impact either.
```

maelvls: > ```md
> ##### Reduced Transitive Dependencies in cert-manager
> ```

The title "Reduced Transitive Dependencies in cert-manager" doesnt state the benefit that end-users can expect from that change.

> cert-manager 1.12 separates some of its functionality into separate Go modules,
> with the aim of decreasing the amount of time and toil for maintainers to patch
> vulnerable dependencies.

I think first paragraph should focus on the problem that we have been solving: reducing the impacts of breaking the rule "no minor bumps in patches" when a CVE is declared in an unsupported minor version of a dependency. Hence, we are increasing (or sustaining) the security of the end-user installations.

> ```md
> A side effect of this is that it's now easier to import cert-manager in Go,
> in terms of transitive dependencies that might show up in your `go.mod` files or
> potential version conflicts between cert-manager and your other dependencies.
> 
> The caveat here is that we still only recommend importing cert-manager in
> [very specific circumstances](../contributing/importing.md), and the module changes
> mean that if you imported some paths (specifically under `cmd` or some paths under `test`)
> you might see broken imports when you try to upgrade.
> 
> If you experience a break as part of this, we're sorry and we'd be interested to chat about it.
> The vast majority of projects using cert-manager should notice no impact, and there should
> be no runtime impact either.
> ```

These three paragraph look great, I hadn't thought about these.

maelvls: I added the three paragraphs and re-wrote my paragraphs. Please take another look

maelvls:
@irbekrm Just in case, I wanted to let you know that I have finished editing the release notes of every PR.

Many of the `release-note` blocks were worded as commit messages instead of user-facing release notes.

For example, this release note isn't helpful:

> Adds missing comparisons for certain fields which were incorrectly skipped if a LiteralSubject was set

A user-oriented release note would be

> When using the `literalSubject` field on a Certificate resource, the IPs, URIs, DNS names, and email addresses segments are now properly compared.

At some point we need to address this, but I have no idea how 😅

content/docs/release-notes/release-notes-1.12.md
SgtCoDFish: > These three paragraph look great, I hadn't thought about these.

I did a _really_ awful job of communicating this change, wow

SgtCoDFish: New text looks fine!

content/docs/release-notes/release-notes-1.12.md
```diff
@@ -3,8 +3,115 @@ title: Release 1.12
...
+(`cert-manager#6074`)[https://github.com/cert-manager/cert-manager/issues/6074].
+
+Additionally, controller no longer watches and caches all `Pod` and `Service`
+resources.
+See [`cert-manager#5976`](https://github.com/cert-manager/cert-manager/pull/5976) for implementation.
+
+#### Cainjector
+
+[Cainjector's](../concepts/ca-injector.md) control loops have been refactored, so by default it should
+consume up to twice less memory, see
```
SgtCoDFish, line 62: drive-by suggestion: minor rewording

```suggestion
consume up to half as much memory as before, see
```

irbekrm (changes requested):
@maelvls could we update the sections to add links to the relevant PRs with implementation/design? I usually find that useful when reading release notes for kube etc

maelvls:
I have added the links to the documentation, design, and PRs in the "major themes". The Go modules one was totally missing and people would have been confused not to see the link to the design document 😅

PR #1221, opened 2023-05-08
Fix a typo in usage/certificate docs

chamankang:
Get rid of a double quote from the issue-temporary-certificate annotation key.


inteon (approved):
Thanks a lot!
/lgtm
/approve

PR #1222, opened 2023-05-12
Fix leading whitespaces in YAML code block

thomasvn:
The leading whitespaces seemed to have caused an issue when rendered to markdown


inteon (approved):
Thanks!
/lgtm
/approve

PR #1226, opened 2023-05-19
trust-manager v0.5.0

SgtCoDFish:
We changed the trust-manager Helm CRD in https://github.com/cert-manager/trust-manager/pull/102 because we made it optional at install time, and in turn added a 'plain' CRD at a different path, so this PR makes that change.

Also adds detail on the new JKS support and updates the API reference doc.

### DIFF
--- a/content/docs/projects/trust-manager/api-reference.md
+++ b/content/docs/projects/trust-manager/api-reference.md
@@ -226,6 +226,13 @@ Target is the target location in all namespaces to sync source data to.
         </tr>
     </thead>
     <tbody><tr>
+        <td><b><a href="#bundlespectargetadditionalformats">additionalFormats</a></b></td>
+        <td>object</td>
+        <td>
+          AdditionalFormats specifies any additional formats to write to the target<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
         <td><b><a href="#bundlespectargetconfigmap">configMap</a></b></td>
         <td>object</td>
         <td>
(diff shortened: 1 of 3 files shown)


content/docs/projects/trust-manager/api-reference.md
```diff
@@ -243,6 +250,56 @@ Target is the target location in all namespaces to sync source data to.
...
+            <th>Required</th>
+        </tr>
+    </thead>
+    <tbody><tr>
+        <td><b><a href="#bundlespectargetadditionalformatsjks">jks</a></b></td>
+        <td>object</td>
+        <td>
+          KeySelector is a reference to a key for some map data object.<br/>
+        </td>
+        <td>false</td>
```
SgtCoDFish, line 273: self review: this isn't a great description for the field, but it comes directly from the CRD and so changing it will be a pain. I think it's fine, the main documentation file is clear enough IMO

content/docs/projects/trust-manager/api-reference.md
```diff
@@ -243,6 +250,56 @@ Target is the target location in all namespaces to sync source data to.
...
+      </tr></tbody>
+</table>
+
+
+### `Bundle.spec.target.additionalFormats.jks`
+
+
+KeySelector is a reference to a key for some map data object.
```
SgtCoDFish, line 281: self review: As above, not ideal but probably file.

content/docs/projects/trust-manager/api-reference.md
```diff
@@ -428,6 +492,56 @@ Target is the current Target that the Bundle is attempting or has completed sync
 </table>
 
 
+### `Bundle.status.target.additionalFormats`
+
+
+AdditionalFormats specifies any additional formats to write to the target
```
SgtCoDFish, line 498: self review: having this documented in such detail for the status (as well as the spec above) is probably a little redundant too but it's not the end of the world.

scripts/gendocs/generate-trust-manager
```diff
@@ -47,7 +47,7 @@ gendocs() {
 	echo "+++ Generating trust-manager reference docs..."
 
 	$CRDOC \
-		--resources "$tmpdir/deploy/charts/trust-manager/templates/trust.cert-manager.io_bundles.yaml" \
+		--resources "$tmpdir/deploy/crds/trust.cert-manager.io_bundles.yaml" \
```
SgtCoDFish, line 50: self review: see PR description / commit messages

jakexks (approved):
I was just looking for these docs :)

SgtCoDFish:
Thank you @jakexks :heart:

PR #1234, opened 2023-05-25
Correct the cmctl release generation flow

irbekrm:
The flow for cmctl release should be:

- get all commits we want to release in release branch where the top commit is X
- bump cert-manager dependency version to X in cmctl go.mod and cherry-pick that change to release branch
- continue with release as normal by pushing the two tags etc


inteon:
I think we do not need this change anymore, because we moved cmctl to a separate repo: https://github.com/cert-manager/cmctl
We do however still need to update the release process and remove the cmctl steps.

PR #1235, opened 2023-05-26
Add deletion of validatingwebhookconfigurations to stuck uninstall

oernii:


content/docs/installation/kubectl.md
```diff
@@ -117,6 +117,11 @@ First, delete existing cert-manager webhook configurations, if any:
 kubectl delete mutatingwebhookconfigurations cert-manager-webhook
 ```
 
+Then try do delete the corresponding validatingwebhookconfigurations
+```bash
+kubectl delete validatingwebhookconfigurations cert-manager-webhook
```
irbekrm, line 122: Could we just move this line to the already existing code block under the line that deletes mutating configurations?

irbekrm (changes requested):
Thanks for the PR @oernii !

I've added a small suggestion, otherwise happy to approve. 

We could also add the same section to helm uninstall doc too https://github.com/cert-manager/website/blob/master/content/docs/installation/helm.md#namespace-stuck-in-terminating-state

content/docs/installation/kubectl.md
oernii: Sure

inteon: done

PR #1236, opened 2023-05-26
Fix supported-releases.md

maelvls:
I mistakenly set the next release to September 2023 when it should say July 2023.

PR #1237, opened 2023-05-30
fix(supported-releases): fix the year of the 1.12 release date and eol

rinx:
I've found that the release date of 1.12 is written as `2024`, but I think it should be `2023`.

Also, please check the comment in #1236.
https://github.com/cert-manager/website/pull/1236/files#r1209815760


inteon (approved):
Thank you for fixing this!

PR #1239, opened 2023-06-02
Explain what FAO stands for

irbekrm:
Explain what FAO stands for in `controller.cert-manager.io/fao` label.

See https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1685704657807239


maelvls (approved):
/lgtm

Thanks!

PR #1244, opened 2023-06-12
document new subject ingress/gateway annotations

ctrought:
Follow up to add docs for additional ingress/gateway annotations added in v1.12 https://github.com/cert-manager/cert-manager/pull/4502


inteon:
/approve
/lgtm
Thanks for adding documentation for this new feature!

PR #1248, opened 2023-06-15
Document CRD considerations when using helm

hawksight:
Attempting to clarify the available options for managing cert-manager CRDs and how they differ to standard helm method. I did add a recommendation for the user which I am unsure if that is the "right" thing to do here. My default is to be secure and managing CRDs outside of helm gives the most safety in that you shouldn't ever have a scenario where helm removes those CRDs causing issues.

All that said, the `--set installCRDs=true` option is very convenient and I use it daily. It seems most of the users I asked use this option and don't have concerns. So the intent here is just to highlight the considerations and give users an idea about which method they should choose. Ultimately this is down to each user of course and people can do what they like.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -601,6 +601,7 @@ v4.4.1
 liveness
 apiservices
 arm64
+IaC
 
 # TEMPORARY
 # these are temporarily ignored because the spellchecker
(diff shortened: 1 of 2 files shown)


maelvls (approved):
I read the recommendations, they look good! Thank you so much for taking the time to write this PR.

/lgtm

PR #1251, opened 2023-06-19
Add queried endpoint to check Vault status when initializing issuer

nrogerarkhn:
The Vault Issuer queries an endpoint to check is the Vault instance is healthy. When configuring the Issuer with a remote Vault instance protected behind a firewall, the 'Vault sealed or unintialized' error could be thrown when the endpoint could simply not be reached. There was no sign of this endpoint in the docs, so adding it for commodity.


inteon (approved):
/approve
/lgtm
Thanks for the improvement!

PR #1229, opened 2023-05-23
[Release 1.12 + 1.13.0-alpha.0] Update Release Process

maelvls:
I re-arranged the website PR instructions so that the pre-release steps for the website are shown first. I also made it clearer that we produce two types of release notes: the GitHub release description and the "Release Note" page on the website.

- Mention "How we determine supported Kubernetes".
- Mention how to transform GitHub references into links in the `release-notes.md` file.
- Mention bump 1.12 in generate-new-import-path-docs
- Mention the freeze of docs/ that gets copied into v1.12-docs/.
- Mention how to bump the gen scripts.
- Show a command that to bump versions in install instructions.
- Mention manifest.json that needs to be updated.
- Mention the file release-notes/README.md's index that needs updating.

I also had to re-order the bullet points due to conflicts with Irbe's changes to the release process document. Apologies for the noise that these changes introduce!

Preview: https://deploy-preview-1229--cert-manager-website.netlify.app/docs/contributing/release-process/

Issue #1243, opened 2023-06-05
V1.12 is missing an upgrade document


plnordquist:
Prior versions of cert-manager have upgrade documents, for example https://cert-manager.io/docs/installation/upgrading/upgrading-1.10-1.11.  The new v1.12 version is missing it's own document describing any breaking changes.

maelvls:
Thanks for reporting this issue! This is my fault, I forgot to add upgrade notes for this version of cert-manager.

PR #1252, opened 2023-06-19
fix: https://cert-manager.io/v1.12-docs doesn't work

maelvls:
I forgot to change the manifest.json when releasing v1.12. The command is:

```bash
sed -i.bak 's|/docs/|/v1.12-docs/|g' content/v1.12-docs/manifest.json
cat content/v1.12-docs/manifest.json \
  | jq 'del(.. | select(.path? | select(.) | test(".*(installation/supported-releases.md|installation/upgrading|release-notes).*")))' \
  | jq 'del(.. | select(.routes? == []))' >/tmp/manifest \
     && mv /tmp/manifest content/v1.12-docs/manifest.json
```

I also edited the release process document to reflect that in #1229.

### DIFF
--- a/content/v1.12-docs/manifest.json
+++ b/content/v1.12-docs/manifest.json
@@ -6,187 +6,187 @@
       "routes": [
         {
           "title": "Introduction",
-          "path": "/docs/README.md"
+          "path": "/v1.12-docs/README.md"
         },
         {
           "title": "Getting Started",
-          "path": "/docs/getting-started/README.md"
+          "path": "/v1.12-docs/getting-started/README.md"
         },
         {
           "title": "Installation",
           "routes": [
             {
               "title": "Introduction",
-              "path": "/docs/installation/README.md"
+              "path": "/v1.12-docs/installation/README.md"
             },
             {
               "title": "Supported Releases",
-              "path": "/docs/installation/supported-releases.md"
+              "path": "/v1.12-docs/installation/supported-releases.md"
             },
             {
               "title": "Cloud Compatibility",
-              "path": "/docs/installation/compatibility.md"
+              "path": "/v1.12-docs/installation/compatibility.md"
             },
             {
               "title": "kubectl apply",
-              "path": "/docs/installation/kubectl.md"
+              "path": "/v1.12-docs/installation/kubectl.md"
             },
             {
               "title": "Helm",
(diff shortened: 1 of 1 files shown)


SgtCoDFish:
The preview build failure looks legitimate!

content/v1.12-docs/manifest.json
```diff
@@ -436,268 +314,151 @@
...
-            {
-              "title": "v0.2",
-              "path": "/docs/release-notes/release-notes-0.2.md"
-            },
-            {
-              "title": "v0.1",
-              "path": "/docs/release-notes/release-notes-0.1.md"
-            }
-          ]
+          "routes": []
```
SgtCoDFish, line 430: suggestion: If the list of routes is empty, this will create a dropdown menu item with no entries in it (visible in the preview). Can we remove this section entirely for v1.12?

(It doesn't exist for, say, v1.11)

content/v1.12-docs/manifest.json
```diff
@@ -2,191 +2,69 @@
...
-                {
-                  "title": "v0.3 to v0.4",
-                  "path": "/docs/installation/upgrading/upgrading-0.3-0.4.md"
-                },
-                {
-                  "title": "v0.2 to v0.3",
-                  "path": "/docs/installation/upgrading/upgrading-0.2-0.3.md"
-                }
-              ]
+              "routes": []
```
SgtCoDFish, line 51: suggestion: Should this be removed too?

SgtCoDFish (changes requested):
Thanks for fixing this! Couple of things that might need a second look.

hawksight:
Relates to #1243

content/v1.12-docs/manifest.json
maelvls: Good point, I'll add that to the `jq` command in release-process.md.

content/v1.12-docs/manifest.json
maelvls: Done! Thanks

content/v1.12-docs/manifest.json
maelvls: Done!

SgtCoDFish (approved):
/lgtm
/approve

This looks good, thanks!

Issue #1241, opened 2023-06-04
Remove Bitnami kubeprod as installation method


ThomasVitale:
In the "Installation > Other tools" page, the Bitnami kubeprod project is listed as an alternative method to install cert-manager. However, that project has been archived two years ago and hasn't been updated since. See [announcement](https://github.com/vmware-archive/kube-prod-runtime/commit/a79bcda68a3423de01280c198f68ac05a1fa077f) for further details.

SgtCoDFish:
Thank you for raising this! Obviously took a while to get to, but I picked this at random and decided to solve it 😁

PR #1253, opened 2023-06-20
Remove Bitnami kubeprod as installation method

mprochowski:
PR is based on the issue https://github.com/cert-manager/website/issues/1241 

> In the "Installation > Other tools" page, the Bitnami kubeprod project is listed as an alternative method to install cert-manager. However, that project has been archived two years ago and hasn't been updated since. See [announcement](https://github.com/vmware-archive/kube-prod-runtime/commit/a79bcda68a3423de01280c198f68ac05a1fa077f) for further details.

So this page is no longer needed in the newest version of the documentation.


content/docs/manifest.json
```diff
@@ -39,10 +39,6 @@
               "title": "OperatorHub (OLM)",
               "path": "/docs/installation/operator-lifecycle-manager.md"
             },
-            {
-              "title": "Other tools",
-              "path": "/docs/installation/other-tools.md"
-            },
```
SgtCoDFish, line 45: suggestion: Rather than entirely removing this page, could we maybe replace the kubeprod installation instructions with a notice saying that it's no longer supported?

If we remove this page it'll break any links that people have to the page and we try to avoid that!

(another approach is to add a redirect from the old older-tools page to a different page, but that might be confusing!)

SgtCoDFish (commented):
Hey thanks so much for raising this! Always appreciate updates to this kind of doc which doesn't get looked at much!

I've added a comment - what do you think? :grin:

content/docs/manifest.json
inteon: I think it is clear that the kube-prod-runtime project is archived & should not be used anymore.
I think it is good to simplify the side menu and remove this item.
For that reason, I would prefer just adding a redirect.

inteon:
Thanks @mprochowski we included your proposed change in https://github.com/cert-manager/website/pull/1311.

PR #1264, opened 2023-07-26
Add release notes for 1.12.1, 1.12.2 and 1.12.3

inteon:
These release notes were missing.
Based on https://github.com/cert-manager/cert-manager/releases

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -74,6 +74,7 @@ AzureDNS
 BKPR
 Bazel
 Bitnami
+Bugfixes
 BundleSource
 BundleTarget
 BundleCondition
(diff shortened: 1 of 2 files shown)


maelvls (approved):
Thanks for updating the change log. 

/lgtm

**nit:** the latest released version should come on top since that's what people should be looking at when they arrive on the 1.12 page: they should be aware of what the earliest version on the 1.12 branch is. This is what Kubernetes does: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md

PR #1272, opened 2023-08-18
Update supported releases about 1.13

SgtCoDFish:


SgtCoDFish:
As discussed with @inteon  and @wallrj !

PR #1278, opened 2023-09-03
Add clarification to selfsigned bootstrapping

Nickmman:
This adds further clarification when bootstrapping a cluster with `ClusterIssuer` and a Root CA `Certificate`. PR stemmed from cert-manager/cert-manager#6313

### DIFF
--- a/content/docs/configuration/selfsigned.md
+++ b/content/docs/configuration/selfsigned.md
@@ -136,7 +136,7 @@ apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
   name: my-selfsigned-ca
-  namespace: sandbox
+  namespace: cert-manager
 spec:
   isCA: true
   commonName: my-selfsigned-ca
--- a/content/docs/configuration/selfsigned.md
+++ b/content/docs/configuration/selfsigned.md
@@ -117,6 +117,48 @@ spec:
     secretName: root-secret
 ```
 
+Alternatively, if you are looking to use `ClusterIssuer` for signing `Certificates` anywhere in your cluster with the `SelfSigned` `Certificate` CA, use the YAML below (slight modification to the last step):
+
+```yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: sandbox
+---
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: selfsigned-issuer
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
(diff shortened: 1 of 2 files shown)


erikgb:
LGTM, but I think our PKI expert should take a look. 😄 

/cc @SgtCoDFish

content/docs/configuration/selfsigned.md
```diff
@@ -117,6 +117,48 @@ spec:
...
+metadata:
+  name: selfsigned-issuer
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: my-selfsigned-ca
+  namespace: sandbox
```
erikgb, line 139:
```suggestion
  namespace: cert-manager
```

I think I cluster issuer will look for the secret in the namespace where cert-manager is installed?

content/docs/configuration/selfsigned.md
```diff
@@ -117,6 +117,48 @@ spec:
...
+---
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: my-ca-issuer
+spec:
+  ca:
+    secretName: root-secret
+```
+The first `ClusterIssuer` is used to sign the Root CA, in this case. The second `ClusterIssuer` is used to sign certificates using said Root CA.
```
erikgb, line 160: Maybe better if we refer to the cluster issuers by name? WDYT?

erikgb (commented):
Would it be possible to explain your cluster issuer use-case without semi-duplicating the YAML? IMO this would be easier to understand by describing the differences in using an issuer vs cluster issuer.

content/docs/configuration/selfsigned.md
SgtCoDFish: I'd agree with this, it's a good suggestion!

SgtCoDFish (changes requested):
First of all, thanks for contributing - really great to see!

I don't have anything to add beyond what Erik already said. I think if his comments are addressed I'd gladly merge!

Nickmman:
@erikgb @SgtCoDFish Apologies for the delay. Thank you for taking the time to review and comment on this PR. I've accepted the suggestions and also pushed a new commit where I reword and clarify the excerpt regarding double ClusterIssuers.

SgtCoDFish (approved):
/lgtm
/approve

Looks like a great improvement, thank you so much for contributing 😁

SgtCoDFish:
Ah, you might need to add "selfsigned-issuer" to .spelling!

PR #1260, opened 2023-07-11
WIP: Clearer top-level menu layout

wallrj:
**Preview**: https://deploy-preview-1260--cert-manager-website.netlify.app/docs/

This is just to provide a preview of how the top-level menu of cert-manager.io could be simplified,
based on @inteon 's comment in https://github.com/cert-manager/website/pull/1075#pullrequestreview-1471417638

I am opening this PR to generate a website preview containing the desired menu layout, so that I can refer to that in other PRs where I make changes towards that end goal.

image

PR #1282, opened 2023-09-05
Rename the "Configuration" and "Usage" top-level menu items

inteon:
Based on what was proposed in https://github.com/cert-manager/website/pull/1260

### DIFF
--- a/content/docs/configuration/venafi.md
+++ b/content/docs/configuration/venafi.md
@@ -100,7 +100,7 @@ vaas-issuer    True    Venafi issuer started  2m
 You are now ready to issue certificates using the newly provisioned Venafi
 `Issuer` and Venafi as a Service.
 
-Read the [Issuing Certificates](../usage/certificate.md) document for
+Read the [Requesting Certificates](../usage/certificate.md) document for
 more information on how to create Certificate resources.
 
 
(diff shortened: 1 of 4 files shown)


wallrj (changes requested):
I personally still prefer "Issuing Certificates",
but if you're going to use "Requesting Certificates", then that should also be the title of the Overview / Introduction page for that section:
 * https://deploy-preview-1282--cert-manager-website.netlify.app/docs/usage/

And ideally, the URL path segment  should also change.

inteon:
> I personally still prefer "Issuing Certificates", but if you're going to use "Requesting Certificates", then that should also be the title of the Overview / Introduction page for that section:
> 
> * https://deploy-preview-1282--cert-manager-website.netlify.app/docs/usage/
> 
> And ideally, the URL path segment should also change.

I did not change the URL, because I think it this is the best way to not break any links.

inteon:
@wallrj I updated the content of those pages.

wallrj (approved):
Ok. As we discussed this morning, merging this will leave the website in an inconsistent state where the section called "Requesting Certificates" contains documents about prometheus metrics.
How about we merge this and the followup PR into a feature branch.
Then the feature branch preview can be presented in #cert-manager-dev for feedback before we make it live.

/lgtm
/hold in case you want to merge this into a feature branch

inteon:
I think it's not too disruptive, I fear that having a feature branch will just complicate things a lot (given that we might want to make more additional changes on master too).
/unhold

PR #1287, opened 2023-09-06
Use same method for creating on-page references links across all pages

inteon:
We often want to add a linkable on-page reference (eg. https://cert-manager.io/docs/installation/supported-releases/#bug-fixes-support). We are not consistent in the methods we use to create such links.

I think we can agree that creating an empty <a> tag before your title works best. This PR replaces all other methods with this method.


maelvls (approved):
/lgtm

PR #1289, opened 2023-09-07
Add diagrams to explain all the "requesting certificates" flows

inteon:
I split the diagram from https://cert-manager.io/docs/concepts/certificate/ into multiple diagrams and placed them on the different "requesting certificates" pages.


public/images/request-certificate-debug/gateway-shim-flow.svg
maelvls, line 1: Isn't it missing a `kind: Secret` under `certificateRefs`? Right now, it makes me think that `certificateRefs` targets a Certificate resource 😅

Unrelated: the latest version seems to be `gateway.networking.k8s.io/v1`, are we stuck with `v1alpha2`?

inteon: The docs are talking about `v1alpha2` still, I'm not sure what we support in the code.

content/docs/usage/certificaterequest.md
```diff
@@ -6,6 +6,10 @@ description: 'cert-manager core concepts: CertificateRequests'
 > **apiVersion:** cert-manager.io/v1  
 > **kind:** CertificateRequest
 
+<div style={{textAlign: "center"}}>
+<object data="/images/request-certificate-overview/request-certificate-certificaterequest.svg"></object>
+</div>
```
maelvls, line 11: I'm unsure about these diagrams. Unless the reader is comfortable with all the terms (CSI driver, etc), the diagram doesn't really help understanding. What idea are you trying to convey with these diagrams?

public/images/request-certificate-debug/gateway-shim-flow.svg
inteon: I added `kind: Secret`

content/docs/usage/certificaterequest.md
maelvls: IMO, this diagram should only show the "resources at play": the Gateway (with the principal annotation that enables the workflow) + the CertificateResource + the Secret. Anything else will confuse people.

public/images/request-certificate-overview/request-certificate-gateway.svg
maelvls, line 1: IMO, this diagram should only show the "resources at play": the Gateway (with the principal annotation that enables the workflow) + the CertificateResource + the Secret. Anything else will confuse people.

content/docs/usage/certificaterequest.md
inteon: It aims to show you what resources are created by the user & what resources are auto-generated for each of the different paths. (eg. If you annotate an Ingress, that will create a Certificate resource.

public/images/request-certificate-overview/request-certificate-gateway.svg
inteon: `Knowing what resources are at play is important since folks will need to know about them almost immediately to debug things.`
That is my goal for this diagram.
Do you know how we can adapt these diagrams to make that more clear?

maelvls: I would probably only show one box with a Gateway, one box with the CertificateResource, and one box with the Secret.

maelvls:
Thank you for digging into my old diagrams and for fixing them!! The new CR diagrams should also help.

/lgtm
/approve

PR #1293, opened 2023-09-12
Add bulletpoint for stackit webhook

PatrickKoss:
We in STACKIT have developed a webhook to work with the DNS01 ACME Issuer. We would like to be added as provider in the webhook section.


inteon:
/approve
/lgtm
Thanks for creating the webhook & sharing it with the world!

PR #1296, opened 2023-09-13
Add 1.13 release notes

inteon:
As described in https://cert-manager.io/docs/contributing/release-process/


SgtCoDFish (approved):
/lgtm
/approve

Looks like an obvious improvement. If there are any typos or whatever I missed, easy to fix those later! Thanks!

PR #1297, opened 2023-09-13
add Akeyless Issuer to External Issuers

akljph:


inteon:
Hi @akljph, thanks for the contribution & for reaching out via Slack.
I only just realised that this akeyless solution is compatible with the Vault API & that you are using our core Vault issuer.
So, I think I incorrectly advised you to add your issuer to this external-issuer section (because you are just using the core Vault issuer).

I propose the following instead: where we talk about Hashicorp Vault on the website, we can also mention Akeyless; but only where it makes sense & does not make the text more complex to read. Also, we could replace specific mentions of Hashicorp Vault with "Hashicorp Vault compatible APIs".
I'm not yet 100% sure how to fix this, but I think that as long it is useful to our users, we should also mention that there exist other platforms with compatible APIs.

Sorry for incorrectly asking you to add your platform to the external issuers list. I'm looking forward to how we can improve the website by also mentioning "Hashicorp Vault - compatible" solutions.

PR #1301, opened 2023-09-20
Update csi-driver project page

SgtCoDFish:
This page has a lot of prose which I think benefits from clarification. There are also many typos and the first section of our installation instructions talks about an ancient Kubernetes version, with instructions which will apply to approximately nobody.

This PR should hopefully make the docs simpler and easier to follow.

### DIFF
--- a/content/docs/projects/csi-driver.md
+++ b/content/docs/projects/csi-driver.md
@@ -12,15 +12,15 @@ directly into the pod, with no intermediate Secret being created.
 
 ## Why use csi-driver?
 
-- Ensure private keys never leave the node and are never sent over the network. All private keys are stored locally on the node
-- Unique key and certificate per application replica.
+- Ensure private keys never leave the node and are never sent over the network. Private keys are stored in memory, and shouldn't be written to disk.
+- Unique key and certificate per application replica
 - Fewer `Certificate` resources means writing less YAML
 - Keys and certificates are destroyed when an application terminates
 - No `Secret` resources needed for storing the certificate means less RBAC
 
 ## Installation
 
-You must have a working installation of cert-manager present on the cluster.
+You must have a working installation of cert-manager present on your cluster and be running at least Kubernetes v1.16.
 
 Instructions on how to install cert-manager can be found [on this website](../installation/README.md).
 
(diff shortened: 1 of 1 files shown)


content/docs/projects/csi-driver.md
```diff
@@ -1,42 +1,30 @@
...
+csi-driver is a [Container Storage Interface (CSI)](https://kubernetes-csi.github.io/docs/) driver plugin for Kubernetes
+which works alongside cert-manager.
+
+Pods which mount the cert-manager csi-driver will request certificates from cert-manager
+without needing a `Certificate` resource to be created. These certificates will be mounted
+directly into the pod, with no intermediate Secret being created.
+
+## Why use csi-driver?
+
+- Ensure private keys never leave the node and are never sent over the network. All private keys are stored locally on the node
```
inteon, line 15: I think we can say that these private keys do not hit the disk, they are kept in a memory-backed mount.

content/docs/projects/csi-driver.md
```diff
@@ -1,42 +1,30 @@
...
+- Unique key and certificate per application replica.
+- Fewer `Certificate` resources means writing less YAML
+- Keys and certificates are destroyed when an application terminates
+- No `Secret` resources needed for storing the certificate means less RBAC
+
```
inteon, line 20: Could we also talk about mTLS here? This example was removed from the intro.

content/docs/projects/csi-driver.md
```diff
@@ -80,17 +68,29 @@ items:
 ...
 ```
 
-The CSI driver is now installed and is ready to be used for pods in the cluster.
+### Kubernetes Requirements
```
inteon, line 71: I think this section is not very useful anymore, no-one should be running cert-manager 1.15/1.16 anymore (no public cloud provider still supports these versions).

content/docs/projects/csi-driver.md
SgtCoDFish: We could, but I'm not really sure why we call out mTLS at all. I can do mTLS just fine without csi-driver just like I can get a server cert just fine without csi-driver, right?

Is mentioning mTLS just to drop the keyword "mTLS"? Is csi-driver helping in some way I'm missing here?

inteon: Maybe we can state that this is a perfect solution for "ephemeral certificates" (eg. mTLS certificates).

I think it is important that we also clearly state that this is not a solution for long-lived certificates that should survive node restarts etc. (because each new pod instance will get a new cert).

SgtCoDFish: Good shout, yeah, I'll try to capture that!

inteon (approved):
Thanks @SgtCoDFish, this makes the page a lot easier to read!

PR #1312, opened 2023-10-02
Fix sidemenu collapse

inteon:
Fixes missing logic to unfold side menu when going to the link directly.

eg. if you go to https://deploy-preview-1312--cert-manager-website.netlify.app/docs/devops-tips/backup/, you will see that the side menu is expanded.

PR #1314, opened 2023-10-02
Mobile Menu: Partially revert #1312

inteon:
In #1312, I fixed the side menu which was not being unfolded when you directly go to a URL.
That PR removed the old "setSidebarCollapsed" logic, because it was not clear what it was used for ( I thought it was broken logic that was meant to unfold the menu when you go to a URL directly ).

Now, I found that the "setSidebarCollapsed" is for the mobile menu, it collapses the menu when you click a link that takes you to a new page.
This PR partially reverts the removal of "setSidebarCollapsed" that was done in #1312, making the sidemenu work on desktop and mobile & also keeps the sidemenu unfolding when you go to a URL directly.

### DIFF
--- a/components/docs/Sidebar/Dropdown.jsx
+++ b/components/docs/Sidebar/Dropdown.jsx
@@ -6,6 +6,7 @@ import SidebarLink from './SidebarLink'
 export default function Dropdown({
   routes,
   parentOpen = true,
+  setSidebarCollapsed,
   setParentOpen
 }) {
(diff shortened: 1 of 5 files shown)


wallrj (approved):
Thanks @inteon 

I tried the current site in Microsoft Edge with a narrow window and recreated the problem you described, that the menu stays on top after you click a link, obscuring the new page content.
I then tried the preview of this PR and observed that the menu collapses after I click on one of the links.

/lgtm

PR #1329, opened 2023-10-18
Add overview diagrams for "Requesting Certificates" section

inteon:
Quick overview diagrams on each of the "Requesting Certificates" pages.

https://deploy-preview-1329--cert-manager-website.netlify.app/docs/usage/
image

https://deploy-preview-1329--cert-manager-website.netlify.app/docs/usage/certificate/
image

https://deploy-preview-1329--cert-manager-website.netlify.app/docs/usage/ingress/
image


maelvls:
Nice. The unrelated boxes (CSI driver etc) are now greyed out. It is much less confusing. The diagram now clearly shows a Certificate in the center: that's great because this page is about the Certificate resource.

As a first-time reader of this page, I would probably still skip the diagram entirely because this diagram because it looks confusing:

1. The top of the page talks about the Certificate and Issuer resources, but the Issuer resource doesn't show on the diagram; that's surprising.
2. Do I need to create an "Annotated Ingress" to create a Certificate? That's what the diagram alludes to.
3. Although I know what is an Ingress, what is an "Annotated Ingress"?
4. Since I didn't know what was an Annotated Ingress, I did ctrl+F but didn't find any reference to "Annotated Ingress".
5. I know about Istio's Gateway resource, but what is an "Annotated Gateway"?

One way of making it much clearer would be to write a legend to this diagram to explain what the reader should understand from the diagram

inteon:
> Nice. The unrelated boxes (CSI driver etc) are now greyed out. It is much less confusing. The diagram now clearly shows a Certificate in the center: that's great because this page is about the Certificate resource.
> 
> As a first-time reader of this page, I would probably still skip the diagram entirely because this diagram because it looks confusing:
> 
> 1. The top of the page talks about the Certificate and Issuer resources, but the Issuer resource doesn't show on the diagram; that's surprising.
> 2. Do I need to create an "Annotated Ingress" to create a Certificate? That's what the diagram alludes to.
> 3. Although I know what is an Ingress, what is an "Annotated Ingress"?
> 4. Since I didn't know what was an Annotated Ingress, I did ctrl+F but didn't find any reference to "Annotated Ingress".
> 5. I know about Istio's Gateway resource, but what is an "Annotated Gateway"?
> 
> One way of making it much clearer would be to write a legend to this diagram to explain what the reader should understand from the diagram

1. The diagram does not really apply to issuers, each resource references the issuer resource (annotated Ingresses, Certificates, CertificateRequests, ...). So I did not include that to not overwhelm the diagram.
2. There are multiple ways to create a Certificate resource, I updated the diagram with a set of letters (indicating the different paths/ options) and a legend that explains what the different arrows mean.
3&4&5. On the Ingress and Gateway page, we talk about adding annotations to your ingress resource, so I think this should be clear to the user?

maelvls:
This is an improvement over the existing page, so let's go with that!

My concerns are:

1. The grey stuff isn't needed to understand each page. It makes things confusing.
2. The "a", "b", "c" on the arrows should say the actual action (e.g., "creates").
4. It's not clear which resources are created by cert-manager, "automated action" doesn't really say that; using a line separation between "created by the user" and "created by cert-manager" would help.
5. The "Secret" should show on the diagram too since the point is to explain what Kubernetes resources are at play. I would even show the Issuer, but that's less important.

/lgtm
/approve

PR #1330, opened 2023-10-18
Write about best-practice high availability and scaling of cert-manager components

wallrj:
**Preview**: https://deploy-preview-1330--cert-manager-website.netlify.app/docs/installation/best-practice/#high-availability

We've added various Helm chart values which allow users to configure cert-manager for HA and scalability, 
but we've never documented any recommendations explaining how these settings should be used in production.

My original / ultimate plan was to change some of the Helm chart defaults to include useful default topology constraints,
as first suggested by @ThatsMrTalbot in a cert-manager dev meeting  and discussed further in https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1697561450000799

But first I want to document what we think the best practice settings are.

### DIFF
--- a/content/docs/installation/best-practice.md
+++ b/content/docs/installation/best-practice.md
@@ -71,6 +71,9 @@ So it is *especially* important to keep at multiple replicas of the cert-manager
 > 📖 Read [Ensure control plane stability when using webhooks](https://cloud.google.com/kubernetes-engine/docs/how-to/optimize-webhooks)
 > in the Google Kubernetes Engine (GKE) documentation,
 > for examples of how webhook disruptions might disrupt your cluster.
+>
+> 📖 Read [The dark side of Kubernetes admission webhooks](https://techblog.cisco.com/blog/dark-side-of-kubernetes-admission-webhooks)
+> on the Cisco Tech Blog, to learn more about potential issues caused by webhooks and how you can avoid them.
 
 ### Topology Spread Constraints
 


content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+### PodDisruptionBudget
+
+For high availability you should also deploy a `PodDisruptionBudget` resource,
+with `minAvailable=1` *or* with `maxUnavailable=1`.
+This ensures that a *voluntary* disruption, such as the draining of a Node, can not proceed
+until at least one other replica has been successfully scheduled and started on another Node.
+The Helm chart has parameters to enable and configure a PodDisruptionBudget
+for each of the long-running cert-manager components.
+
+
```
hawksight, line 136: Perhaps include the relevant subset of YAML to configure PDBs? As you have done with the topology settings?

```yaml
podDisruptionBudget:
  enabled: true
  minAvailable: 1
webhook:
  podDisruptionBudget:
    enabled: true
    minAvailable: 1
cainjector:
  podDisruptionBudget:
    enabled: true
    minAvailable: 1
```

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+so that each component can make efficient use of the CPU cores assigned to it.
+
+By default the controller [uses 5 workers per controller](https://github.com/cert-manager/cert-manager/blob/3b0a5cec4140e92ba12f3eace362a4bd65fbb30e/cmd/controller/app/options/options.go#L180-L181)
+which should be adequate for most clusters.
+You can change the number of workers using the `--concurrent-workers` [flag of the controller manager](../cli/controller.md).
+
+The cainjector uses a single worker for each of the resource types that it reconciles,
+
+> ⚠️ This this is not yet configurable.
+
```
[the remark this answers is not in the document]
hawksight: I think you are right in that it's not "best practice" to change this setting. It's more of an operational concern at a considerable scale. So I'd :+1: removing it from this document.

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
 and in that case you can modify the installation configuration using Helm chart values to override the defaults.
 
+## High Availability
+
+cert-manager has three long-running components: controller, cainjector, and webhook.
+Each of these components has a Deployment and by default each Deployment has 1 replica
+but this does not provide high availability.
+The Helm chart for cert-manager has parameters to configure the `replicaCount` for each Deployment,
+and in production you should use 2 or more replicas to achieve high availability.
+
```
hawksight, line 32:
```yaml
replicaCount: 2
webhook:
  replicaCount: 3
cainjector:
  replicaCount: 2
```

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+#### Memory
+
+Use vertical scaling to assign sufficient memory resources to these components.
+The memory requirements will be higher on clusters with very many API resources or with large API resources.
+This is because each of the components reconciles one or more Kubernetes API resources,
+and each component will cache the metadata and sometimes the entire resource in memory,
+so as to reduce the load on the Kubernetes API server.
+
+For example, if the cluster contains very many CertificateRequest resources,
```
hawksight, line 159: Perhaps a use case is more illustrative. Eg,

```suggestion
If your cluster contains a high volume of `CertificateRequest` resources such as when using many ephemeral or short lived certificate rotates frequently,
```

Unless I have missed the point here.

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+you will need to increase the memory limit of the controller Pod.
```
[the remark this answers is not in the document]
hawksight: Is it best practice to limit this to only the installation namespace?

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+      matchLabels:
+        app.kubernetes.io/instance: cert-manager
+        app.kubernetes.io/component: webhook
+  - maxSkew: 1
+    topologyKey: kubernetes.io/hostname
+    whenUnsatisfiable: ScheduleAnyway
+    labelSelector:
+      matchLabels:
+        app.kubernetes.io/instance: cert-manager
+        app.kubernetes.io/component: webhook
```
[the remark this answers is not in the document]
hawksight: I'm easy here, not having it is less YAML.
Having it means it's explicit.

hawksight (approved):
Very well written, just a few minor notes to break up the text a little.

For reference, I just checked for the labels on my GKE nodes:
```
kubernetes.io/hostname=gke-demo-istio-5d390798-5v7i,kubernetes.io/os=linux,node.kubernetes.io/instance-type=e2-medium,topology.gke.io/zone=europe-west1-b,topology.kubernetes.io/region=europe-west1,topology.kubernetes.io/zone=europe-west1-b
```

So the labels chosen looks appropriate, although I have not checked any other managed offering.

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+### controller and cainjector
+
+The controller and cainjector components use [leader election](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection)
+to ensure that only one replica is active.
+This prevents conflicts which would arise if multiple replicas were reconciling the same API resources.
+So in these components you can use multiple replicas to achieve high availability but not for horizontal scaling.
+
+Use two replicas to ensures that there is a standby Pod scheduled to a Node which is ready to take leadership,
+should the current leader encounter a disruption.
+For example, if the Node on which the leader Pod is running is drained.
```
jsoref, line 42:
```suggestion
For example, when the leader Pod is drained from its node.
```

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+Or, if the leader Pod encounters an unexpected deadlock.
+
+There is little justification for using more than 2 replicas of these components.
+Further replicas *may* add a degree of resilience
+if you have the luxury of sufficient Nodes
+with sufficient CPU and memory to accommodate the standby replicas.
```
jsoref, line 48:
```suggestion
with sufficient CPU and memory to accommodate additional standby replicas.
```

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+
+### webhook
+
+By default the cert-manager webhook Deployment has 1 replica, but in production you should use 3 or more.
+If the cert-manager webhook is unavailable, all API operations on cert-manager custom resources will fail,
+and this will disrupt any software that creates, updates or deletes cert-manager custom resources.
```
jsoref, line 54: It might also generally disrupt the cluster -- certainly GKE yelled at me about this, I think.

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+      matchLabels:
+        app.kubernetes.io/instance: cert-manager
+        app.kubernetes.io/component: cainjector
+```
+
+### PodDisruptionBudget
+
+For high availability you should also deploy a `PodDisruptionBudget` resource,
+with `minAvailable=1` *or* with `maxUnavailable=1`.
+This ensures that a *voluntary* disruption, such as the draining of a Node, can not proceed
```
jsoref, line 131:
```suggestion
This ensures that a *voluntary* disruption, such as the draining of a Node, cannot proceed
```

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,174 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+
+cert-manager has three long-running components: controller, cainjector, and webhook.
+The Helm chart does not include resource requests and limits for any of these,
+so you should supply resource requests and limits which are appropriate for your cluster.
+
+### controller and cainjector
+
+The controller and cainjector components use leader election to ensure that only one replica is active.
+This prevents conflicts which would arise if multiple replicas were reconciling the same API resources.
+You can not use horizontal scaling for these components.
```
jsoref, line 148:
```suggestion
You cannot use horizontal scaling for these components.
```

content/docs/installation/best-practice.md
wallrj: Done.

content/docs/installation/best-practice.md
wallrj: Done.

content/docs/installation/best-practice.md
wallrj: Done.

content/docs/installation/best-practice.md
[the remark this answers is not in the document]
wallrj: I suppose it's not technically "best practice" but I've added it anyway. 
See what you think.

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,201 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+cert-manager has three long-running components: controller, cainjector, and webhook.
+Each of these components has a Deployment and by default each Deployment has 1 replica
+but this does not provide high availability.
+The Helm chart for cert-manager has parameters to configure the `replicaCount` for each Deployment.
+In production we recommend the following `replicaCount` parameters:
+
+```yaml
+replicaCount: 2
+webhook:
+  replicaCount: 3
```
inteon, line 36: I think a user might want even more webhook replicas, maybe we should explain why we set the replicas for leader-elected controllers to 2 and for non-leader elected controllers to >= 3?

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,201 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+There is little justification for using more than 2 replicas of these components.
+Further replicas *may* add a degree of resilience
+if you have the luxury of sufficient Nodes
+with sufficient CPU and memory to accommodate additional standby replicas.
+
+### webhook
+
+By default the cert-manager webhook Deployment has 1 replica, but in production you should use 3 or more.
+If the cert-manager webhook is unavailable, all API operations on cert-manager custom resources will fail,
+and this will disrupt any software that creates, updates or deletes cert-manager custom resources,
```
inteon, line 62:
```suggestion
and this will disrupt any software that creates, updates or deletes cert-manager custom resources (including cert-manager itself),
```

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,201 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+
+Use two replicas to ensures that there is a standby Pod scheduled to a Node which is ready to take leadership,
+should the current leader encounter a disruption.
+For example, when the leader Pod is drained from its node.
+Or, if the leader Pod encounters an unexpected deadlock.
+
+There is little justification for using more than 2 replicas of these components.
+Further replicas *may* add a degree of resilience
+if you have the luxury of sufficient Nodes
+with sufficient CPU and memory to accommodate additional standby replicas.
```
inteon, line 56:
```suggestion
```

I don't think there is a real reason here, we are basically saying that you can run more than 2 but only if you like to waste compute resources.

content/docs/installation/best-practice.md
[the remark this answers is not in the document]
inteon: I think if we can verify that this works by default, we should not tell people to configure this (because we want to keep our docs as simple as possible). Instead, we can just link to the documentation you shared and say that this works correctly by default.

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,201 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+    labelSelector:
+      matchLabels:
+        app.kubernetes.io/instance: cert-manager
+        app.kubernetes.io/component: cainjector
+```
+
+### PodDisruptionBudget
+
+For high availability you should also deploy a `PodDisruptionBudget` resource,
+with `minAvailable=1` *or* with `maxUnavailable=1`.
```
inteon, line 141: I'm not sure about `maxUnavailable=1`, what happens if you only have 1 replica running? Is it OK to remove that replica in that situation?
Also, if you have a lot of replicas, `maxUnavailable > 1` might be ok too.

jsoref: `maxUnavailable: 1` requires you to have 2 running, otherwise you're stuck. Speaking from experience.

content/docs/installation/best-practice.md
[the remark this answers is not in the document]
wallrj: I agree that according to the docs the desired anti-affinity scheduling should happen by default.
I've updated this paragraph.

content/docs/installation/best-practice.md
wallrj: It would help if you were unlucky enough to simultaneously lose two data centers or if you lose  two nodes of a single zone cluster...but I think might remove it for the sake of making the document shorter and easy to digest.

content/docs/installation/best-practice.md
wallrj: Both settings require at least 2 replicas which is why Kyverno omit the PodDisruptionBudget if Replicas < 2

https://github.com/kyverno/kyverno/blob/53eab3d777d3dc33a1cfac73ea069c910e4b3f45/charts/kyverno/templates/background-controller/poddisruptionbudget.yaml#L1-L16

I've added a warning note about this.

I think `maxUnavailable > 1` might ok for the webhook if you have 3 replicas and you want to upgrade Nodes 2 at a time.
I'll just recommend `minAvailable=1` for the sake of simplicity.

PR #1331, opened 2023-10-20
Explain why and how to isolate the cert-manager workloads

wallrj:
Followup to #1330 

In this PR I want to give an example of how to use the affinity and toleration Helm values 
and I propose running the cert-manager Pods on dedicated "platform" nodes,
for security reasons, but there may be other good use cases.

@erikgb Please take a look


content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,109 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+```
+
+> 📖 Read more about [Taints and Tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)
+> in the Kubernetes documentation.
+>
+> 📖 Read the [Guide to isolating tenant workloads to specific nodes](https://aws.github.io/aws-eks-best-practices/security/docs/multitenancy/#isolating-tenant-workloads-to-specific-nodes)
+> in the EKS Best Practice Guides,
+> for an in-depth explanation of these techniques.
+>
+> 📖 Learn how to [Isolate your workloads in dedicated node pools](https://cloud.google.com/kubernetes-engine/docs/how-to/isolate-workloads-dedicated-nodes) on Google Kubernetes Engine.
```
jsoref, line 115: i'm half inclined to have `Kubernetes documentation`, `EKS Best Practice Guides`, and `Google Kubernetes Engine [Documentation]` be links as well...

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,109 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+>
+> 📖 Read more about the [`node-restriction.kubernetes.io/` prefix and the `NodeRestriction` admission plugin](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction).
+>
+> ℹ️ On a multi-tenant cluster,
+> consider enabling the [`PodTolerationRestriction` plugin](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podtolerationrestriction)
+> to limit which tolerations tenants may add to their Pods.
+> You may also use that plugin to add default tolerations to the `cert-manager` namespace,
+> which obviates the need to explicitly set the tolerations in the Helm chart.
+>
+> ℹ️ Alternatively, you could use Kyverno to limit which tolerations Pods are allowed to use.
```
jsoref, line 125: I' want `Kyverno` to be a link

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,109 @@ are designed for backwards compatibility rather than for best practice or maximu
 You may find that the default resources do not comply with the security policy on your Kubernetes cluster
 and in that case you can modify the installation configuration using Helm chart values to override the defaults.
 
+## Isolate cert-manager on dedicated node pools
+
+cert-manager is a cluster scoped operator and you should treat it as part of your platform control plane.
```
jsoref, line 27: perhaps?
```suggestion
cert-manager is a cluster scoped operator and you should treat it as part of your platform's control plane.
```

this depends on what you mean. If you mean `Platform Control Plane`, then I'd probably write it as such instead...

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,109 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+
+The cert-manager controller caches all the Secret resources of the cluster in memory,
```
jsoref, line 29: This is kinda weird and is definitely surprising.

Can it at least get an emoji/callout?

Also, is there a reason it can't only cache `Certificate` `Secrets` / `Secrets` it creates/manages?

erikgb (commented):
Thanks for working on this, @wallrj!

We simply use https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector to schedule cert-manager workloads to dedicated platform nodes. I think that should be included as the simplest way of achieving the desired goal. The examples you have put up are simpler to express with `nodeselector`, I think?

content/docs/installation/best-practice.md
wallrj: You're right, cert-manager [since release 1.12 does cache fewer Secrets](https://cert-manager.io/docs/releases/release-notes/release-notes-1.12/#lower-memory-footprint).
I've updated the words.

wallrj:
> We simply use https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector to schedule cert-manager workloads to dedicated platform nodes. I think that should be included as the simplest way of achieving the desired goal. The examples you have put up are simpler to express with `nodeselector`, I think?

Done. I agree that nodeSelector is much simpler and works the same as nodeAffinity so I've changed it.
I had to explain that there is a default OS nodeSelector which you must explicitly add to your values.

erikgb (commented):
I am not an expert in this area, but why do we need tolerations in addition to the nodeSelector? Looks like a duplication to me, and according to https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector, just the nodeSelector should be sufficient. And this setup works well for us.

wallrj:
> I am not an expert in this area, but why do we need tolerations in addition to the nodeSelector? Looks like a duplication to me, and according to https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector, just the nodeSelector should be sufficient. And this setup works well for us.

I'm not an expert either, but I guess one reason is:
*  in a multi-tenant cluster, 
*  **non-platform** Pods might not have `nodeSelector` or `affinity.nodeAffinity` settings,
*  so those could be scheduled to any Nodes, including those that you reserved for your platform's Pods.
* You can prevent this by adding a taint to the platform Nodes.

How do you prevent this happening in your cluster?
Do you have an admission webhook that overrides `nodeSelector` of the Pods of your tenants?

hawksight (approved):
A really useful addition. Had a quick read and looks good to me and seems better to have this live now and correct anything later if needed.

gintautassulskus:
Great contribution, @wallrj! The documentation sounds correct. A minor remark, I do not recognise `cainjection` and `startupapicheck` used in the example, but it's been a long time since I used this package.

erikgb (approved):
/lgtm

Very nice addition to the docs, @wallrj

erikgb:
> How do you prevent this happening in your cluster? Do you have an admission webhook that overrides `nodeSelector` of the Pods of your tenants?

No idea, but I can check. 😉 We run on Openshift, and it's usually pretty "secure by default".

erikgb:
> > How do you prevent this happening in your cluster? Do you have an admission webhook that overrides `nodeSelector` of the Pods of your tenants?
> 
> No idea, but I can check. 😉 We run on Openshift, and it's usually pretty "secure by default".

I think this is handled by some Openshift "magic" described [here](https://docs.openshift.com/container-platform/4.13/nodes/scheduling/nodes-scheduler-node-selectors.html#nodes-scheduler-node-selectors-cluster_nodes-scheduler-node-selectors). When a normal user, without access to namespace resources, schedules a workload, the pod will always get a "worker" label added to the pod nodeSelector. Since none of our nodes matches worker + something else, it means that all end-user workloads will be scheduled on worker nodes. Or not scheduled at all - if a user tries to set `nodeSelector`.

wallrj:
> I think this is handled by some Openshift "magic" described [here](https://docs.openshift.com/container-platform/4.13/nodes/scheduling/nodes-scheduler-node-selectors.html#nodes-scheduler-node-selectors-cluster_nodes-scheduler-node-selectors). 

@erikgb Thanks so much for digging into that! I've added a link to that doc.

content/docs/installation/best-practice.md
```diff
@@ -22,6 +22,109 @@ are designed for backwards compatibility rather than for best practice or maximu
...
+> ℹ️ This example uses `nodeSelector` to *place* the Pods but you could also use `affinity.nodeAffinity`.
+> `nodeSelector` is chosen here because it has a simpler syntax.
+> Read [Assigning Pods to Nodes](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) to learn more.
+>
+> ℹ️ The default `nodeSelector` value `kubernetes.io/os: linux` [avoids placing cert-manager Pods on Windows nodes in a mixed OS cluster](https://github.com/cert-manager/cert-manager/pull/3605),
+> so that must be explicitly included here too.
+>
+> 📖 Read more about [Taints and Tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)
+> in the [Kubernetes documentation](https://kubernetes.io/docs/).
+>
```
schelv, line 108: It would be more convenient for readers if this link were introduced earlier, right when these concepts are first mentioned.

SgtCoDFish (approved):
/lgtm
/approve
/hold

There's a reasonable looking comment just been added, so I've added a hold if you want to incorporate that suggestion. Ping me for a re-review if needed!

SgtCoDFish (approved):
/lgtm
/approve

PR #1333, opened 2023-10-25
Fix title casing in sidebar

SgtCoDFish:
I find the lowercase `u` really stands out in a bad way - that's my personal opinion, but I think it's worth us being consistent in our titles and most of our other titles start with a capital except where it doesn't make sense (e.g. `kubectl apply`).


wallrj (approved):
/lgtm

PR #1335, opened 2023-10-30
Update documentation to point to latest patch version

inteon:
Following the [Release Process](https://cert-manager.io/docs/contributing/release-process/).

### DIFF
--- a/content/docs/cli/controller.md
+++ b/content/docs/cli/controller.md
@@ -14,7 +14,7 @@ Usage:
   controller [flags]
 
 Flags:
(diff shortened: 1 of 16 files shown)


SgtCoDFish (changes requested):
nitpick: This doesn't seem to update the patch version for release-1.12 in content/v1.12-docs - I think it'd be worth bumping those while we're at it?

SgtCoDFish (approved):
/lgtm
/approve

Cheers!

PR #1337, opened 2023-11-01
Update README.md to reflect that the trust namespace is only for secrets

tspearconquest:


content/docs/trust/trust-manager/README.md
```diff
@@ -157,8 +157,8 @@ we currently have available.
 One of the more important configuration options you might need to consider at install time is which "trust namespace" to use,
 which can be set via the Helm value `app.trust.namespace`.
 
-The trust namespace is the only one in which `Secret` and `ConfigMap` sources can be read. This restriction is in place
-for security reasons - we don't want to give trust-manager the permission to read all `Secret`s or `ConfigMap`s in all namespaces.
+The trust namespace is the only one in which `Secret` sources can be read. This restriction is in place
```
JoshVanL, line 160:
```suggestion
By default, the trust namespace is the only one in which `Secret`s can be read. This restriction is in place
```

JoshVanL: "By default" because trust-manager can be put into a mode where Secrets can be used as a target. https://github.com/cert-manager/trust-manager/blob/ad150e553b943a507949d183a3e71e869a2c5d83/deploy/charts/trust-manager/templates/clusterrole.yaml#L59

JoshVanL (changes requested):
Cheers!

content/docs/trust/trust-manager/README.md
tspearconquest: Hmm, I'm not sure I like this wording as much, because it's not quite clear enough. I have another suggestion, will post momentarily.

content/docs/trust/trust-manager/README.md
```diff
@@ -157,8 +157,8 @@ we currently have available.
 One of the more important configuration options you might need to consider at install time is which "trust namespace" to use,
 which can be set via the Helm value `app.trust.namespace`.
 
-The trust namespace is the only one in which `Secret` and `ConfigMap` sources can be read. This restriction is in place
-for security reasons - we don't want to give trust-manager the permission to read all `Secret`s or `ConfigMap`s in all namespaces.
+By default, the trust namespace is the only one in which `Secret`s can be read. This restriction is in place
```
tspearconquest, line 160:
```suggestion
By default, the trust namespace is the only namespace where`Secret`s will be read. This restriction is in place
```

content/docs/trust/trust-manager/README.md
```diff
@@ -157,8 +157,8 @@ we currently have available.
...
+for security reasons - we don't want to give trust-manager the permission to read all `Secret`s in all namespaces.
```
tspearconquest, line 161:
```suggestion
for security reasons - we don't want to give trust-manager the permission to read all `Secret`s in all namespaces. With additional configuration, secrets may be read from or written to other namespaces.
```

JoshVanL:
Your new suggestions look good @tspearconquest! Please commit them yourself- If I sign off on the suggestion then I also become an author and then don't believe I can merge :slightly_smiling_face:

PR #1360, opened 2023-12-08
Verify Mastodon social media link

JamesLaverack:
Adding the rel="me" attribute to a Mastodon link causes Mastodon to treat the link as 'verified', and will display a green check next to the website link in the cert-manager Mastodon profile. See https://docs.joinmastodon.org/user/profile/#verification for more details on this feature.

### DIFF
--- a/components/Header.jsx
+++ b/components/Header.jsx
@@ -47,7 +47,7 @@ export default function Header() {
           target="_blank"
           icon="mastodon"
           iconWClass="w-4"
-	  rel="me"
+           rel="me"
           className="px-3 py-1 text-11px mx-4 "
         />
       </div>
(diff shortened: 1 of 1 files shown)


JamesLaverack:
I'm not actually 100% on if this works on a `<button>` element actually. 🤔

components/Header.jsx
```diff
@@ -47,6 +47,7 @@ export default function Header() {
           target="_blank"
           icon="mastodon"
           iconWClass="w-4"
+	  rel="me"
```
maelvls, line 50:
```suggestion
           rel="me"
```

maelvls:
Hi James! Nice you see you there! let's make the change and see if it works 😇

/lgtm
/approve

components/Header.jsx
JamesLaverack: Whoops. 😅

PR #1369, opened 2023-12-19
adding new version of adcs-issuer

djkormo:
Adding new version of adcs-issuer


inteon (changes requested):
Thank you for maintaining this issuer!
I'm also very happy that we can move this issuer to tier 2.
@djkormo would it be possible to order the issuers alphabetically within tier 2?

djkormo:
Added order the issuers alphabetically within tier

inteon (approved):
Thanks a lot for maintaining and sharing this issuer.

/approve
/lgtm

PR #1371, opened 2023-12-20
Update CLI reference

inteon:
Adds newly added flags and removes hidden flags.


SgtCoDFish (commented):
How was this generated? This doesn't match the changes from `./scripts/gendocs/generate-new-import-path-docs` when run locally.

This is obviously an improvement, but my recollection was that this was done by a script!

inteon:
> How was this generated? This doesn't match the changes from `./scripts/gendocs/generate-new-import-path-docs` when run locally.
> 
> This is obviously an improvement, but my recollection was that this was done by a script!

Thanks for pointing me to that script, I did not realise it existed.
I updated some issues that caused the script to fail for me & pointed the script to master, so it is up-to-date for release-next.

SgtCoDFish (approved):
/lgtm
/approve

Cheers!

PR #1373, opened 2023-12-21
Venafi Issuer: Explain why "Do not select 'refresh token enabled'"

maelvls:
A while back, when the "Access Token" authentication method was documented by Richard in https://github.com/cert-manager/website/pull/346, the following was added:

> NOTE: Do not select "Refresh Token Enabled" and set a *long* "Token Validity (days)".

I have read a few questions about "why is cert-manager not supporting refresh tokens?", so I figured I would explain why in this PR.


wallrj (approved):
/lgtm

PR #1376, opened 2024-01-04
feat: add documentation for using TLS on the metric endpoint

ThatsMrTalbot:


content/docs/devops-tips/prometheus-metrics.md
```diff
@@ -64,6 +64,48 @@ spec:
...
+kind: ControllerConfiguration
+metricsTLSConfig:
+  dynamic:
+    secretNamespace: "cert-manager"
+    secretName: "cert-manager-metrics-ca"
+    dnsNames:
+    - cert-manager-metrics
+    - cert-manager-metrics.cert-manager
+    - cert-manager-metrics.cert-manager.svc
+```
```
wallrj, line 107: This worked, but how should I configure my client with the CA certificate?
The problem is that the Secret resource contains the CA certificate **and** the private key so I don't really want to mount that Secret into my prometheus client Pod.

I could implement some other script that copies the CA certificate to another ConfigMap, but the CA is only valid for 1 year, so I'd need to schedule the script to run at the time the CA certificate gets renewed.

Also, I'm unsure about whether these DNS names are useful.
We recommend users to run two replicas of the controller, and 
I suppose that my prometheus client needs to collect metrics from each of the replicas.
These names would resolve to the loadbalancer service IP which would round robin between the Pods.

Seems like the IP addresses of the Pods should also be added to the serving certificate too, 
so that the prometheus client can connect to the Pod by IP address.
Is that how Prometheus will operate if using a PodMonitor resource (as recommended in the section above this)?

content/docs/devops-tips/prometheus-metrics.md
```diff
@@ -64,6 +64,48 @@ spec:
...
+An example config file would be:
+
+```yaml
+apiVersion: controller.config.cert-manager.io/v1alpha1
+kind: ControllerConfiguration
+metricsTLSConfig:
+  filesystem:
+    certFile: "/path/to/cert.pem"
+    keyFile: "/path/to/key.pem"
+```
```
wallrj, line 86: I haven't tried this yet, but how would you get these files into the Pod filesystem?
Using the `.volumes` and `.mounts` Helm values?

And if so, how would the user manage those Certifcates and keys? 
It seems like the answer should be to use a cert-manager itself to create a Secret and mount that, 
but there will be a bootstrapping problem.

So maybe  we should explain that here in the docs.

wallrj (changes requested):
I've tried it out, but I'm unsure about how easy it is going to be to set up Prometheus with the CA certificate it needs to connect to the TLS metrics endpoint.

Let's talk about it more in the standup tomorrow.

content/docs/devops-tips/prometheus-metrics.md
ThatsMrTalbot: If you are using prometheus operator, you can point reference a Secret as part of a PodMonitor or ServiceMonitor object https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.TLSConfig

The way Prometheus operator implements this is to copy the referenced secret keys into the Prometheus secret, since only the keys specified are copied, the private key would not be present inside the prometheus pod without a mis-configuration (see https://github.com/prometheus-operator/prometheus-operator/blob/547fea903e3df86272573e20c3b1479d0557a1a3/pkg/assets/store.go for implementation).

PodMonitors also let you set the serverName you wish to use when validating, which means you don't strictly _need_ the IP address, but I can also see the benefit in including it in the cert

content/docs/devops-tips/prometheus-metrics.md
ThatsMrTalbot: You can indeed pre-create them and use volumes and mounts to add to the filesystem.

I also worked out that if you can use cert-manager itself without having a bootstrap problem (kind of, the metrics endpoint will not work for a short while)

In Kubernetes you can mark a secret volume as optional, this means the pod is not blocked from starting. Then cert manager can be used to provision the secret, kubelet would eventually notice the secret now exists and mount it in the correct place where cert manager would pick it up

ThatsMrTalbot: I've added a note that the certificate and key must be mounted into the pod

content/docs/devops-tips/prometheus-metrics.md
ThatsMrTalbot: I've added an example on how PodMonitor can be used to trust the CA and validate the serverName

wallrj: Thanks for adding that example and thanks for helping me get prometheus set up to test this.

I finally got it working and can see the cert-manager targets in the Prometheus Web UI,
using HTTPS URLs.

image

In that case I patched the `ServiceMonitor` resource which is installed by the Helm chart:

```diff
$ git diff
diff --git a/service-monitor.live.yaml b/service-monitor.live.yaml
index b77ffb52f..1aa6a3992 100644
--- a/service-monitor.live.yaml
+++ b/service-monitor.live.yaml
@@ -22,5 +22,12 @@ spec:
   - targetPort: 9402
     path: /metrics
     interval: 60s
+    scheme: https
     scrapeTimeout: 30s
     honorLabels: false
+    tlsConfig:
+      ca:
+        secret:
+          key: tls.crt
+          name: cert-manager-metrics-ca
+      serverName: cert-manager-metrics
```

content/docs/devops-tips/prometheus-metrics.md
```diff
@@ -64,6 +64,83 @@ spec:
...
+spec:
+  jobLabel: app.kubernetes.io/name
+  selector:
+    matchLabels:
+      app: cert-manager
+      app.kubernetes.io/name: cert-manager
+      app.kubernetes.io/instance: cert-manager
+      app.kubernetes.io/component: "controller"
+  podMetricsEndpoints:
+    - port: http
```
wallrj, line 133: I think this port name  should be "http-metrics", here and in the example above.

```sh
$ kubectl -n cert-manager get deploy cert-manager -o yaml | grep -A6 ports
        ports:
        - containerPort: 9402
          name: http-metrics
          protocol: TCP
        - containerPort: 9403
          name: http-healthz
          protocol: TCP
```

content/docs/devops-tips/prometheus-metrics.md
```diff
@@ -64,6 +64,83 @@ spec:
...
+      honorLabels: true
```
wallrj, line 134: I also added `scheme: https` because the API docs say that the default is `http`....but perhaps the default changes based on the existence of a `tlsConfig` stanza?

wallrj (changes requested):
Please also add a note about this new feature to the release notes, linking to this new section for more information:
 * https://github.com/cert-manager/website/blob/release-next/content/docs/releases/release-notes/release-notes-1.14.md#other

content/docs/devops-tips/prometheus-metrics.md
ThatsMrTalbot: Yea, thats what the helm chart names the port. I've updated docs to reflect this.

wallrj (approved):
/lgtm
/approve

I'll add the release note when I do next alpha.

Sebhansson:
Should this not be namned https-metrics in the docs? :)

https://github.com/cert-manager/website/blob/master/content/docs/devops-tips/prometheus-metrics.md?plain=1#L37

PR #1386, opened 2024-01-12
Add FAQ entry on ca.crt

SgtCoDFish:
We regularly see people using ca.crt or requesting it be changed. This FAQ entry is intended to be something we can link to in those situations to try and clarify why it can be dangerous!


jakexks (approved):

SgtCoDFish:
Thanks very much @jakexks  - happy new year!

PR #1389, opened 2024-01-15
Improve + fix signature verification examples

SgtCoDFish:
The current examples don't work, so there's a need to fix that with ignoring the tlog.

As a bonus, this removes the need to download the key locally since cosign will happily download it for us!


inteon (commented):
Thanks for fixing this error on the website.
/approve
/lgtm

PR #1398, opened 2024-01-22
Release process updates

SgtCoDFish:
There were a few links which needed updating here. Made a few other tweaks too.


SgtCoDFish:
Thanks for the review!

(Adding self-approval manually given the nature of this PR)

PR #1421, opened 2024-02-08
Bump hashes for csi-driver-spiffe files

SgtCoDFish:
The commit SHA we used was from an old version of the repo.

The files haven't changed, but it's worth bumping these from time to time to improve user UX when they browse the repo after getting there from these links


inteon (approved):
All new links seem to work.

PR #1424, opened 2024-02-09
Update istio-csr installation for Istio >= 1.16

SpectralHiss:
Updated the guide to incorporate the simplifications to the istio installation which are the result of improvements done in https://github.com/istio/istio/pull/39375

We've also changed the examples here: https://github.com/cert-manager/istio-csr/pull/271


inteon (commented):
Great work! Thanks for the docs update.
/approve
/lgtm
/hold

PR #1429, opened 2024-02-16
Document known issue: wrong certificate chain is used if preferredChain is configured

wallrj:
**Preview**: https://deploy-preview-1429--cert-manager-website.netlify.app/docs/releases/release-notes/release-notes-1.14#known-issues

In https://github.com/cert-manager/cert-manager/pull/6755#issuecomment-1944231410 we agreed to:
> update existing release notes with a note that explains that this is a known issue and that the immediate fix is to remove the preferredChain: ISRG Root X1 property
* https://github.com/cert-manager/cert-manager/pull/6755

I'm struggling to make the situation clear so happy to receive any suggestions.

### DIFF
--- a/content/docs/releases/release-notes/release-notes-1.14.md
+++ b/content/docs/releases/release-notes/release-notes-1.14.md
@@ -20,7 +20,7 @@ support for creating [CA certificates with "Name Constraints" and "Authority Inf
 
 ### Known Issues
 
-#### ACME Issuer (Let's Encrypt): wrong certificate chain may be used if `preferredChain` is configured
+#### ACME Issuer (Let's Encrypt): wrong certificate chain may be used if `preferredChain` is configured - [#6755](https://github.com/cert-manager/cert-manager/pull/6755), [#6757](https://github.com/cert-manager/cert-manager/issues/6757)
 
 On Thursday, Feb 8th, 2024, [Let's Encrypt stopped providing their cross-signed certificate chain by default](https://letsencrypt.org/2023/07/10/cross-sign-expiration), in requests made to their `/acme/certificate` API endpoint.
 Instead the short-chain is returned by default and the long-chain (cross-signed) certificate chain is now included among the ["alternate" chains](https://www.rfc-editor.org/rfc/rfc8555#section-7.4.2).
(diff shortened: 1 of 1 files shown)


content/docs/releases/release-notes/release-notes-1.14.md
```diff
@@ -18,6 +18,33 @@ support for creating [CA certificates with "Name Constraints" and "Authority Inf
...
+  And then renew any certificates which use that issuer and which have been renewed since Feb 8th 2024.
+  The new certificate will have the shorter chain which terminates at the self-signed root certificate for `ISRG Root X1`.
+
+* **You can do nothing**.
+  The affected certificates will have a longer chain which terminates at `DST Root CA X3` and
+  contains the cross-signed intermediate certificate for `ISRG Root X1`, which expires on September 30 2024.
+  But that's OK because both those certificates are currently trusted by clients
+  and your 90 day leaf certificate is certain to be be renewed before that date,
+  and certain to be renewed **after** June 6 2024, on which day Let's Encrypt will
+  stop providing the longer cross-signed chain entirely.
```
wallrj, line 46: In https://github.com/cert-manager/cert-manager/pull/6755#issuecomment-1944181769 @import-shiburin points out an edge case of a client with a bug that caused it to crash when `DST_Root_CA_X3` is in the chain.
It might be worth saying that some very few clients may have a problem with the unintended change of chain,
in which case the first work around should be used.

inteon: I think that would be useful to mention.

content/docs/releases/release-notes/release-notes-1.14.md
```diff
@@ -18,6 +18,33 @@ support for creating [CA certificates with "Name Constraints" and "Authority Inf
...
+**Some** users who set `Isser.spec.acme.preferredChain: ISRG Root X1` in order to get early access to the Let's Encrypt short-chain certificates, will now get long-chain (cross-signed) certificates when they renew.
+**Most** users will not be affected. Their new certificates will contain the short-chain (not cross-signed) which terminates at `ISRG Root X1`.
+
+> 🔖 Read [cert-manager PR 6755 (bugfix: wrong certificate chain is used if `preferredChain` is configured)](https://github.com/cert-manager/cert-manager/pull/6755) to learn about the bug and to see the proposed fix.
+>
+> 🔖 Read [Let’s Encrypt: chain of trust](https://letsencrypt.org/certificates/) to learn about the hierarchy of root and intermediate certificates.
+
+##### Workarounds
+
+* **You can remove the `spec.acme.preferredChainChain` field** from the Issuer or ClusterIssuer.
```
inteon, line 36:
```suggestion
* **You can remove the `spec.acme.preferredChainChain: "ISRG Root X1"` field** from the Issuer or ClusterIssuer.
```
I think it would be good to repeat here that this is only a problem if you are using "ISRG Root X1" value and Let's encrypt.

PR #1433, opened 2024-03-01
Introduce variables for versions

inteon:
Introduces a new variable format `[[VAR::cert_manager_latest_version]]` which can be used inside codeblocks and reads variables from the `variables.json` file.
The solution is based on https://github.com/facebook/docusaurus/issues/395#issuecomment-1727777067 and works around an issue with MDX which does not allow you to have variables in codeblocks (https://github.com/orgs/mdx-js/discussions/2288).

### DIFF
--- a/content/docs/contributing/policy.md
+++ b/content/docs/contributing/policy.md
@@ -108,7 +108,7 @@ version: 0.1.0
 appVersion: "0.1.0"
 dependencies:
   - name: cert-manager
-    version: v1.8.0
+    version: [[VAR::cert_manager_latest_version]]
     repository: https://charts.jetstack.io
     alias: cert-manager
     condition: cert-manager.enabled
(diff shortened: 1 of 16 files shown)


maelvls:
Hey, great finding! 

Can you update the Release Process document to reflect that? I think we can remove a lot of `sed` that I had added a while back?

/lgtm
/approve

inteon:
I'll create a seperate PR that updates the release doc.
@maelvls 
/unhold

joebowbeer:
References to the variable `[[VAR::cert_manager_latest_version]]` started showing up in release notes and upgrade instructions, leading to confusing instructions such [as](https://cert-manager.io/docs/releases/upgrading/upgrading-1.18-1.19):

> When upgrading to cert-manager 1.19, use the latest patch version: v1.20.2.

PR #1436, opened 2024-03-06
On the approver-policy installation page, use the [[VAR::cert_manager_latest_version]] variable

inteon:
Instead of `REPLACE-WITH-YOUR-CERT-MANAGER-VERSION`, we can now use `[[VAR::cert_manager_latest_version]]` which is dynamically replaced with the latest cert-manager version.


maelvls (approved):
/lgtm
/approve

PR #1439, opened 2024-03-08
trust-manager: improve install instructions

inteon:
*Problem solved:*
This PR makes the install instructions more readable ( adds newlines ).
This PR adds an explicit version to the install instructions, which is our recommended way of installing cert-manager (see https://cert-manager.io/docs/installation/helm/). This gets us a step closer to consistent instructions on all pages of the website.

*Testing done:*
Ran the install instructions.

### DIFF
--- a/content/docs/trust/trust-manager/installation.md
+++ b/content/docs/trust/trust-manager/installation.md
@@ -15,8 +15,17 @@ and as such trust-manager is also installed into the cert-manager namespace.
 
 ```bash
 helm repo add jetstack https://charts.jetstack.io --force-update
-helm upgrade -i -n cert-manager cert-manager jetstack/cert-manager --set installCRDs=true --wait --create-namespace
-helm upgrade -i -n cert-manager trust-manager jetstack/trust-manager --wait
+
+helm upgrade cert-manager jetstack/cert-manager \
+  --install \
+  --create-namespace \
+  --namespace cert-manager \
+  --set installCRDs=true
+
+helm upgrade trust-manager jetstack/trust-manager \
+  --install \
+  --namespace cert-manager \
+  --wait
 ```
 
 ## Installation Options
@@ -39,7 +48,12 @@ As of trust-manager v0.6.0 you can choose to automatically add an approver-polic
 will approve the trust-manager webhook certificate:
 
 ```bash
-helm upgrade -i -n cert-manager trust-manager jetstack/trust-manager --set app.webhook.tls.approverPolicy.enabled=true --set app.webhook.tls.approverPolicy.certManagerNamespace=cert-manager --wait
+helm upgrade trust-manager jetstack/trust-manager \
+  --install \
+  --namespace cert-manager \
+  --wait \
+  --set app.webhook.tls.approverPolicy.enabled=true \
+  --set app.webhook.tls.approverPolicy.certManagerNamespace=cert-manager
 ```
 
 Note that if you've installed cert-manager to a different namespace, you'll need to pass that namespace in `app.webhook.tls.approverPolicy.certManagerNamespace`!


wallrj (changes requested):
> This PR makes the install instructions more readable ( adds newlines ) and adds an explicit version to the install instructions.

What is the advantage of adding an explicit version? 

I think it adds additional burden to the trust-manager release manager.
They now have to remember to update this documentation variable.
Have you updated the release process documentation for trust-manager?

And what happens when we update the versioned documentation? Will those past versions of cert-manager be forever locked to an older version of trust-manager and approver-policy? By versioned documentation I mean :
 * https://cert-manager.io/v1.13-docs, for example.

inteon:
> > This PR makes the install instructions more readable ( adds newlines ) and adds an explicit version to the install instructions.
> 
> What is the advantage of adding an explicit version?
> 
> I think it adds additional burden to the trust-manager release manager. They now have to remember to update this documentation variable. Have you updated the release process documentation for trust-manager?
> 
> And what happens when we update the versioned documentation? Will those past versions of cert-manager be forever locked to an older version of trust-manager and approver-policy? By versioned documentation I mean :
> 
> * https://cert-manager.io/v1.13-docs, for example.

We want to advise people to explicitly provide the version when installing (same goes for https://cert-manager.io/docs/installation/helm/).

This will additionally notify the user in case of a out-of-date chart repo that does not have the specified version.

inteon:
> And what happens when we update the versioned documentation? Will those past versions of cert-manager be forever locked to an older version of trust-manager and approver-policy? By versioned documentation I mean :

We probably should not make the trust-manager and approver-policy versions dependant on the cert-manager version.
Maybe we should have a set of global variables that apply to all the cert-manager versions?

wallrj (changes requested):
> We want to advise people to explicitly provide the version when installing (same goes for https://cert-manager.io/docs/installation/helm/).

Why do we want to advise people to explicitly provide the version?
I regard these as quick-start installation instructions and the changes here make the instructions more complicated.

And why do we want to advise people to **upgrade** cert-manager to 1.14.2 before they install trust-manager. 
What if they're trying this out on a cluster where cert-manager 1.12 is already installed and accidentally upgrade cert-manager.

> This will additionally notify the user in case of a out-of-date chart repo that does not have the specified version.

But the first line is `helm repo add jetstack https://charts.jetstack.io --force-update`,
so that should never happen, right?

inteon:
> > We want to advise people to explicitly provide the version when installing (same goes for https://cert-manager.io/docs/installation/helm/).
> 
> Why do we want to advise people to explicitly provide the version? I regard these as quick-start installation instructions and the changes here make the instructions more complicated.
> 
> And why do we want to advise people to **upgrade** cert-manager to 1.14.2 before they install trust-manager. What if they're trying this out on a cluster where cert-manager 1.12 is already installed and accidentally upgrade cert-manager.
> 
> > This will additionally notify the user in case of a out-of-date chart repo that does not have the specified version.
> 
> But the first line is `helm repo add jetstack https://charts.jetstack.io --force-update`, so that should never happen, right?

As explained in the PR description, moving away from **upgrade** is not one of the problems this PR is solving.

wallrj (changes requested):
> Problem solved:
> This PR makes the install instructions more readable ( adds newlines ).

I agree that the new commands are more readable on narrow screens.
Thanks.

> This PR adds an explicit version to the install instructions, which is our recommended way of installing cert-manager (see https://cert-manager.io/docs/installation/helm/). This gets us a step closer to consistent instructions on all pages of the website.

I disagree with this change because:
* it makes it seem that trust-manager 0.9.0  is dependent on cert-manager 1.14.4
* it adds an additional burden to the trust-manager release process.
* it makes it even more difficult to maintain the versioned cert-manager docs.

inteon:
@wallrj I removed the versions, will create another PR for approver-policy too.

wallrj (approved):
Thanks.

/lgtm

PR #1446, opened 2024-03-13
Explain that cmctl x uninstall is now safe by default

inteon:
This PR adds the documentation for the feature that was added in https://github.com/cert-manager/cmctl/pull/13.

### DIFF
--- a/content/docs/reference/cmctl.md
+++ b/content/docs/reference/cmctl.md
@@ -332,8 +332,6 @@ cmctl x uninstall
 cmctl x uninstall --namespace my-cert-manager
 
 cmctl x uninstall --dry-run
-
-cmctl x uninstall --no-hooks
 ```
 
 ### Upgrade
--- a/content/docs/reference/cmctl.md
+++ b/content/docs/reference/cmctl.md
@@ -317,7 +317,10 @@ cmctl x uninstall
 
 This command uninstalls any Helm-managed release of cert-manager.
 
-The CRDs will be deleted if you installed cert-manager with the option `--set CRDs=true`.
+Starting from cmctl `v2.0.0`, the uninstall command is safe and will not delete the CRDs
+by default, even if they were installed with the Helm chart (using the option `--set installCRDs=true`).
+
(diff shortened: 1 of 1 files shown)


content/docs/reference/cmctl.md
```diff
@@ -329,8 +332,6 @@ cmctl x uninstall
 cmctl x uninstall --namespace my-cert-manager
 
 cmctl x uninstall --dry-run
-
-cmctl x uninstall --no-hooks
```
inteon, line 333: unrelated change: this flag is no longer supported since v2.0.0

content/docs/reference/cmctl.md
```diff
@@ -317,7 +317,10 @@ cmctl x uninstall
 
 This command uninstalls any Helm-managed release of cert-manager.
 
-The CRDs will be deleted if you installed cert-manager with the option `--set CRDs=true`.
+Starting from cmctl `v2.0.0`, the uninstall command is safe and will not delete the CRDs
+by default, even if they were installed with the Helm chart (using the option `--set installCRDs=true`).
+
+> 👴 Before `v2.0.0`, cmctl would remove the CRDs if they were installed with the Helm chart (similar to Helm's behavior).
```
wallrj, line 323: Not sure what the old man icon is about :-) Typo?

wallrj (changes requested):

content/docs/reference/cmctl.md
inteon: I thought that would be icon that indicates that this doc is for an old version.

inteon: Maybe 🕐 could be used instead?

wallrj: Oh. Ok. If it was intentional, then I'm happy to go with that.

/lgtm
/hold because I think the 🕐 does look better.

inteon: I updated the icon.

wallrj (approved):
Thanks.

/lgtm

PR #1448, opened 2024-03-15
Remove approver-policy versions from install instructions

inteon:
@wallrj requested these versions to be removed, see https://github.com/cert-manager/website/pull/1439#pullrequestreview-1933998056.


wallrj (approved):
Thanks @inteon 

I think this is better, because 
* it makes the installation more readable and 
* it reduces the steps needed to release a new trust-manager
* the hard coded version of trust-manager won't necessarily be appropriate when viewed in the versioned documentation URLs such as https://cert-manager.io/v1.12-docs.

In a future PR we could also change the previous  `helm upgrade cert-manager` step to `helm install cert-manager`,
because I think a user might accidentally upgrade an important cert-manager installation if they're using the wrong kube context.

/lgtm

PR #1449, opened 2024-03-15
Update install instructions, making them consistent for all sub-projects

inteon:
*Goal:*
Reach base level of consistency for installation commands across all installation pages.
Once we reach consistency, we can think about improving these instructions (keeping in mind the CRD changes coming in 1.15).

Next steps:
1. Prepare 1.15 release (drastically simplify the CRD install section & use the single fool-proof install command everywhere)

### DIFF
--- a/content/docs/installation/helm.md
+++ b/content/docs/installation/helm.md
@@ -28,13 +28,7 @@ Notably, the "Helm stable repository" version of cert-manager is deprecated and 
 helm repo add jetstack https://charts.jetstack.io --force-update
 ```
 
-#### 2. Update your local Helm chart repository cache:
-
-```bash
-helm repo update
-```
-
-#### 3. Install `CustomResourceDefinitions`
+#### 2. Install `CustomResourceDefinitions`
 
 cert-manager requires a number of CRD resources, which can  be installed manually using `kubectl`,
 or using the `installCRDs` option when installing the Helm chart. Both options
@@ -61,7 +55,7 @@ Uncomment the relevant line in the next steps to enable this.
 
 Note that if you're using a `helm` version based on Kubernetes `v1.18` or below (Helm `v3.2`), `installCRDs` will not work with cert-manager `v0.16`. See the [v0.16 upgrade notes](../releases/upgrading/upgrading-0.15-0.16.md#helm) for more details.
 
-#### 4. Install cert-manager
+#### 3. Install cert-manager
 
 To install the cert-manager Helm chart, use the [Helm install command](https://helm.sh/docs/helm/helm_install/) as described below.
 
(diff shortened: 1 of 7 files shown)


wallrj (changes requested):
1. Seems like unnecessary churn to me. 
2. It introduces the cert-manager version everywhere, giving the false impression (I think) that istio-csr, csi-driver, etc depend on that particular version. As I [already said in another review](https://github.com/cert-manager/website/pull/1439#pullrequestreview-1933998056).
3. I think the goal should be to remove the `helm **upgrade**` instruction from each of the component install pages, because it risks users accidentally upgrading an important existing cert-manager installation and this change does not advance that goal.
4. I don't have time to test all the changes.

Have you tested everything?

inteon:
@wallrj

I updated the cert-manager instructions for approver-policy and csi-driver-spiffe (replaced `helm install`/ `helm upgrade --install` with `helm upgrade --reuse-values`. I also tested that this works as expected for the approver-policy installation.

content/docs/installation/helm.md
```diff
@@ -28,13 +28,7 @@ Notably, the "Helm stable repository" version of cert-manager is deprecated and
 helm repo add jetstack https://charts.jetstack.io --force-update
 ```
 
-#### 2. Update your local Helm chart repository cache:
```
inteon, line 31: This section has become redundant.

content/docs/usage/istio-csr/installation.md
```diff
@@ -83,16 +82,18 @@ istio-csr is best installed via Helm, and it should be simple and quick to insta
 are a bunch of other configuration options for the helm chart, which you can check out [here](../../usage/istio-csr/README.md).
 
 ```console
```
ThatsMrTalbot, line 84: Using console here, and terminal everywhere else

content/docs/usage/csi-driver-spiffe/installation.md
```diff
@@ -18,17 +18,13 @@ race with cert-manager and policy enforcement will become useless.
...
-# See explanation above!
+Here's a example which reconfigure an installed cert-manager to run without auto-approver:
 
-helm upgrade -i -n cert-manager cert-manager jetstack/cert-manager \
-  --set extraArgs={--controllers='*\,-certificaterequests-approver'} \
-  --set installCRDs=true \
-  --create-namespace
+```terminal
+helm upgrade cert-manager jetstack/cert-manager \
+  --reuse-values \
```
ThatsMrTalbot, line 25: Reuse values does not keep the version, should probably add a version flag to not upgrade people accidentally

inteon: fixed

PR #1450, opened 2024-03-15
Docker testing and validation

davidnoyes:


dockerfile
ThatsMrTalbot, line 1: Can you rename this `Dockerfile`

davidnoyes: The joys of case-insensitive file systems. Im on it

davidnoyes: Done

davidnoyes:
I'd say that this change doesn't set a new default but just adds another option.

PR #1452, opened 2024-03-15
feat: add makefile modules for local dev

ThatsMrTalbot:
This adds makefile-modules to the repo, this gives us:
- `make help` consistent with the other repos
- `node` and `npm` are vendored in _bin like we do for all other tools 
- Targets like `make build` and `make dev` to simply development (no need to install node/npm)


inteon:
There were quite a few people who told me that setting up node and npm for this project is complex and that this PR will improve their experience.
/approve
/lgtm

PR #1454, opened 2024-03-27
Bump required k8s version for csi-driver

SgtCoDFish:
This comes from using CSIDriver v1 unilaterally in v0.8.0

Note: K8s docs may differ slightly on the exact version when CSIDriver v1 was introduced.

[Here](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#storage-resources-v122) points to v1.19 as the first version with v1 available.

That doesn't match [here](https://kubernetes-csi.github.io/docs/csi-driver-object.html#status) which claims GA (and presumably v1) came in v1.18. We use v1.19 here for safety and because these versions are so ancient that it really doesn't seem to matter.

### DIFF
--- a/content/docs/usage/csi-driver/installation.md
+++ b/content/docs/usage/csi-driver/installation.md
@@ -5,7 +5,7 @@ description: 'Installation guide for cert-manager csi-driver'
 
 ## Installation Steps
 
-You must have a working installation of cert-manager present on your cluster and be running at least Kubernetes `v1.16`.
+You must have a working installation of cert-manager present on your cluster and be running at least Kubernetes `v1.19`.
 
 Instructions on how to install cert-manager can be found [on this website](../../installation/README.md).
 
--- a/content/docs/usage/csi.md
+++ b/content/docs/usage/csi.md
@@ -7,40 +7,43 @@ description: 'cert-manager usage: CSI driver'
 <object data="/images/request-certificate-overview/request-certificate-csi.svg"></object>
 </div>
 
-## Enabling mTLS of Pods using the cert-manager CSI Driver
+## Enabling mTLS of Pods using cert-manager csi-driver
 
-A [Container Storage Interface (CSI)
-driver](csi-driver/README.md) has been created to
-facilitate mTLS of Pods running inside your cluster through use of cert-manager.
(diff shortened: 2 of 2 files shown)


content/docs/usage/csi-driver/installation.md
```diff
@@ -5,7 +5,7 @@ description: 'Installation guide for cert-manager csi-driver'
 
 ## Installation Steps
 
-You must have a working installation of cert-manager present on your cluster and be running at least Kubernetes `v1.16`.
+You must have a working installation of cert-manager present on your cluster and be running at least Kubernetes `v1.19`.
```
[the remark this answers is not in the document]
SgtCoDFish: Thanks for spotting that! I've updated it quite extensively to try and improve. Does it look better now?

content/docs/usage/csi.md
```diff
@@ -7,40 +7,43 @@ description: 'cert-manager usage: CSI driver'
 <object data="/images/request-certificate-overview/request-certificate-csi.svg"></object>
 </div>
 
-## Enabling mTLS of Pods using the cert-manager CSI Driver
+## Enabling mTLS of Pods using cert-manager csi-driver
```
inteon, line 10: This page should actually contain info for both csi-driver and csi-driver-spiffe (based on the location in the website structure.)

content/docs/usage/csi.md
```diff
@@ -7,40 +7,43 @@ description: 'cert-manager usage: CSI driver'
...
-> number of certificates that can be issued for a single domain. Like Pods,
-> these certificate key pairs are designed to be non-immutable and can be
+that the Pod is scheduled to.
+
+The life cycle of the certificate key pair matches that of the Pod; the certificate is issued
+when the Pod is creation, and destroyed during termination.
+
+This driver also handles renewal of live certificates on the fly.
+
+## What's a CSI driver?
```
inteon, line 24: No comments for this section, looks very good!

content/docs/usage/csi.md
SgtCoDFish: I'll probably hold off on making that change - the scope of this has already increased.

PR #1397, opened 2024-01-19
Document External Vault secretless auth options

SpectralHiss:
Fixes https://github.com/cert-manager/cert-manager/issues/6150

* Explain when to use `kubernetes` vs `jwt` auth for secretless auth
* Public clouds/generic + Openshift in-depth guide

Preview: https://deploy-preview-1397--cert-manager-website.netlify.app/docs/configuration/vault/#secretless-authentication-with-a-service-account

PR #1455, opened 2024-03-28
Vault: document serviceAccountRef with an external Vault

maelvls:
We forgot to document this feature after https://github.com/cert-manager/cert-manager/issues/6666 was merged.


content/docs/configuration/vault.md
```diff
@@ -260,7 +262,43 @@ needs to talks to Vault.
...
+  vault:
+    path: pki_int/sign/example-dot-com
+    server: https://vault.local
+    auth:
+      kubernetes:
+        role: my-app-1
+        mountPath: /v1/auth/kubernetes
+        serviceAccountRef:
+          name: vault-issuer
+          audiences: [https://kubernetes.default.svc.cluster.local]
```
SpectralHiss, line 296: Shouldn't the audiences also include something like vault to tighten the integration?

SpectralHiss (commented):
Nice this looks good, probably means we can close [this](https://github.com/cert-manager/website/pull/1397) and not confuse most cert-manager users!

content/docs/configuration/vault.md
maelvls: The default "computed" audience (e.g., `vault://namespace/issuer-name`) is still added to the audiences by default, so I'd say it would be redundant to add `vault` there? Of course, that's only useful if folks are also using `audience` in their vault role, otherwise the audience isn't checked by vault.

SpectralHiss: Ah yeah, perhaps I would add a sentence to say that the default behavior will also include `vault://namespace/issuer-name` which you can use in the vault role as well.

maelvls: Done

SpectralHiss (approved):

content/docs/configuration/vault.md
```diff
@@ -464,7 +466,47 @@ needs to talks to Vault.
 Although it is not recommended, you can also use the same Vault role for all of
 your Issuers and ClusterIssuers by omitting the `audience` field and re-using
 the same service account.
-<a name="static-service-account-token"></a>
+
+#### Secretless Authentication with a Service Account (External Vault)
+
```
SpectralHiss, line 471: @maelvls I just realized we need to add a little footnote to say this feature is available starting cert-manager 1.15 (just like it says for the above paragraph line 366

ThatsMrTalbot (commented):
Need to add `v1.15.0` to the spelling file for the checks to pass

.spelling
```diff
@@ -532,6 +532,7 @@ v1.12.0
 v1.12.1.
 v1.12.2.
 v1.12.3.
+v1.15.0
```
ThatsMrTalbot, line 535:
```
[spelling]       472 |  available in cert-manager >= v1.15.0. 
```

I think it wants the full stop included. I don't think the spelling module is very smart.

SpectralHiss (approved):

PR #1456, opened 2024-03-29
Release Process: cmctl no longer shipped as part of cert-manager releases

maelvls:
While releasing 1.15.0-alpha.0, I found a couple of inconsistencies due to the fact that `cmctl` now lives in its own repository.

### DIFF
--- a/content/docs/contributing/release-process.md
+++ b/content/docs/contributing/release-process.md
@@ -385,7 +385,10 @@ page if a step is missing or if it is outdated.
       > kicking off a build using the steps in `gcb/build_cert_manager.yaml`. Users with access to
       > the cert-manager-release project on GCP should be able to view logs in [GCB build history](https://console.cloud.google.com/cloud-build/builds?project=cert-manager-release).
 
-9. **(1.12, 1.13, and 1.14)** In this step, we make sure the Go module
+9. <details>
+   <summary>**ONLY for (1.12, 1.13, and 1.14)**</summary>
+
+   In this step, we make sure the Go module
    `github.com/cert-manager/cert-manager/cmd/ctl` can be imported by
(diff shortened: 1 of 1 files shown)


content/docs/contributing/release-process.md
```diff
@@ -472,7 +471,7 @@ page if a step is missing or if it is outdated.
 
        ```bash
        # Must be run from the cert-manager folder.
-       export GITHUB_TOKEN=*your-token*
+       export GITHUB_TOKEN=$(gh auth token)
```
inteon, line 474: Wow, haven't tried this yet, but might just be the amazing timesaver I was looking for.
I have always done this through the UI, which takes a lot of time IMO.

content/docs/contributing/release-process.md
```diff
@@ -385,7 +385,7 @@ page if a step is missing or if it is outdated.
       > kicking off a build using the steps in `gcb/build_cert_manager.yaml`. Users with access to
       > the cert-manager-release project on GCP should be able to view logs in [GCB build history](https://console.cloud.google.com/cloud-build/builds?project=cert-manager-release).
 
-9. **(1.12 and above)** In this step, we make sure the Go module
+9. **(1.12, 1.13, and 1.14)** In this step, we make sure the Go module
```
inteon, line 388: Is there maybe a way to collapse this item, so it becomes harder to run these steps by accident?

inteon: I implemented this in https://github.com/cert-manager/website/pull/1456/commits/d83ebca4c4606124e939547227d7bf279b9d1b6d PTAL.

maelvls: LGTM

PR #1461, opened 2024-04-19
Supported releases / k8s updates

SgtCoDFish:


inteon (commented):
Thanks for the updates, we should also update our tests asap to test against the 1.30 release.

PR #1463, opened 2024-04-24
Update release process to reflect that most of it is more netural now

SgtCoDFish:
This is part of the graduation process!


content/docs/contributing/release-process.md
```diff
@@ -32,25 +29,12 @@ following conditions:
...
-    --- a/cert_manager_release.tf
-    +++ b/cert_manager_release.tf
-    @@ -17,6 +17,7 @@ locals {
-         var.personal_email["..."],
-         var.personal_email["..."],
-         var.personal_email["..."],
-    +    var.personal_email["mael-valais"],
-       ])
-     }
-    ```
```
SgtCoDFish, line 53: self review: Rather than update this diff (which doesn't match how this would be changed today) I removed it. It seems pretty clear to me that anyone who's able to reach the status of maintainer would be able to work out how to add themselves to the list without us having to describe the process here

PR #1470, opened 2024-05-01
Add notes about required tools for macOS

SgtCoDFish:
This was after https://github.com/cert-manager/cert-manager/pull/6978 where it was discovered that an old version of bash caused performance issues.


inteon (commented):
/approve
/lgtm
/hold In case someone with a mac wants to verify.

SgtCoDFish:
(fixed casing of "Coreutils")

PR #1502, opened 2024-06-11
Fix broken link in istio-csr docs

inteon:
The link `https://github.com/cert-manager/istio-csr/tree/main/hack` is no longer valid, instead `https://github.com/cert-manager/istio-csr/tree/5d2e629f9f9aec7f476337d6ae69482cea9a7a71/make/config/istio` can be used.

### DIFF
--- a/content/v1.11-docs/projects/istio-csr.md
+++ b/content/v1.11-docs/projects/istio-csr.md
@@ -32,12 +32,12 @@ Running istio-csr requires a few steps and preconditions in order:
 3. An `Issuer` or `ClusterIssuer` which will be used to issue Istio certificates
 4. istio-csr installed (likely via helm)
 5. Istio [installed](https://istio.io/latest/docs/setup/install/istioctl/) with
-   some custom config required, e.g. using the example config from the [repository](https://github.com/cert-manager/istio-csr/tree/main/make/config/istio).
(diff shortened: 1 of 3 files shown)


content/v1.11-docs/projects/istio-csr.md
```diff
@@ -32,12 +32,12 @@ Running istio-csr requires a few steps and preconditions in order:
 3. An `Issuer` or `ClusterIssuer` which will be used to issue Istio certificates
 4. istio-csr installed (likely via helm)
 5. Istio [installed](https://istio.io/latest/docs/setup/install/istioctl/) with
-   some custom config required, e.g. using the example config from the [repository](https://github.com/cert-manager/istio-csr/tree/main/hack).
+   some custom config required, e.g. using the example config from the [repository](https://github.com/cert-manager/istio-csr/tree/main/make/config/istio).
```
SgtCoDFish, line 35: suggestion: It's great that we're fixing this and this is obviously an improvement.

While we're here, could we not use `main`? That's the reason this broke in the first place. If we linked to a specific revision in the docs it wouldn't auto update but it also wouldn't auto break. I'd rather have it not automatically break in the future

SgtCoDFish (approved):
/lgtm
/approve
/hold

I don't think this should be merged while it still refers to `main` but I recognise it's obviously fixing a broken thing, so I'm adding a hold along with an approval.

PR #1508, opened 2024-06-25
chore: use cm maintainers in owners

ThatsMrTalbot:
Also done in this PR is a klone upgrade, to get the latest `OWNERS_ALIASES`.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -11,3 +11,4 @@ aliases:
     - irbekrm
     - sgtcodfish
     - inteon
+    - thatsmrtalbot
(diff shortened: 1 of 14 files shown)


OWNERS_ALIASES
```diff
@@ -11,3 +11,4 @@ aliases:
     - irbekrm
     - sgtcodfish
     - inteon
+    - thatsmrtalbot
```
maelvls, line 14: Seems like the most important part of this PR is that you added yourself to the OWNERS of cert-manager/website. I'm totally OK with that, but I'd like this information shown in the PR description.

maelvls:
Lots of unrelated changes in this PR, but I guess I'm OK with that. What caught my eye is that the PR description doesn't explain the only actual change in this PR: you added yourself to the approvers (which I'm absolutely OK with, but that needs a like in the PR description at the very least)

PR #1512, opened 2024-07-09
Add redirects for cert-manager.dev, trust-manager.io and trust-manager.dev

inteon:
This will change the url from eg. trust-manager.io to cert-manager.io which will help prevent eg. google from thinking that these are different sites.


SgtCoDFish (approved):
/lgtm
/approve

Down the road it would maybe make sense for the trust-manager ones to redirect to the trust-manager install page?

But I think it's important to have a 301 in place ASAP so approving this as-is!

PR #1513, opened 2024-07-10
Remove $ from bash code snippets

lindhe:
While $ is a good way to differentiate between a command and its output, it is best to leave it out whenever there is no output. This way, the user can copy the command by clicking the copy button on the rendered website without having to worry about the leading $ when pasting the command.

This was already the case for some snippets. This change removes the leading $ for some more snippets.

### DIFF
--- a/content/docs/installation/kubectl.md
+++ b/content/docs/installation/kubectl.md
@@ -81,7 +81,7 @@ Best way to fully verify the installation is to issue a test certificate. For th
 
 
 ```bash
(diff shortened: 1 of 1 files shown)


lindhe:
Here's an example of what it looks like before my change:

Screenshot 2024-07-10 092421

The $ will be included if the users copies via the copy button in the web interface.

lindhe:
Looks good:

image

wallrj (approved):
Thanks @lindhe 

/lgtm
/approve

PR #1514, opened 2024-07-10
Markup shell session snippets as console

lindhe:
This change replaces "bash" with "console" as the Markdown code fence keyword. This lets Markdown engines differentiate between the shell command and its output, with a leading $ (or # or >) denoting a command and the rest is just text output with no syntax highlighting.

### DIFF
--- a/content/docs/installation/kubectl.md
+++ b/content/docs/installation/kubectl.md
@@ -31,7 +31,7 @@ you'll need to make modifications to the deployment manifests.
 Once you've installed cert-manager, you can verify it is deployed correctly by
 checking the `cert-manager` namespace for running pods:
 
-```bash
+```console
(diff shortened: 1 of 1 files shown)


lindhe:
Here's what it looks like on the website today:

image

image

Notice how random integers get highlighted, and whatnot.

lindhe:
Seems like the website will render this without any syntax highlighting (based on the preview build):

image

Yet, as I mentioned, I think this is an improvement over having incorrect syntax highlighting.

wallrj (approved):
Thanks @lindhe 

I agree that console is more appropriate than bash for snippets that show the command and the output.

It's a shame there's no syntax highlighting in for console, but better than highlighting weirdly.

/approve
/lgtm

PR #1516, opened 2024-07-16
Clarify differences between example configs for CA issuer and cluster CA issuer

apiwat-chantawibul:
- Especially around the issue that CA root secret has to be in `cert-manager` namespace.
- Previous version starts off by suggesting the differences is only in "the last step" when in fact CA root secret has also moved to another namespace.

## Might need further correction

- [ ] Is the namespace for CA root secret always assumed to be `cert-manager` when creating cluster CA issuer?
  - What happens if cert-manger is not installed in `cert-manager` namespace?
  - I don't know these so I didn't write more specifically in the edit.

### DIFF
--- a/content/docs/configuration/selfsigned.md
+++ b/content/docs/configuration/selfsigned.md
@@ -117,14 +117,10 @@ spec:
     secretName: root-secret
 ```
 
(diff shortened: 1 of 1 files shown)


erikgb (commented):
/lgtm
/approve

Thanks! Sorry for the delayed review!

PR #1517, opened 2024-07-16
Gateway API: Recommend the `config` approach rather than `extraArgs`

maelvls:
The `config` approach is much more convenient than to have to concatenate flags in the `extraFlags` field. The `config` approach isn't immediately obvious to users, as seen in https://github.com/cert-manager/cert-manager/pull/7121 where we found that there was no need for adding an extra helm value: one could simply use `config.enableGatewayAPI`.


maelvls:
@wallrj Can you take a look? Thanks!

wallrj (changes requested):
Thanks @maelvls 

This looks like a nicer way to configure that feature, but it won't work until the next cert-manager release, right?
It depends on https://github.com/cert-manager/cert-manager/commit/1b9c02e999a90bc5183d441a51700b9c3aaf888b which has not been backported or released yet.

And since this documentation relates to the next version of cert-manager, it should be merged into release-next, along with the release note for the change which makes this possible.

And the DCO check seems to be failing.

hongbo-miao:
Thanks @maelvls @wallrj for the info. I would be great to document which way is better? Right now I saw 3 ways including this way:

```yaml
featureGates: ExperimentalGatewayAPISupport=true
```

```yaml
config:
  featureGates:
    ExperimentalGatewayAPISupport: true
```

```yaml
extraArgs:
  - --feature-gates=ExperimentalGatewayAPISupport=true
```

maelvls:
I'll clarify further why and what the recommended approaches are depending on the cert-manager version. Thanks for the feedback!

maelvls:
Here would be my recommendation depending on whether you are using 1.14 or below, 1.15, or 1.16 and above:

If you are using 1.15 and above, you no longer have to set a feature flag, but
you still need to enable the feature. The recommended way to turn on the Gateway
API support in 1.15 is to use the [file-based configuration](https://cert-
manager.io/docs/installation/configuring-components/#configuration-file) using
the `config` Helm value:

```yaml
config:
  apiVersion: controller.config.cert-manager.io/v1alpha1
  kind: ControllerConfiguration
  enableGatewayAPI: true
```

If you are using 1.16 and above, you can omit `apiVersion` and `kind`:

```yaml
config:
  enableGatewayAPI: true
```

Another less practical way is to rely on the  command line flag `--enable-gateway-api`:

```yaml
extraArgs:
  - --enable-gateway-api
```

---

If you are using 1.14 and below, you will have to use the feature flag to turn on the Gateway API support. The recommended way is to use the `featureGates` Helm value:


Another way is to use the `config` Helm value:


```yaml
config:
  apiVersion: controller.config.cert-manager.io/v1alpha1
  kind: ControllerConfiguration
  featureGates:
    ExperimentalGatewayAPISupport: true
```

A less readable way (as it requires a comma-separated list of feature flags) is to use the `featureGates` Helm value:

```yaml
featureGates: ExperimentalGatewayAPISupport=true
```

which is equivalent to:

```yaml
extraArgs:
  - --feature-gates=ExperimentalGatewayAPISupport=true
```

I'm not sure how to put that into the documentation. Should I only talk about the latest version, or can I explain how to set up Gateway API depending on the version? @wallrj

maelvls:
Discussed in this morning's open standup:

- I will only keep the 1.15 instructions, and only show the `config` approach since it's the most sensible one (we think).
- I will add a note that explains that the feature flag that existed in 1.14 no longer exists and was replaced with a configuration option `enableGatewayAPI`.

content/docs/usage/gateway.md
```diff
@@ -54,22 +54,28 @@ To install v1.5.1 Gateway API bundle (Gateway CRDs and webhook), run the followi
...
+The corresponding Helm command is:
 
-  ```yaml
-  args:
-    - --enable-gateway-api
-  ```
+```sh
+helm upgrade --install cert-manager jetstack/cert-manager --namespace cert-manager \
+  --set apiVersion="controller.config.cert-manager.io/v1alpha1" \
+  --set config.kind=ControllerConfiguration" \
```
maelvls, line 76: Oops, typo here.

```suggestion
  --set config.kind="ControllerConfiguration" \
```

wallrj (approved):
/approve
/lgtm

maelvls:
Now that cert-manager 1.16 has been released, `--set config.enableGatewayAPI=true` is now the recommended approach for projects that show instructions on how to enable cert-manager's gateway API support, especially on visible projects like Cilium:

Projects that use `extraArgs` + the deprecated `--feature-gates=ExperimentalGatewayAPISupport=true` feature flag:

- Cilium: https://docs.cilium.io/en/latest/network/servicemesh/tls-termination/ (source: [`tls-cert.rst`](https://github.com/cilium/cilium/blob/aabda4bdbb578f8159627ca145ece7e66ad3df7a/Documentation/network/servicemesh/tls-cert.rst?plain=1#L45))

Projects that use `extraArgs` + `--enable-gateway-api` instead of `config.enableGatewayAPI`:

- Traefik: https://github.com/traefik/traefik-helm-chart/blob/954e8ac290e1eee8661320c071aca37d8c36deb2/EXAMPLES.md?plain=1#L936
- Tetrate's Envoy Proxy: https://github.com/tetratelabs/eg-course/blob/66194c52c99c780f9606fcef7157e1c7ebda6693/docs/https.md?plain=1#L29
- ksctl: https://github.com/ksctl/ksctl/blob/34299eaf5b57aaa52960ca614a0495f8b3dee491/internal/kubernetes/components/certmanager.go#L72
- https://github.com/search?q=language%3Ayaml+cert-manager+extraArgs+enable-gateway-api&type=code&p=2

PR #1526, opened 2024-07-24
Add release notes for v1.16.0-alpha.0

wallrj:
**Preview**: https://deploy-preview-1526--cert-manager.netlify.app/docs/releases/release-notes/release-notes-1.16#changes-since-v1150

Part of the v1.16.0-alpha.0 release process. 
 * https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1721836875640659

### DIFF
--- a/content/docs/releases/release-notes/release-notes-1.16.md
+++ b/content/docs/releases/release-notes/release-notes-1.16.md
@@ -26,6 +26,32 @@ Thanks also to the CNCF, which provides resources and support, and to the AWS op
 
 In addition, massive thanks to Venafi for contributing developer time and resources towards the continued maintenance of cert-manager projects.
 
-## `v1.16.0`
+## Changes since `v1.15.0`
 
-TODO
+### Feature
+
+- Add `SecretRef` support for Venafi TPP issuer CA Bundle ([#7036](https://github.com/cert-manager/cert-manager/pull/7036), [`@sankalp-at-gh`](https://github.com/sankalp-at-gh))
+- Add a metrics server to the cainjector ([#7194](https://github.com/cert-manager/cert-manager/pull/7194), [`@wallrj`](https://github.com/wallrj))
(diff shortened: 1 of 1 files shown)


content/docs/releases/release-notes/release-notes-1.16.md
```diff
@@ -26,6 +26,32 @@ Thanks also to the CNCF, which provides resources and support, and to the AWS op
...
 In addition, massive thanks to Venafi for contributing developer time and resources towards the continued maintenance of cert-manager projects.
 
-## `v1.16.0`
-
-TODO
+## Changes since `v1.15.0`
+
+### Feature
+
+- Add `SecretRef` support for Venafi TPP issuer CA Bundle ([#7036](https://github.com/cert-manager/cert-manager/pull/7036), [`@sankalp-at-gh`](https://github.com/sankalp-at-gh))
```
maelvls, line 33: **nit, don't feel the need to change anything**: I wish we could write release notes that are slightly more user-focused, but I realize that it isn't that easy since each contributor writes their own release notes. Among others:

- Most of these items focus on what was technically done to achieve the feature (e.g., "add X" rather than explaining whether/how it can impact the user.
- "Features" is super heterogenous; I think the Helm improvements should be in their own section or subsection, and maybe also group "Runtime Improvements", and "API Changes".

Example:

> - Add `SecretRef` support for Venafi TPP issuer CA Bundle

should be reworded as:

> - The Venafi issuer now lets you load your custom CA bundle to connect to TPP from a Secret resource using the new field `secretRef`.

I think we would benefit greatly from having a technical writter look at our release notes from time to time. Maybe we could get @ditatechwriter (Michael) to give some advice on this.

maelvls:
Since this is an alpha, I don't think it's super important to write great release notes. But for the final release, we surely do! I've left a comment in that sense.

/lgtm
/approve

Issue #1474, opened 2024-05-02
Inconsistent opinion on common name field


stevenpitts:
https://github.com/cert-manager/website/blob/eab7224578999dde790641627052b5a635bc3579/content/docs/usage/certificate.md?plain=1#L66-L67

As explored and summarized in https://github.com/caddyserver/caddy/issues/3755#issue-709367646, the deprecation of the CN field has some ambiguity. I don't have an opinion one way or the other, but cert-manager discourages its use *and* documents usage of it in a "Bootstrapping CA Issuers" example:

https://github.com/cert-manager/website/blob/eab7224578999dde790641627052b5a635bc3579/content/docs/configuration/selfsigned.md?plain=1#L100

If `commonName` were removed from that example, applying it would result in an error:

```
at least one of commonName, dnsNames, uriSANs, ipAddresses, emailSANs or otherNames must be set
```

The language in the comment should be made less absolute, or the usage of `commonName` in examples should be replaced with a field that is *not* deprecated.

SgtCoDFish:
Fair comment to make!

I think we should say that the use of commonName is discouraged for end-entity certificates, where the SAN extension is simply better. For CA certs commonName is _fine_. I'll create a PR to try and capture that 🔨

inteon:
fixed in https://github.com/cert-manager/website/pull/1548

PR #1548, opened 2024-09-05
Update note on commonName field

SgtCoDFish:
See #1474

Also changes the example org!


inteon:
fixes https://github.com/cert-manager/website/issues/1474

PR #1560, opened 2024-09-18
Fix bug in docs saying group should be core instead of the empty string

inteon:
Fixes https://github.com/cert-manager/cert-manager/issues/6948

Updates the documentation examples based on https://pkg.go.dev/sigs.k8s.io/gateway-api@v1.1.0/apis/v1#SecretObjectReference.Group.

### DIFF
--- a/content/docs/usage/gateway.md
+++ b/content/docs/usage/gateway.md
@@ -187,7 +187,7 @@ spec:
         certificateRefs:
           - name: example-com-tls
             kind: Secret
-            group: core
+            group: ""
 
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm
/approve

I don't really have bandiwidth to dig on this to test, but I assume it's an improvement! Happy to review followups if needed!

inteon:
FYI, here is an example of a valid configuration used in production: https://github.com/cert-manager/testing/blob/master/prow/cluster/gateway.yaml#L24-L25

PR #1563, opened 2024-09-25
Update supported releases page

SgtCoDFish:
1. Add abbr tags to define abbreviations
2. Update k8s versions for cloud vendors
3. Treat OpenShift EUS as a separate vendor
4. Move "old releases" to the bottom of the page
5. Update supported versions for current releases to include k8s 1.31


SgtCoDFish:
/approve

Cheers Erik!

PR #1565, opened 2024-09-26
Release notes updates for cert-manager v1.16.0-beta.0

wallrj:
Preview:
 * https://deploy-preview-1565--cert-manager.netlify.app/docs/installation/
 * https://deploy-preview-1565--cert-manager.netlify.app/docs/releases/release-notes/release-notes-1.16
 * https://deploy-preview-1565--cert-manager.netlify.app/docs/releases/upgrading/upgrading-1.15-1.16

There's more work to do, but in this PR I've updated the raw release notes,
and started filling out the themes of the release and summarizing the breaking changes.

I'd like to merge this into release-next and continue improving the documentation in follow up PRs
so that content will be available to beta testers who click on the `next` link in the website version selector.

PR #1520, opened 2024-07-18
Release cert-manager 1.16

wallrj:
**Preview**: https://deploy-preview-1520--cert-manager.netlify.app/

Not to be merged until cert-manager 1.16.0 is released and when all the cert-manager 1.16 release notes have been merged to the release-next branch.

/hold

PR #1568, opened 2024-09-27
Update breaking changes details 1.16

wallrj:
**Preview**: https://deploy-preview-1568--cert-manager.netlify.app/docs/releases/release-notes/release-notes-1.16/#breaking-changes

* Applied @maelvls 's suggestions (from https://github.com/cert-manager/website/pull/1520#discussion_r1778271475)
* Prioritized the Helm schema information (as suggested by @SgtCoDFish in https://github.com/cert-manager/website/pull/1565#discussion_r1777437666) 
* Added some further reading to the Helm schema section.


content/docs/releases/release-notes/release-notes-1.16.md
```diff
@@ -7,12 +7,31 @@ cert-manager 1.16 includes various improvements to the metrics in the cert-manag
...
+### Helm
+
+The Helm chart now includes a JSON schema which will validate the values that you supply when installing the chart.
+This will help you to get your Helm values right first time.
+It will alert you to typos and unrecognized fields in your existing Helm values files.
+And it will make it easier for the cert-manager maintainers to maintain the Helm chart,
+avoiding typos and mistakes in the default values file.
+
+> ⚠️ Helm schema validation may reject your existing Helm values files if they contain typos or unrecognized fields.
+> You can use `helm template cert-manager --repo https://charts.jetstack.io --version 1.16.x --values values.cert-manager.yaml`
```
jsoref, line 28: the `1.16.x` bit doesn't work...

Today, I can use:
helm template cert-manager --repo https://charts.jetstack.io --version 1.16.0-beta.0 --values values.yaml 

But, as a reader, I expect a command like this to work as is in the docs, and not require guessing.

SgtCoDFish, line 28: suggestion: Having the command in just backticks like this means it gets split across lines:

Screenshot 2024-09-27 at 18 31 35

That makes it harder to copy and I think a lot of people upgrading will want to copy it!

I'd suggest having it in triple backticks instead so it's rendered as a code block:

```suggestion
>   ```bash
>   helm template cert-manager --repo https://charts.jetstack.io --version 1.16.x --values values.cert-manager.yaml
>   ```
```

(the suggestion might not work as-is, but you get the idea)

wallrj: @SgtCoDFish Done. And added an example of an error.

@jsoref You're suggested command would only work for your specific sub-chart use-case, right? For general users it is better to state the repo URL (otherwise they'd need to run `helm repo update` first) and it's better to state the version otherwise they may not be testing against the cert-manager 1.16 release. 

I wrote `--version 1.16.x` deliberately, because I thought it would continue to work for future 1.16 patch releases.
It does not work with 1.16 pre-releases without an additional `--devel` flag.
So I've added an explicit version variable instead which should be updated with future patch releases.

SgtCoDFish (approved):
/lgtm
/approve

I think this looks good. I don't actually think we need to update the `--version` unless we notice a major bug in the chart for 1.16.1, since this is just to check that pre-existing values work. But no major thing either way.

Thanks for this!

PR #1580, opened 2024-10-05
Update trust-manager API docs

erikgb:
After some refactoring of the CRD handling in trust-manager, the script to generate trust-manager API docs seems broken. This PR applies a quick fix to the generate-script and regenerates the trust-manager API docs.

Some upstream K8s godocs copied into the Bundle CRD is rendering invalid HTML and needs to be "escaped"/replaced.

Reported in https://github.com/cert-manager/trust-manager/issues/428.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -87,6 +87,7 @@ APIServices
 APIs
 AWS
 Akamai
+ANDed
 Anthos
 AppRole
 approvers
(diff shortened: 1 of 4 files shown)


SgtCoDFish (approved):
/lgtm
/approve

Thanks!

PR #1595, opened 2024-10-29
Add new logo for Venafi, tweak wording

SgtCoDFish:
I used a PNG rather than the SVG to make this simpler.


wallrj (approved):
Thanks @SgtCoDFish 

I opened  the preview and it looks ok to me.

Agree that PNG is simpler than the embedded SVG.

/approve
/lgtm

PR #1603, opened 2024-11-28
Add more warnings about using the CA issuer safely

SgtCoDFish:

PR #1605, opened 2024-12-02
Some fixes to new language in CA issuer configuration

SgtCoDFish:
Thanks @james-w for the suggestion!

This also makes the whitespace more uniform


SgtCoDFish:
Follow up to #1603

PR #1627, opened 2025-01-27
Update supported releases

SgtCoDFish:
1. Add initial details about 1.18
2. Clarify LTS support
3. Add support (but not testing) for k8s 1.32
4. Move "inspired by" to the bottom of the page (it's important to add that on the page, but it isn't important enough for it to be the first thing on the page)


inteon (commented):
/approve
/lgtm

Seems to be correct afaics

PR #1658, opened 2025-03-19
approver-policy doc: fix indentation

rptaylor:
matchNames and matchLabels should be at the same indentation.

```
$ kubectl explain CertificateRequestPolicy.spec.selector.namespace
GROUP:      policy.cert-manager.io
KIND:       CertificateRequestPolicy
VERSION:    v1alpha1

FIELD: namespace <Object>

DESCRIPTION:
    Namespace is used to match by namespace, meaning the
    CertificateRequestPolicy will only match CertificateRequests
    created in matching namespaces.
    If this field is omitted, resources in all namespaces are checked.
    
FIELDS:
  matchLabels	<map[string]string>
    MatchLabels is the set of Namespace labels that select on
    CertificateRequests which have been created in a namespace matching the
    selector.

  matchNames	<[]string>
    MatchNames is the set of namespace names that select on
    CertificateRequests that have been created in a matching namespace.
    Accepts wildcards "*".
```

### DIFF
--- a/content/docs/policy/approval/approver-policy/README.md
+++ b/content/docs/policy/approval/approver-policy/README.md
@@ -342,9 +342,9 @@ spec:
       matchNames:
       - "default"
       - "app-team-*"
-    matchLabels:
-      foo: bar
-      team: dev
+      matchLabels:
+        foo: bar
+        team: dev
 ```
 
 ```yaml


erikgb:
Thanks for this! Looks like a clear bugfix. 

I am going to close and reopen the PR now to see if it can fix CI.

erikgb (commented):
/lgtm
/approve

Thanks! 🚀

PR #1664, opened 2025-04-02
Add gonicus/azure-keyvault-issuer

joshmue:
I'd like to add the https://github.com/gonicus/azure-keyvault-issuer issuer to the issuer list. It is intended to issue certificates using an Azure Keyvault Key.

While it is "unmaintained" (-> 🥉), it worked quite well and IMHO is worth sharing.

I sprinkled some "3rd party" and "unofficial issuer controller" here and there because Azure's trademark policies looked somewhat unclear to me, but please feel free to remove these remarks as you see fit.

### DIFF
--- a/content/docs/configuration/issuers.md
+++ b/content/docs/configuration/issuers.md
@@ -31,6 +31,7 @@ The following list contains all known cert-manager issuer integrations.
 | 🥉   | tcs-issuer                  | [📄][config:tcs-issuer]             | [Intel's SGX technology][ca:tcs-issuer]                                | -                                                 | [❌][release:tcs-issuer]              | ✔️             |
 | 🥉   | freeipa-issuer              | [📄][config:freeipa-issuer]         | [FreeIPA][ca:freeipa-issuer]                                           | -                                                 | [❌][release:freeipa-issuer]          | ✔️             |
 | 🥉   | kms-issuer                  | [📄][config:kms-issuer]             | [AWS KMS][ca:kms-issuer]                                               | -                                                 | [❌][release:kms-issuer]              | ✔️             |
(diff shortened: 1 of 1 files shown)


wallrj (changes requested):
Thanks. 

I fixed a typo and pushed to your branch.

I removed the "third party" annotation, because it isn't used for any of the other external issuers....although I agree that it might be a good idea to have a new column to flag whether they are maintained by the company that runs the issuer or by another third party org.

I went looking for a release of the keyvault-issuer and a corresponding Docker image, but couldn't find anything.

I see that the documentation suggests using the Kustomize configs in the main branch of the repo....but which Docker image should be used?
Are users expected to build the image themselves?

I think that needs explainining before we begin linking to the issuer from cert-manager.io.

joshmue:
Thanks for the review!

> I fixed a typo and pushed to your branch.

Regarding the `Azure Keyvault Key` -> `Azure Key Vault` change: The use of `Keyvault` instead of `Key Vault` was indeed definitely wrong. Yet, the `... Key` suffix served the intentional purpose of indicating that the controller uses the `Key` resource of Azure Key Vaults - not e. g. the `Certificate` resource of Azure Key Vaults which would be another imaginable resource to integrate with.
Not saying that the `Key` suffix has to be part of the name here; Just wanted to clarify that including it was not an accident :D

> Are users expected to build the image themselves?

That was indeed implicitly expected, but I can see how this is suboptimal.

Added a build pipeline, tag and container image. Will also proceed to add installation instructions.

joshmue:
Added a bunch of details into the "Getting started" section :+1: 

(Also fixed a bug and created an issue about another one regarding ClusterIssuers)

wallrj:
Thanks. Then let's merge this as-is and you can create another PR when you think the issuer should be promoted to tier 2.

/approve
/lgtm

wallrj (approved):

PR #1694, opened 2025-05-15
enhance restore cert-manager resources

Muhannadimad:
Safely restore cert-manager resources by removing only metadata.uid and resourceVersion, preserving ownerReferences.uid to maintain resource relationships.


erikgb (commented):
Sorry for the delayed review, and thanks a lot for improving our documentation! Highly appreciated!

I tested the updated procedure in one of our clusters, and it appears to work as it should.

/lgtm
/approve

erikgb (commented):
/label tide/merge-method-squash
/lgtm

PR #1698, opened 2025-05-21
Fixed case of 'values.yaml' in Helm installation guide.

Norskov:
The Helm installation guide contained the wrong case for 'values.yaml' causing it to be ignored.


wallrj (approved):
Thanks!

/approve
/lgtm

PR #1731, opened 2025-07-18
Add some basic docs on oss-fuzz

SgtCoDFish:
I've found that we use this infrequently enough to make it difficult to remember context in between visits. As a first pass, this doc should give some basic details and give us a central place to document oss-fuzz stuff later


erikgb (commented):
Excellent start, Ashley!

/lgtm
/approve

PR #1733, opened 2025-07-23
Add info about orphaned leases when uninstalling

erikgb:
I was made aware of https://github.com/kubernetes-sigs/kubebuilder/issues/4712. While I don't understand the actual issue here, we should probably document that cert-manager leases might be orphaned in the cluster after uninstalling. I wanted to use labels to select the leases to delete, e.g., `kubectl delete lease -n <namespace> -l app.kubernetes.io/name=cert-manager`, but it turns out that cert-manager doesn't add any labels to its lease resources.

I will create a cert-manager PR to add labels to leases, as this should be supported now, ref. https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/resourcelock/leaselock.go#L63. The leases label feature was added quite recently in https://github.com/kubernetes/client-go/commit/34f791d2b1d1b3ad3d5d8e7a06f9e1b02abfa7b2, so we have to wait until we upgrade cert-manager dependencies to K8s 1.34.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -448,6 +448,7 @@ kube-ingress-aws-controller
 kube-lego
 kube-apiserver
 kube-proxy
+kube-system
 tcpdump
 kubebuilder
 kubectl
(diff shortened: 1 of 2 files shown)


wallrj (approved):
Thanks @erikgb 

For reference, the lease was originally in the cert-manager namespace but later moved to kube-system in a later PR. See:
* https://github.com/cert-manager/cert-manager/pull/548
* https://github.com/cert-manager/cert-manager/pull/2155

In a future PR, we should explain how to configure the leader-election settings using the config-file method, and why:
 * https://cert-manager.io/docs/installation/configuring-components/#configuration-file

PR #1747, opened 2025-08-24
Manual self upgrade

erikgb:
As our automated self-upgrade doesn't support workflow updates.


erikgb:
No idea why this isn't auto-merging. I will merge it manually.

PR #1796, opened 2025-10-08
Limit version dropdown to last two releases

wallrj-cyberark:
**Preview**: https://deploy-preview-1796--cert-manager.netlify.app/docs/

It was impossible to see the `release-next` link in the versions drop-down list because it went off the bottom of the page.

I've limited it now without actually deleting any of those old pages from the site.

- Show only the two most recent releases in the version dropdown
- Implemented by adding .slice(0, 2) in VersionSelect.jsx

new
image

vs 

old
image

## Testing

I've verified that it works in Microsoft edge and Google Chrome and Firefox Mobile. 

CyberArk tracker: [VC-45942](https://venafi.atlassian.net/browse/VC-45942)

### DIFF
--- a/components/docs/VersionSelect.jsx
+++ b/components/docs/VersionSelect.jsx
@@ -23,6 +23,7 @@ export default function VersionSelect({
       return compareVersions(labelFromVersion(first), labelFromVersion(second))
     })
     .reverse()
+    .slice(0, 2) // Only show the last two releases in the dropdown list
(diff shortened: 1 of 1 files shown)


erikgb (commented):
/lgtm
/approve

Should we clean out old docs in a follow-up PR? How can a user running an old version find the versioned docs now?

PR #1838, opened 2025-11-07
Update cloud vendor k8s versions

SgtCoDFish:
This was done manually. Includes a few formatting tweaks and other minor fixes. Also removes a bunch of "openshift version mappings" because there were a lot of very old EOL releases in there!


erikgb (commented):
/lgtm
/approve

Thanks for this, @SgtCoDFish!

PR #1866, opened 2025-12-04
Use admonitions similar to docusaurus

inteon:
Use [docusaurus style admonitions](https://docusaurus.io/docs/markdown-features/admonitions) to create info/ warning and danger blocks.

Should not change the UI, except for https://cert-manager.io/docs/contributing/release-process/ (which contained broken divs and classes).

The goal is to make our markdown compatible with docusaurus, increasing familiarity for developers and preparing for a long-term docusaurus migration.

CyberArk tracker: [VC-47737](https://venafi.atlassian.net/browse/VC-47737)

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -822,7 +822,6 @@ zerossl-issuer
 ingress.md
 README.md
 certificate.md
-pageinfo
 md#renew
 
 # END TEMPORARY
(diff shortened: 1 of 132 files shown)


erikgb (commented):
Really boring PR to review! 😆 But thanks for this, Tim! Looks good to me, and the preview works as expected when checking a couple of changed files. If anything is broken, we'll just fix it in a follow-up PR.

/lgtm
/approve

PR #1900, opened 2025-12-23
venafi: Issuer custom fields documentation

k0da:
Document Venafi "global" custom fields feature implemented in https://github.com/cert-manager/cert-manager/pull/8301

### DIFF
--- a/content/docs/configuration/venafi.md
+++ b/content/docs/configuration/venafi.md
@@ -304,3 +304,45 @@ metadata:
       ]
 ...
 ```
+
+### Issuer Custom Fields
+
+Starting `v1.20`, you can use `venafi.cert-manager.io/custom-fields` annotation on an `Issuer` or `ClusterIssuer` resource.
+This configuration would be applied to all Certificate requests created from `Issuer`.
+
+It is possible to override or append custom configuration to `Certificate` resources via the `Issuer` assigned to it. 
+For example with an `Issuer` such as:
+
+```yaml
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: corp-issuer
+  annotations:
+    venafi.cert-manager.io/custom-fields: |-
+      [
+        {"name": "Environemnt", "value": "Dev"},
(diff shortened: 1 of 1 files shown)


content/docs/configuration/venafi.md
```diff
@@ -304,3 +304,36 @@ metadata:
...
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: corp-issuer
+  annotations:
+    [
+       {"name": "Environemnt", "value": "Dev"},
+    ]
+```
+and `Certificate`
```
hawksight:
```suggestion
and a `Certificate` resource:

```

content/docs/configuration/venafi.md
```diff
@@ -304,3 +304,36 @@ metadata:
       ]
 ...
 ```
+Starting `v1.20`, you can use `venafi.cert-manager.io/custom-fields` annotation on `Issuer` or `ClusterIssuer` objects.
```
hawksight: I would suggest that you put this into it's own heading so it's clear it is additional new functionality but also optional:

```suggestion

### Issuer Custom Fields

Starting `v1.20`, you can use `venafi.cert-manager.io/custom-fields` annotation on an `Issuer` or `ClusterIssuer` resource.
```

content/docs/configuration/venafi.md
```diff
@@ -304,3 +304,43 @@ metadata:
       ]
 ...
 ```
+
+### Issuer Custom Fields
+
+Starting `v1.20`, you can use `venafi.cert-manager.io/custom-fields` annotation on an `Issuer` or `ClusterIssuer` resource.
+This configuration would be applied to all Certificate requests created from `Issuer`.
+
+It is possible to override or append custom configuration to a specific `Certificate`, in example.
```
hawksight:
```suggestion
It is possible to override or append custom configuration to `Certificate` resources via the `Issuer` assigned to it. 
For example with an `Issuer` such as:

```

PR #2020, opened 2026-03-22
docs: add ENISA NIS2 reference to best practice intro

Xata:
Fixes #2019

Adds [ENISA NIS2 Technical Implementation Guidance](https://www.enisa.europa.eu/publications/nis2-technical-implementation-guidance) to the list of referenced security standards on the best practice page. Restructures the intro into a list for readability and removes the redundant second paragraph.

Updates ```.spelling``` to include new terms introduced by this change.


wallrj-cyberark (commented):
Thank you for this contribution — the restructuring of the intro into a bulleted list with expanded descriptions is a clear improvement, and the corrections to the NSA/CISA attribution and BSI publication name are both accurate.

I have a concern about the ENISA NIS2 reference, though. The existing three references (CIS, NSA/CISA, BSI) all contain **specific Kubernetes hardening sections** that map directly to the operational guidance on this page (pod security contexts, network policies, RBAC, container image security). The page's framing is: *"how to configure cert-manager to comply with popular security standards."*

I downloaded and searched the [ENISA NIS2 Technical Implementation Guidance PDF](https://www.enisa.europa.eu/sites/default/files/2025-06/ENISA_Technical_implementation_guidance_on_cybersecurity_risk_management_measures_version_1.0.pdf) (170 pages). It contains no mention of "Kubernetes" or "container" anywhere in the document. Its section 9 (Cryptography) does cover certificate lifecycle management — issuing, revoking, and managing public key certificates — and footnote 86 explicitly mentions X.509. So there is a tangential connection to cert-manager's domain: cert-manager helps organisations comply with NIS2 section 9.2(c) by automating certificate lifecycle.

However, that is a different claim from *"here is how to configure cert-manager to comply with NIS2"*, which is what the other three references on this page deliver. The ENISA document operates at the organisational cybersecurity policy level ("have a key management policy"), not the Kubernetes operational hardening level ("set `runAsNonRoot=true`"). Listing it alongside the CIS/NSA/BSI references implies a level of Kubernetes-specific guidance that the document does not contain.

Would you be open to either:
1. Moving the ENISA reference to a separate paragraph that frames it differently — e.g. *"cert-manager also supports compliance with broader regulatory frameworks such as NIS2, which requires automated certificate lifecycle management (section 9.2.c)"* — rather than listing it alongside the Kubernetes hardening guides, or
2. Adding a brief qualifier to the ENISA entry noting that it covers organisational cryptographic policy rather than Kubernetes-specific hardening?

The rest of the changes look good:
- NSA/CISA joint attribution is correct
- BSI "IT-Grundschutz Compendium" is the official name; removing the `#page=475` anchor in favour of naming module APP.4.4 in-text is a reasonable editorial choice (page numbers can shift across editions)
- Removal of the Datree reference is justified — Datree was acquired by Run.ai in 2023 and the product was discontinued
- `.spelling` additions are correct

Xata:
@wallrj-cyberark 
Thank you for the review! Understood, reading those documents back to back makes the content blur together after awhile. I figured that the organisational policy level about certificate lifecycle management would have carried relevance.

I'll add the qualifier to that bullet noting that the doc covers organisational cryptographic policy instead of Kubernetes-specific hardening.

Xata:
Done. The latest commit now has a sentence clarifying that the ENISA NIS2 reference covers organizational policy rather than explicit Kubernetes hardening.

wallrj:
Thanks @Xata for the quick turnaround. (wallrj here — same Richard as wallrj-cyberark; this is my personal account.)

Reading the revision back, I don't think the qualifier quite rescues the entry:

- The intro now says *"The recommendations in this guide are informed by the following standards"*, which is a stronger claim than the old *"such as"* wording — and it isn't true of the ENISA guidance: none of the recommendations on this page were informed by it, and the document contains no Kubernetes content.
- A bullet that has to end with *"covers organisational cryptographic policy rather than Kubernetes-specific hardening"* is effectively telling the reader the reference doesn't belong in that list.
- The frontmatter description still makes the unqualified claim ("…CIS, NSA, BSI, and ENISA").

The EU-framework mention is worth having, though, so rather than trigger another round of review over wording, we'll push a small follow-up commit to your branch (you've allowed maintainer edits): it moves the ENISA reference into a short paragraph after the list — framed as option 1 from the earlier review, i.e. cert-manager helps with NIS2 compliance by automating certificate lifecycle management — reverts the intro to "such as", and adjusts the description to match. It also drops the "(v1.2, August 2022)" pin on the NSA guide, which would go stale. Then I'll approve so this can merge.

*with claude fable-5*

wallrj:
I've pushed the follow-up commit described above. Thanks again @Xata for the contribution and for engaging with the review — the EU regulatory angle is a valuable addition.

/approve
/lgtm
/label tide/merge-method-squash

*with claude fable-5*

Issue #197, opened 2020-05-04
Document ACME account mismatch


meyskens:
We should add how to troubleshoot the error `ACME server URL host and ACME private key registration host differ` to the documentation.
Probably under https://cert-manager.io/docs/faq/acme/

DannyMassa:
I recently moved my baremetal cluster to a new network. I think I'm running into this. What is the fix?

PR #2023, opened 2026-03-24
Adds troubleshooting guide for host missmatch error

epatmalnieks-gc:
There is an open [issue](https://github.com/cert-manager/website/issues/197) of missing troubleshooting info for `ACME server URL host and ACME private key registration host differ` This issue is mentioned also on [issue 3394](https://github.com/cert-manager/cert-manager/issues/3394) 

This PR adds a simple description of the error and what would be the reasonable place to look for.

### DIFF
--- a/content/docs/troubleshooting/acme.md
+++ b/content/docs/troubleshooting/acme.md
@@ -58,6 +58,7 @@ Events:
 * `Failed to update ACME account:400 urn:ietf:params:acme:error:invalidEmail`: the email you specified in the Issuer configuration isn't valid.
 * `Error initializing issuer: Failed to register ACME account: secrets "acme-key" already exists`: there might be a leftover account from a previous issuer that is no longer valid, you should remove the secret so it can be recreated.
 * `Error accepting challenge: 400 urn:ietf:params:acme:error:malformed: Unable to update challenge :: authorization must be pending`: this suggests that the authorization was not in 'pending' state at a time when cert-manager sent a request to the ACME server to accept the challenge. This may be because the domain validation has already failed and the authorization has been marked as 'invalid'. Check the authorization URL on the status of the `Order` or `Challenge` to see the status of the authorization and any additional information.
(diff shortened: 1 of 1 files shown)


wallrj-cyberark (changes requested):
Thank you for addressing issue #197 — this documentation gap has been open for years and is worth filling. The Copilot review has independently flagged the same concerns below.

I checked the [source code](https://github.com/cert-manager/cert-manager/blob/main/pkg/issuer/acme/setup.go#L290-L293) and found a few issues with the technical explanation.

**1. This is not an error — it is an info-level log message.**

The code at `pkg/issuer/acme/setup.go:291` uses `log.V(logf.InfoLevel).Info(...)`. When the condition triggers, cert-manager clears the cached account URI and re-checks the ACME registration. It is a self-healing action, not a failure.

**2. The explanation of the cause is inaccurate.**

The PR says *"host resolved by the node differs from the host in your private key"*. The private key does not contain a host. What the code actually compares (line 290) is `parsedAccountURL.Host` (from `issuer.status.acme.uri` — the ACME account URL cached from a previous registration) against `parsedServerURL.Host` (from `issuer.spec.acme.server` — the configured ACME server URL).

Common causes include:
- Changing the ACME server URL (e.g. from staging to production)
- A stale `search` domain in `/etc/resolv.conf` causing DNS to resolve the ACME server hostname differently (this is the scenario from cert-manager/cert-manager#3394)
- Reusing a private key Secret from a different Issuer or cluster

**3. Missing colon after the backtick-quoted message.**

The existing bullet points all use the pattern \`message\``: explanation`. The new entry is missing the colon separator.

A suggested replacement:

> `ACME server URL host and ACME private key registration host differ`: this is an informational message (not an error) indicating that the host in the configured ACME server URL (`spec.acme.server`) differs from the host where the ACME account was previously registered (`status.acme.uri`). cert-manager will automatically clear the cached registration and re-check the account. Common causes include: changing the ACME server URL (e.g. from staging to production), a stale `search` domain in `/etc/resolv.conf` (see cert-manager/cert-manager#3394), or reusing a private key Secret from a different Issuer. If the re-registration fails, check `/etc/resolv.conf` for incorrect search domains, or delete the private key Secret referenced by the Issuer to force a clean re-registration.

PR #2079, opened 2026-04-26
docs: update ListenerSet example to use http01-parentreffallback annotation

apkatsikas:
Updates the ListenerSet example in the Gateway usage guide to reflect the new `acme.cert-manager.io/http01-parentreffallback` annotation introduced in https://github.com/cert-manager/cert-manager/pull/8749

- Removes the HTTP listener from the ListenerSet (tenants declare HTTPS only)
- Adds the fallback annotation to the example ListenerSet
- Updates the generated Certificate to show Gateway parentRef annotations
- Adds the annotation to the Supported Annotations section

### DIFF
--- a/content/docs/usage/gateway.md
+++ b/content/docs/usage/gateway.md
@@ -572,6 +572,7 @@ spec:
   gatewayClassName: eg
   listeners:
     - name: http
+      hostname: dummy
       protocol: HTTP
       port: 80
   allowedListeners:
@@ -579,7 +580,9 @@ spec:
       from: All
 ```
 
-The Gateway's HTTP listener serves ACME HTTP-01 challenges for all tenant ListenerSets.
+The `dummy` listener is required because the `listeners` cannot be left empty.
+See the [issue 4425](https://github.com/kubernetes-sigs/gateway-api/issues/4425)
+in the Gateway API project to learn more.
 
 In their namespace, the developer creates a Service, ListenerSet, and HTTPRoute
 for their application:
@@ -660,7 +663,6 @@ metadata:
   namespace: dev
   annotations:
     cert-manager.io/cluster-issuer: letsencrypt
-    acme.cert-manager.io/http01-parentreffallback: "true"
 spec:
   parentRef:
     name: eg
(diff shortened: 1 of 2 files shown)


content/docs/usage/gateway.md
```diff
@@ -572,17 +572,14 @@ spec:
...
       port: 80
   allowedListeners:
     namespaces:
       from: All
 ```
 
-The `dummy` listener is required because the `listeners` cannot be left empty.
-See the [issue 4425](https://github.com/kubernetes-sigs/gateway-api/issues/4425)
-in the Gateway API project to learn more.
+The Gateway's HTTP listener serves ACME HTTP-01 challenges for all tenant ListenerSets.
```
ThatsMrTalbot: This is only true if you use the new annotation, I think we should document the new annotation as an alternative, keeping existing documentation as-is, as it is correct. We can add a section at the end on what the annotation does, and the new pattern this allows you to do.

There are many cases where you may _want_ the http listener on each listenerset, for example in the case where you want listernersets to completely own domains and their routing.

apkatsikas: https://github.com/cert-manager/website/pull/2079/changes/5ff7bf9d7a37b6748f2063a70084962976efbe47

apkatsikas: Hi @ThatsMrTalbot how does this look now?

content/docs/usage/gateway.md
```diff
@@ -716,6 +716,79 @@ ListenerSet resources support the same cert-manager annotations as Gateway
 resources. See the [Supported Annotations](#supported-annotations) section
 above.
 
+In addition, ListenerSet supports the following annotation:
+
+- `acme.cert-manager.io/http01-parentreffallback`: when set to `"true"`, cert-manager uses the ListenerSet's `spec.parentRef` (the parent Gateway) as the `parentRef` on the HTTPRoute it creates to serve ACME HTTP-01 challenges, instead of the ListenerSet itself.
+
+### Alternative: HTTPS-only ListenerSet with shared Gateway HTTP listener
```
ThatsMrTalbot: nit: The word alternative is confusing here, alternative to what? I think its a better heading if you just drop the word `Alternative:`

apkatsikas: https://github.com/cert-manager/website/pull/2079/changes/bd82e40b30bc33783a87071169c082e7ec70edc7

apkatsikas:
> `parentrefnamespace` needs adding to https://github.com/cert-manager/website/blob/master/.spelling

https://github.com/cert-manager/website/pull/2079/changes/a514648b957818903a57538136e7c3c4d1cf3bbb

Issue #1935, opened 2026-01-28
add third party cert-manager-webhook-infomaniak


M0NsTeRRR:
Hello,

I've developed this webhook as a third party alternative to the official one :
https://github.com/M0NsTeRRR/cert-manager-webhook-infomaniak

Differences between mine and the official one :
- **Distroless image**: Only the binary, running rootless for enhanced security.
- **Updated dependencies**: Regularly maintained for reliability and security.
- **Open collaboration**: Issues enabled for community feedback and contributions.
- **OCI-compatible artifacts**: Helm chart and container images are published in OCI format and signed, making them easy to integrate with GitOps tools like Flux and ArgoCD.

What is the proper way to add it to the current documentation I know which file to edit but how should I tag it as third party ?
Regards,

PR #2092, opened 2026-05-11
docs: add infomaniak third party provider

M0NsTeRRR:
Hello,

Differences between mine and the official one :
- **Distroless image**: Only the binary, running rootless for enhanced security.
- **Updated dependencies**: Regularly maintained for reliability and security.
- **Open collaboration**: Issues enabled for community feedback and contributions.
- **OCI-compatible artifacts**: Helm chart and container images are published in OCI format and signed, making them easy to integrate with GitOps tools like Flux and ArgoCD.

Fix issue #1935

### DIFF
--- a/content/docs/configuration/acme/dns01/README.md
+++ b/content/docs/configuration/acme/dns01/README.md
@@ -173,6 +173,7 @@ Links to these supported providers along with their documentation are below:
 - [`cert-manager-webhook-hetzner`](https://github.com/vadimkim/cert-manager-webhook-hetzner)
 - [`cert-manager-webhook-ibmcis`](https://github.com/jb-dk/cert-manager-webhook-ibmcis)
 - [`cert-manager-webhook-infomaniak`](https://github.com/Infomaniak/cert-manager-webhook-infomaniak)
+- [`cert-manager-webhook-infomaniak`](https://github.com/M0NsTeRRR/cert-manager-webhook-infomaniak) (third-party alternative)
 - [`cert-manager-webhook-inwx`](https://gitlab.com/smueller18/cert-manager-webhook-inwx)
 - [`cert-manager-webhook-ionos-cloud`](https://github.com/ionos-cloud/cert-manager-webhook-ionos-cloud)
 - [`cert-manager-webhook-linode`](https://github.com/linode/cert-manager-webhook-linode)


wallrj-cyberark (commented):
Thank you for the contribution. Rather than listing a second webhook with the same name, I would suggest contributing your improvements to the [original Infomaniak webhook](https://github.com/Infomaniak/cert-manager-webhook-infomaniak) instead.

The improvements you describe (distroless image, rootless security context, updated dependencies, OCI artifacts) would benefit all existing users of that webhook. The original repository accepts external contributions — it has [14 merged PRs](https://github.com/Infomaniak/cert-manager-webhook-infomaniak/pulls?q=is%3Apr+is%3Aclosed+is%3Amerged) from community contributors. In fact, some of your stated improvements align with work already requested there:

- [#5](https://github.com/Infomaniak/cert-manager-webhook-infomaniak/issues/5): "feat: add default security context" — directly related to your rootless/distroless changes
- There are also [5 open dependabot PRs](https://github.com/Infomaniak/cert-manager-webhook-infomaniak/pulls?q=is%3Apr+is%3Aopen+dependabot) for dependency updates that could use a reviewer

Listing two entries with the identical display name `cert-manager-webhook-infomaniak` would be confusing for readers, and the "(third-party alternative)" annotation does not resolve this — every entry in this list is third-party.

I would encourage you to open PRs upstream and, if the maintainers are unresponsive after a reasonable period, revisit the question of a separate listing at that point.

M0NsTeRRR:
Hello,

Thank you for your input. However, I don’t agree with some parts of your conclusion. I thought about this before creating an alternative project and upstreaming the issue I addressed. The problem is the lack of dependency updates, which is on the maintainer’s side. Even if I reviewed it, the maintainer would still need to review it as well, especially regarding trustworthiness. I’ve already contacted Infomaniak support, provided them with what I did, and they said they would forward it to their appropriate team and I got no answer for now.

Also, like many users (and I assume you as well), before creating a PR for any project I contribute to, I first reach out to the maintainers to see if they’re interested in this kind of evolution. This way, I avoid wasting my time if they don’t agree. Since the issues are closed, I can’t do that.

> Listing two entries with the identical display name cert-manager-webhook-infomaniak would be confusing for readers, and the "(third-party alternative)" annotation does not resolve this — every entry in this list is third-party.

Exactly, that’s why I also opened the issue first and created the PR in the meantime, it didn’t take much time.

PR #2097, opened 2026-05-13
Clarify ACME challenge scheduling behaviour

wallrj:
## What this PR does

Clarifies the ACME challenge scheduling documentation.

This updates the ACME concepts, configuration, and troubleshooting pages to explain that challenge scheduling is intentionally conservative, and that the scheduler should be understood as a coarse back-pressure mechanism rather than a fairness or policy layer.

Related to cert-manager/cert-manager#8781 and cert-manager/cert-manager#8643.

## Why

The current documentation explains what the scheduler does, but not clearly why it behaves this way.

In particular, this PR adds context that:

- cert-manager self-checks from a single network and DNS viewpoint;
- different solver backends do not reliably imply independent ACME-visible validation paths;
- the scheduler therefore keeps a coarse conflict key by design; and
- multi-tenant isolation and DNS-name ownership constraints are better handled through policy, approval, admission, or separate deployments.

It also adds cross-links from the ACME configuration and troubleshooting pages back to the concepts page, and to the policy documentation.

### DIFF
--- a/content/docs/concepts/acme-orders-challenges.md
+++ b/content/docs/concepts/acme-orders-challenges.md
@@ -92,9 +92,28 @@ is a backlog of challenges to complete.
 Instead of attempting to process all challenges at once, challenges are
 'scheduled' by cert-manager.
 
-This scheduler applies a cap on the maximum number of simultaneous challenges
-as well as disallows two challenges for the same DNS name and solver type
-(`HTTP01` or `DNS01`) to be completed at once.
+This scheduler is a coarse back-pressure mechanism. It applies a cap on the
+maximum number of simultaneous challenges, and it also disallows two
+challenges for the same DNS name and challenge type (`HTTP01` or `DNS01`) from
+being completed at once.
 
-The maximum number of challenges that can be processed at a time is 60 as of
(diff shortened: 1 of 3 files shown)


content/docs/concepts/acme-orders-challenges.md
```diff
@@ -92,9 +92,28 @@ is a backlog of challenges to complete.
...
 'scheduled' by cert-manager.
 
-This scheduler applies a cap on the maximum number of simultaneous challenges
-as well as disallows two challenges for the same DNS name and solver type
-(`HTTP01` or `DNS01`) to be completed at once.
-
-The maximum number of challenges that can be processed at a time is 60 as of
-[`ddff78`](https://github.com/cert-manager/cert-manager/blob/ddff78f011558e64186d61f7c693edced1496afa/pkg/controller/acmechallenges/scheduler/scheduler.go#L31-L33).
\ No newline at end of file
+This scheduler is a coarse back-pressure mechanism. It applies a cap on the
```
wallrj: I have expanded this section because the corresponding code change in cert-manager/cert-manager#8781 is primarily documenting design intent rather than altering behaviour. The aim here is to make the same reasoning visible to documentation readers: the scheduler is intentionally conservative because a single cert-manager instance performs self-checks from one network and DNS viewpoint, and this is not the right layer to enforce multi-tenant isolation or DNS-name ownership policy.

lunarwhite (approved):
/lgtm

PR #2112, opened 2026-05-26
Add threat modelling section to devops tips

SgtCoDFish:
This includes a lot of new handwritten detail explaining how we think about RBAC including practical tips on what people should do

Preview of the page: https://deploy-preview-2112--cert-manager.netlify.app/docs/devops-tips/threat-modelling/

### DIFF
--- a/content/docs/devops-tips/threat-modelling.md
+++ b/content/docs/devops-tips/threat-modelling.md
@@ -1,0 +1,93 @@
+---
+title: Threat Model
+description: How we think about permissions and threats in cert-manager
+---
+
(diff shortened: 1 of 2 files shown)


content/docs/devops-tips/threat-modelling.md
```diff
@@ -0,0 +1,93 @@
+---
+title: cert-manager's Threat Model
```
hjoshi123: Should we just call it `Threat Model` to be consistent with other articles in the section?

SgtCoDFish: Yeah fair enough - changed!

PR #2117, opened 2026-05-29
Add Venafi OAuth token metrics and AuthFailed condition docs

FelixPhipps:
Documents the two new Prometheus metrics added in https://github.com/cert-manager/cert-manager/pull/8808 (certmanager_venafi_oauth_token_requests_total and certmanager_venafi_oauth_token_request_duration_seconds) in the Prometheus metrics page, and adds an "Issuer conditions" section to the Venafi configuration docs explaining the new AuthFailed condition reason and how it differs from ErrorSetup.

### DIFF
--- a/content/docs/configuration/venafi.md
+++ b/content/docs/configuration/venafi.md
@@ -367,6 +367,25 @@ more information on how to create Certificate resources.
 | `tokenEndpoint` | No | `https://auth.apps.paloaltonetworks.com/oauth2/access_token` | OAuth 2.0 token endpoint URL used to obtain access tokens. |
 | `url` | No | `https://api.strata.paloaltonetworks.com/ngts` | Base URL for the NGTS API endpoint. |
 
+## Issuer conditions
+
+### AuthFailed
+
+When a Venafi endpoint rejects the supplied credentials (for example, with an HTTP 401 or 403 response), the `Issuer` or `ClusterIssuer` Ready condition transitions to `False` with `reason: AuthFailed`:
+
+```bash
+$ kubectl describe issuer corp-issuer --namespace='NAMESPACE OF YOUR ISSUER RESOURCE'
+...
+Status:
+  Conditions:
+    Message: OAuth token request failed: ...
+    Reason:  AuthFailed
+    Status:  False
+    Type:    Ready
+```
+
+`AuthFailed` indicates that the credentials themselves are invalid. This is distinct from the generic `ErrorSetup` reason, which covers transient network or infrastructure problems. If you see `AuthFailed`, check the credentials in the referenced `Secret` — the API key for CyberArk Certificate Manager SaaS, the access token or username/password for CyberArk Certificate Manager Self-Hosted, or the `client-id` and `client-secret` for NGTS — and ensure they are correct and have not expired.
+
 ## Issuer specific annotations
 
 ### Custom Fields
--- a/content/docs/configuration/venafi.md
+++ b/content/docs/configuration/venafi.md
@@ -384,7 +384,7 @@ Status:
     Type:    Ready
 ```
 
-`AuthFailed` indicates that the credentials themselves are invalid. This is distinct from the generic `ErrorSetup` reason, which covers transient network or infrastructure problems. If you see `AuthFailed`, check the credentials in the referenced `Secret` — the API key for CyberArk Certificate Manager SaaS, the access token or username/password for CyberArk Certificate Manager Self-Hosted, or the `client-id` and `client-secret` for NGTS — and ensure they are correct and have not expired.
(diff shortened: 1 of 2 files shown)


content/docs/configuration/venafi.md
```diff
@@ -281,6 +281,25 @@ You are now ready to issue certificates using the newly provisioned CyberArk
...
+...
+Status:
+  Conditions:
+    Message: OAuth token request failed: ...
+    Reason:  AuthFailed
+    Status:  False
+    Type:    Ready
+```
+
+`AuthFailed` indicates that the credentials themselves are invalid. This is distinct from the generic `ErrorSetup` reason, which covers transient network or infrastructure problems. If you see `AuthFailed`, check the credentials in the referenced `Secret` — the API key for CyberArk Certificate Manager SaaS, the access token or username/password for CyberArk Certificate Manager Self-Hosted, or the `client-id` and `client-secret` for Next-Gen Trust Security (NGTS) — and ensure they are correct and have not expired.
```
maelvls: `AuthFailed` is self-explanatory, is it really worth documenting it? I'd suggest leaving the entire "Issuer Conditions" section out.

content/docs/devops-tips/prometheus-metrics.md
```diff
@@ -179,6 +179,17 @@ Check the health of the cert-manager scrape targets on the Prometheus status pag
 
 ![](/docs/devops-tips/prometheus-metrics/prometheus-status-targets.png)
 
+## Venafi metrics
+
+cert-manager exposes the following Prometheus metrics for Venafi OAuth token requests. These are emitted by the controller whenever a Venafi `Issuer` or `ClusterIssuer` authenticates against a TPP, Cloud, or Next-Gen Trust Security (NGTS) endpoint.
+
+| Metric | Type | Labels | Description |
+|--------|------|--------|-------------|
+| `certmanager_venafi_oauth_token_requests_total` | Counter | `status` (`success` \| `failure`) | Total number of Venafi OAuth token requests made by cert-manager. |
```
maelvls, line 188: I'm surprised `certmanager_venafi_oauth_token_requests_total`'s `status` is "success" and "failure". I would have expected the HTTP status code instead. That would give a bit more information: 403 = probably misconfiguration, 500 = probably something wrong with the backend.

But maybe success/failure makes sense, I don't know.

maelvls (approved):
I left a comment about success/failure vs. HTTP status code, but feel free to ignore if you feel it doesn't add much.

content/docs/devops-tips/prometheus-metrics.md
FelixPhipps: Success/Failure is what we're using in the implementation as oppose to status codes.

maelvls: But would it be worth going with status codes instead of success/failure in the implementation? I know I come a bit late (I didn't review the other PR), I am just wondering if it would be a little more usable with HTTP status codes.

Unless the auth flow contains multiple HTTP calls, and in that case it wouldn't make sense?

FelixPhipps: The token endpoint is a single call, so status codes would technically work, but success/failure also covers non-HTTP failures like network timeouts. 

PR #8808 is already merged so changing labels now would be a breaking change.

maelvls: No problem in changing what's in `master`, we can break it as much as we want until we cut the first beta. 👍

Non-HTTP errors returned by the OS (io timeout, can't reach server, and such) can be captured in the same label:

```
certmanager_venafi_oauth_token_requests_total{path="/oauth2/access_token", status="200"} 4902
certmanager_venafi_oauth_token_requests_total{path="/oauth2/access_token", status="200"} 45
certmanager_venafi_oauth_token_requests_total{path="/oauth2/access_token", status="net_error"} 3
```

or something like that.

example: https://github.com/jetstack/venafi-connection-lib/blob/443dd7e48a1aec2d75c071e996c06667f0d656d2/http_client/roundtripper_metrics.go#L76 (link private to Palo Alto)

maelvls: I am not super hyped by these two metrics because this information is already available on the issuer/clusterissuer itself, and these already have metrics. The two new metrics don't any more information and don't add value IMO. And the epic didn't say anything about metrics:

> Token request failures and successes can be tracked by reviewing the logs at the corresponding log level.
>
> The Ready condition status for Issuer/ClusterIssuer should accurately indicate OAuth auth readiness. For instance, if a token cannot be obtained, the status should be set to NotReady with a reason such as AuthFailed or a similar equivalent.

I don't know why the ticket under the epic started talking about metrics. I guess AI just went "oh well let's have everything we can think of just in case".

I've had a chat with Felix, and we decided to keep the two metrics in cert-manager (https://github.com/cert-manager/cert-manager/pull/8808) as Felix is going on vacation and removing the metrics would take work.

fyi @inteon

maelvls:
This has been merged to `master` instead of `release-next`.

- Revert this PR: https://github.com/cert-manager/website/pull/2137
- Create a new PR against `release-next`: ...

PR #2114, opened 2026-05-27
Add NGTS issuer documentation

FelixPhipps:
Adds a new "Creating an Issuer for Palo Alto Networks Next Generation Trust Services" section to the Venafi configuration docs, covering OAuth 2.0 Client Credentials setup, an ngts stanza YAML example, and a field reference table. 

Also adds ngts to the spelling dictionary.

PR #2140, opened 2026-06-08
Fix wording: NGTS stands for Security, not Services

maelvls:
#2114 documented "NGTS" as "Next Generation Trust Services". As pointed out by Ben, it's "Security", not "Services".

The link to SASE is also wrong, it should point to NGTS instead (https://www.paloaltonetworks.com/network-security/next-gen-trust-security).

### DIFF
--- a/content/docs/configuration/venafi.md
+++ b/content/docs/configuration/venafi.md
@@ -5,9 +5,10 @@ description: 'cert-manager configuration: CyberArk Issuers'
 
 ## Introduction
 
-The CyberArk `Issuer` obtains certificates from
-[CyberArk Certificate Manager](https://www.cyberark.com/products/certificate-manager/) SaaS or Self-Hosted,
-or from [Palo Alto Networks Next Generation Trust Services (NGTS)](https://www.paloaltonetworks.com/sase).
+The CyberArk `Issuer` obtains certificates from [CyberArk Certificate
+Manager](https://www.cyberark.com/products/certificate-manager/) SaaS or
+Self-Hosted, or from [Palo Alto Networks Next-Generation Trust Security
+(NGTS)][ngts].
 
 The `Issuer` was formerly known as the Venafi `Issuer`, and for backwards compatibility reasons is configured using older product names - "Venafi Cloud" corresponds to CyberArk Certificate Manager SaaS and "Venafi TPP" corresponds to CyberArk Certificate Manager Self-Hosted.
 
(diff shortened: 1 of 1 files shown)


content/docs/configuration/venafi.md
```diff
@@ -283,12 +284,14 @@ You are now ready to issue certificates using the newly provisioned CyberArk
...
+## Creating an Issuer for Palo Alto Networks Next-Generation Trust Security
 
-The CyberArk `Issuer` supports [Palo Alto Networks Next Generation Trust
-Services (NGTS)](https://www.paloaltonetworks.com/sase), a cloud-native
-certificate management platform. Authentication uses OAuth 2.0 Client
-Credentials, so no API key or username/password is required.
+The CyberArk `Issuer` supports [Palo Alto Networks Next-Generation Trust
+Security (NGTS)][ngts], a cloud-native certificate management platform.
+Authentication uses OAuth 2.0 Client Credentials, so no API key or
+username/password is required.
```
wallrj-cyberark, line 292:
```suggestion
Authentication uses OAuth 2.0 Client Credentials.
```

I'd drop that last part, personally. It seems unnecessary. Why would the user think that an API key or u/p ARE required?

content/docs/configuration/venafi.md
```diff
@@ -317,7 +320,7 @@ or `ClusterIssuer` resource. If you are creating a `ClusterIssuer` resource,
 you must change the `kind` field to `ClusterIssuer` and remove the
 `metadata.namespace` field.
 
-The `zone` is the name of the Certificate Issuing Template (CIT) in NGTS that will be used to issue certificates. Unlike CyberArk Certificate Manager SaaS, NGTS does not use an Application prefix — the zone is just the CIT name.
+The `zone` is the name of the Certificate Issuing Template (CIT) in NGTS that will be used to issue certificates. Unlike CyberArk Certificate Manager SaaS which requires both an Application name and CIT alias in the format `Application\CIT`, NGTS only requires the CIT name.
```
wallrj-cyberark, line 323:
```suggestion
For NGTS, the `zone` is **only** the name of the Certificate Issuing Template (CIT), that will be used to issue certificates. 

> ℹ️ If you are migrating from CyberArk Certificate Manager SaaS or Venafi Cloud, notice that `zone` has a different format. Zone MUST be `CIT`, NOT `Application\CIT`.
```

I'm not sure if this is clearer. I imagine that most users will skip straight to the NGTS section and may have no knowledge of how CM-SaaS zones are configured....unless they're migrating.

wallrj-cyberark (approved):
A couple of optional suggestions.

/hold in case you want to address those.

/cc @wallrj

wallrj:
/unhold 

My suggestions can be added in another PR, if necessary.

content/docs/configuration/venafi.md
maelvls: Good point. Your suggestion looks clearer.

PR #2255, opened 2026-09-01
docs: document OpenShift incompatibility with ACMEHTTP01IngressPathTypeExact

m-altaifi:
Documents the OpenShift incompatibility with the `ACMEHTTP01IngressPathTypeExact` feature gate, as a new subsection of the existing "ACME HTTP01 challenge paths now use `PathType` `Exact` in Ingress routes" section in the v1.18 release notes.

OpenShift's `route-controller-manager` converts `Ingress` resources into `Route` resources for the built-in HAProxy router, but it [skips any rule which uses `pathType: Exact`](https://github.com/openshift/route-controller-manager/blob/59697cf7af4517dd44e28179a57f7f35b6ea0e22/pkg/route/ingress/ingress.go#L460-L464) — the `Route` API has no equivalent exact path matching mode. The HTTP-01 solver `Ingress` is therefore never converted into a `Route`, so the challenge URL is not served and the challenge fails with a 503.

The new subsection mirrors the `ingress-nginx` treatment directly above it: what breaks, why, and the feature-gate workaround. It also notes that the cert-manager Operator for Red Hat OpenShift already disables the gate by default from its 1.18.0 release, so users of that distribution are unaffected, and that the durable fix is `Exact` path type support in OpenShift's Ingress-to-Route conversion.

This replaces cert-manager/cert-manager#9219, which documented the same thing in the cert-manager repository. @lunarwhite and @erikgb pointed out there that the website is the right place for it:

> As I said: cert-manager/cert-manager#9210 (comment), I'm afraid having it documented in cert-manager codebase would be way too narrow. If we really want to document something, I suggest open a PR against doc site

I will close #9219 in favour of this.

Refs cert-manager/cert-manager#9210

**Testing**

`npm ci` then, against the changed file: `cspell` 0 issues, `remark --frail` clean, `eslint` clean, `markdown-link-check` clean.

### DIFF
--- a/content/docs/installation/compatibility.md
+++ b/content/docs/installation/compatibility.md
@@ -114,3 +114,34 @@ the webhook listens. See the warning at the top of this page for more details.
 Because Fargate forces you to use its networking, you cannot manually set the networking
 type and options such as `webhook.hostNetwork` on the helm chart will cause your
 cert-manager deployment to fail in surprising ways.
+
+## OpenShift
+
+### HTTP01 challenges fail with a 503 when using the built-in router
+
+Since cert-manager 1.18, the HTTP01 solver `Ingress` uses `pathType: Exact`
+for the challenge path (feature gate `ACMEHTTP01IngressPathTypeExact`, on by
+default). OpenShift's `route-controller-manager` converts `Ingress` resources
+into `Route` resources for the built-in HAProxy router, but it
+[skips any rule which uses `pathType: Exact`](https://github.com/openshift/route-controller-manager/blob/59697cf7af4517dd44e28179a57f7f35b6ea0e22/pkg/route/ingress/ingress.go#L460-L464),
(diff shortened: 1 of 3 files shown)


m-altaifi:
@JoshVanL

wallrj (approved):
Thank you. I pushed one commit to your branch so that OpenShift users can find this without reading the 1.18 release notes:

- New "OpenShift" section on `content/docs/installation/compatibility.md`, alongside the GKE and EKS sections, holding the explanation, the feature gate values and the Operator note.
- New "Got 503 status code" entry under HTTP01 troubleshooting in `content/docs/troubleshooting/acme.md`, linking to it.
- The release-note subsection is now a three-line pointer to the compatibility page, with the Copilot grammar fix applied.

cspell, remark, eslint and markdown-link-check are clean on the three files.

/approve
/lgtm

[Claude Fable 5]

PR #2271, opened 2026-09-07
Explain what to do when an application expects ca.crt in the certificate Secret

### LINKED ISSUE #222: [Feature] - Ability to inject a CA cert into a cert-manager managed secret resource
After reading through this cert-manager issue: https://github.com/cert-manager/cert-manager/issues/1571 I had a temporary bright spark, although the actual solution might not be the one suggested here.

## tl;dr

A single secret where cert-manager controlls `tls.crt` and `tls.key`, but `ca.crt` is injected from trust-manager.

## Problems

1) The pattern of a single secret containing `tls.crt`, `tls.key` and `ca.crt` seems to have become a standard for many tools mounting certificates in kubernetes.
2) There is an inconsistency that cert-manager issuers don't always provide a ca.crt and the current position is that this is no a pattern we want to promote.
3) A CA might not pass back a CA.crt, and the CA.crt might not be what the user actually wants anyway.
4) Using both cert-manager and trust-manager with a Secret + ConfigMap might sounds like a complete solution, but refer back to no1 where any number of k8s operators might not only work with 1 secret with defined keys. (redis operator, opensearch etc..)

## Goals / Themes

1) We want user to be able to adopt cert-manager and trust-manager broadly across their k8s applications
2) We want to maintain certificate issuance and trust distribution via separate processes.
3) Ease of use and automation, eg. avoid having to manually or have custom scripts to combine secrets, rename fields etc..

## Solution(s)

The main idea that struck me was off the back of [this comment](https://github.com/cert-manager/cert-manager/issues/1571#issuecomment-784979272), specifcially:

> Or even just a flag so we can manually set the ca.crt contents and cert-manager will include this in the secret as ca.crt

This seems a valid solution, could you manually insert the `ca.crt` into the secret itself. But why do this manually, trust-manager is about automating trust distribution. So what if trust-manager had a watch / trigger for `Secrets` that required a certain `Bundle` output to be inserted.

Just an idea but what about:

```yaml
kind: Secret
metadata:
  annotations:
    trust.cert-manager.io/ca-inject: "true"
    trust.cert-manager.io/bundle: "my-bundle"
    cert-manager.io/alt-names: alpha.dev.peter-fiddes-gcp.jetstacker.net
    cert-manager.io/certificate-name: alpha-acme
    ...
```

Annotations are terrible? and there is probably a better way. But with `ManagedFields` it's now possible to safely have multiple controllers manage a resource (or so I believe). This could potentially be added in stages to the two controllers. For example if trust-manager was to watch secrets with valid annotations, then these annotations could be added using `certificate.spec.secretTemplate.annotations`.. eg..

```yaml
spec:
  secretTemplate:
    annotations:
      trust.cert-manager.io/ca-inject: "true"
      trust.cert-manager.io/bundle: "my-bundle"
```

Basically no new cert-manager certificate CRD fields to link to bundle.. but perhaps in the future:

```yaml
spec:
  caBundleRef: "my-bundle"
```

Assuming it was as above, you could potentially avoid annotations and have cert-manager ask trust-manager for the bundle ca at issuance time. Or alternatively, just mark the `Secret` field `ca.crt` in some fashion for trust-manager to do the injection.

## Other solutions

- Utilise another controller / tool to bring the two parts together, similar to how [secrets-transform](https://github.com/maelvls/secret-transform) currently manages extra keys. "pki-combinator"? :joy: 

## Concerns

- CA certs are not secret and probably should be in `Secrets`, but then again, a workload certificate `tls.crt` arguably doesn't need to be secret either, only the key.
- It's a completely new distribution mechanism for trust-manager, adding a new resource to watch & keep updated.
- What if the `ca.crt` changes from the bundle being propagated, would you want cert-manager to reissue the actual cert as well? I think this is combated with good practice.

## Other

I haven't fully thought this out, just wanted to get a brain dump out and collaboratively find the best solution. Also see if this is a good idea, so please vote / comment with input.

wallrj:
Rendered preview:

- [Trusting certificates: My application expects `ca.crt` in the same Secret as the certificate](https://deploy-preview-2271--cert-manager.netlify.app/docs/trust/#ca-crt-in-same-secret)
- [FAQ: Why isn't my certificate's chain in my issued Secret's `ca.crt`?](https://deploy-preview-2271--cert-manager.netlify.app/docs/faq/#chain-cacrt)

Fixes: https://github.com/cert-manager/cert-manager/issues/1571
Fixes: https://github.com/cert-manager/trust-manager/issues/222

The FAQ and the trust page explain why `ca.crt` must not be used as a trust store. They do not say what to do when a chart or operator insists on a single Secret with `tls.crt`, `tls.key` and `ca.crt`. That is the recurring question behind cert-manager/cert-manager#1571 and cert-manager/trust-manager#222, and it has kept both issues alive for years.

This PR adds a section to the trust page, "My application expects `ca.crt` in the same Secret as the certificate", and links to it from the FAQ entry. The section:

- explains that `ca.crt` is usually a trust store for client or peer authentication, so a public CA must never be used there;
- lists the applications users report and the separate CA option each one already has (OpenSearch operator `caSecret.name`, RabbitMQ cluster operator `caSecretName`, Istio Gateway `<credentialName>-cacert`, ingress-nginx `auth-tls-secret`, Gateway API `BackendTLSPolicy`);
- shows a projected volume that merges the cert-manager Secret and the trust-manager ConfigMap into one directory;
- explains why writing `ca.crt` into the cert-manager Secret from another controller is unsafe.

<details>
<summary>Evidence for the claims in the new section</summary>

- cert-manager only writes `ca.crt` when the issuer returns a CA: [issuing/internal/secret.go#L165-L167](https://github.com/cert-manager/cert-manager/blob/79dd67caae8e6ea319ac9cf7f0a380b77c3a9f2b/pkg/controller/certificates/issuing/internal/secret.go#L165-L167).
- The keystore check treats the presence of `ca.crt` in the Secret as "issuer provides CA" ([policies/checks.go#L107](https://github.com/cert-manager/cert-manager/blob/79dd67caae8e6ea319ac9cf7f0a380b77c3a9f2b/internal/controller/certificates/policies/checks.go#L107)), but truststores are built from the CertificateRequest CA ([secret.go#L305](https://github.com/cert-manager/cert-manager/blob/79dd67caae8e6ea319ac9cf7f0a380b77c3a9f2b/pkg/controller/certificates/issuing/internal/secret.go#L305), [#L363](https://github.com/cert-manager/cert-manager/blob/79dd67caae8e6ea319ac9cf7f0a380b77c3a9f2b/pkg/controller/certificates/issuing/internal/secret.go#L363)). An injected `ca.crt` with keystores enabled therefore never satisfies the check.
- OpenSearch operator `caSecret.name`: [userguide v2.8.0 line 214](https://github.com/opensearch-project/opensearch-k8s-operator/blob/v2.8.0/docs/userguide/main.md?plain=1#L214).
- RabbitMQ cluster operator `caSecretName`: [rabbitmqcluster_types.go v2.22.5 line 374](https://github.com/rabbitmq/cluster-operator/blob/v2.22.5/api/v1beta1/rabbitmqcluster_types.go#L374).
- Redis `tls-ca-cert-file` authenticates clients and peers: [redis.conf 8.10.1 lines 233-237](https://github.com/redis/redis/blob/8.10.1/redis.conf#L233-L237).

</details>

[Claude Fable 5.1]

### DIFF
--- a/content/docs/trust/README.md
+++ b/content/docs/trust/README.md
@@ -108,8 +108,10 @@ It is tempting to have a second controller, a mutating webhook or a Job add
 `ca.crt` to the Secret that cert-manager manages. Do not do this. cert-manager
 reconciles that Secret and reads it back to decide whether to reissue. For example,
 if `keystores` are enabled, cert-manager treats the presence of `ca.crt` as proof that
-the issuer provided a CA and expects a matching truststore, which it cannot create
-for an ACME issuer, so it reissues on every reconcile.
+the issuer provided a CA and expects a matching trust store. cert-manager builds
+trust stores from the CA returned by the issuer, not from the Secret, and an ACME
+issuer returns none. The trust store is never written, so the check fails and
+cert-manager reissues on every reconcile.
 
 If the application accepts only a single Secret with all three keys and offers no
 separate CA option, ask the project to add one, and link to this page. In the
--- a/content/docs/trust/README.md
+++ b/content/docs/trust/README.md
@@ -28,3 +28,91 @@ the client will fail to connect to the compromised server.
 
 The same concept also applies when configuring a server for mutually-authenticated TLS;
 don't give the server access to Secret containing the client certificate and private key.
+
+<a id="ca-crt-in-same-secret"></a>
+## My application expects `ca.crt` in the same Secret as the certificate
+
+Many Helm charts and operators assume a `kubernetes.io/tls` Secret with three keys:
+`tls.crt`, `tls.key` and `ca.crt`. cert-manager only writes `ca.crt` when the issuer
+reports a CA. The ACME issuer never does, so that Secret has no `ca.crt`, and the
+application refuses to start or the volume mount fails with an error like
+`references non-existent secret key: ca.crt`.
+
+The question to answer first is what the application does with `ca.crt`. For most
+applications it is a trust store: the set of CAs allowed to sign client or peer
+certificates. For example, OpenSearch nodes use it to authenticate other nodes and
+admin clients, and Redis uses `tls-ca-cert-file` to
+[authenticate clients and replicas](https://github.com/redis/redis/blob/8.10.1/redis.conf#L233-L237).
+A trust store must contain only CAs you control. If the certificate was issued by a
+public CA such as Let's Encrypt, the issuing CA must not be used as a trust store,
+because anyone with a certificate from that CA could then authenticate to your
+application.
+
+So the CA you want in `ca.crt` is a decision you make, and its lifecycle is separate
+from the leaf certificate. Distribute it with trust-manager and connect the two
+Secrets in one of these ways.
+
+### Give the application a separate CA Secret
+
+Check whether the application accepts the CA separately. Many of the applications
+users report do:
+
+- The OpenSearch operator has
+  [`caSecret.name`](https://github.com/opensearch-project/opensearch-k8s-operator/blob/v2.8.0/docs/userguide/main.md?plain=1#L214)
+  next to `secret.name`.
+- The RabbitMQ cluster operator has
+  [`tls.caSecretName`](https://github.com/rabbitmq/cluster-operator/blob/v2.22.5/api/v1beta1/rabbitmqcluster_types.go#L374)
+  next to `tls.secretName`.
+- Istio Gateway reads the CA for mutual TLS from a Secret named
+  [`<credentialName>-cacert`](https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings).
+- ingress-nginx reads the client CA from the Secret named in the
+  [`auth-tls-secret`](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#client-certificate-authentication)
+  annotation.
+- Gateway API `BackendTLSPolicy` reads CAs from a
+  [ConfigMap](https://gateway-api.sigs.k8s.io/api-types/backendtlspolicy/).
+
+Point those fields at a trust-manager `Bundle` target. `Bundle` targets are
+ConfigMaps by default; enable Secret targets in trust-manager if the application
+only accepts a Secret. See [Targets](./trust-manager/README.md#targets).
+
+### Merge the two into one volume
+
+If you write the Pod spec yourself, a
+[projected volume](https://kubernetes.io/docs/concepts/storage/projected-volumes/)
+presents the cert-manager Secret and the trust-manager ConfigMap as one directory.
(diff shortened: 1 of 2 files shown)


content/docs/trust/README.md
```diff
@@ -28,3 +28,91 @@ the client will fail to connect to the compromised server.
...
+```
+
+### Do not write `ca.crt` into the cert-manager Secret
+
+It is tempting to have a second controller, a mutating webhook or a Job add
+`ca.crt` to the Secret that cert-manager manages. Do not do this. cert-manager
+reconciles that Secret and reads it back to decide whether to reissue. For example,
+if `keystores` are enabled, cert-manager treats the presence of `ca.crt` as proof that
+the issuer provided a CA and expects a matching truststore, which it cannot create
+for an ACME issuer, so it reissues on every reconcile.
```
lunarwhite:
```suggestion
the issuer provided a CA and expects a matching truststore, which it rebuilds from
whatever `ca.crt` it finds, then drops again on the next renewal since ACME supplies none.
```

lunarwhite (approved):
/lgtm

only a wording suggestion, please feel free to take or leave it as-is.

content/docs/trust/README.md
wallrj: Thank you for the review. I have not taken the suggestion because it describes behaviour cert-manager does not have, but it showed me the sentence could be read that way, so I have reworded it in b6994959.

cert-manager does not rebuild the trust store from the `ca.crt` it finds in the Secret. It builds trust stores from the CA returned by the issuer on the CertificateRequest, and an ACME issuer returns none:

- the check reads only the presence of `ca.crt` in the Secret: [checks.go#L107](https://github.com/cert-manager/cert-manager/blob/79dd67caae8e6ea319ac9cf7f0a380b77c3a9f2b/internal/controller/certificates/policies/checks.go#L107)
- the trust store is written from `data.CA`, which comes from the CertificateRequest: [secret.go#L305](https://github.com/cert-manager/cert-manager/blob/79dd67caae8e6ea319ac9cf7f0a380b77c3a9f2b/pkg/controller/certificates/issuing/internal/secret.go#L305) and [secret.go#L363](https://github.com/cert-manager/cert-manager/blob/79dd67caae8e6ea319ac9cf7f0a380b77c3a9f2b/pkg/controller/certificates/issuing/internal/secret.go#L363)

So the trust store is never written, the check never passes, and cert-manager reissues on every reconcile. The new wording says that directly. The new commit will have dropped the lgtm label; please re-add it if you are happy with the change.

[Claude Fable 5.1]

PR #1785, opened 2025-09-28
Add release-notes generator script and Makefile target

wallrj-cyberark:
Producing release notes has always been one of the most laborious parts of the cert-manager release process. This PR adds a Go tool that automates the tedious, error-prone steps — extracting the changelog from GitHub PR history, formatting contributor and maintainer lists, and maintaining the structured MDX markers the website expects — so we spend less time on formatting and more time reviewing content.

## What this adds

**`scripts/release-notes/` — Go CLI tool**

- Uses `k8s.io/release/pkg/notes` to collect changelog entries from merged PRs between two git revisions.
- Auto-detects the end revision: tries the release tag (`v1.21.0`), then the release branch (`release-1.21`), then falls back to `master`. The tool can therefore be run before the tag exists.
- Generates a new release-notes file from a template (if not already present) with placeholder sections for summary, themes, contributors, maintainers, steering committee, and changelog.
- On subsequent runs, updates only the relevant sections (contributors, maintainers, steerers, changelog), leaving hand-edited sections untouched.
- Handles legacy files (created before this tool existed) by inserting MDX markers using structural heuristics, then regenerating the changelog section.
- Quotes PR numbers and GitHub handles for spell-checker compatibility.

**`make/02_mod.mk` — `generate-release-notes` target**

```bash
export GITHUB_TOKEN=$(gh auth token)
make generate-release-notes        # uses cert_manager_latest_version from variables.json
make generate-release-notes CERT_MANAGER_VERSION=v1.21.0
```

**`content/docs/contributing/release-process.md`** — updated to document the new tool and Makefile target.

**`content/docs/releases/release-notes/release-notes-1.21.md`** — first real use of the tool: generated the v1.21.0 pre-release notes from 216 commits since v1.20.0.

## Testing

Unit tests cover:
- Creating a new versioned changelog section from an unversioned placeholder
- Inserting a new version above existing versioned sections (newest-first ordering)
- Replacing an existing section for the same version (handles pre-release → final transitions)
- Legacy files: inserting community markers via structural heuristics
- Legacy files: replacing a bare `## \`vX.Y.Z\`` heading with generated content on first run
- Contributor extraction from changelog handles (`[@handle]` and `` [`@handle`] `` forms)

Manually verified by running `make generate-release-notes CERT_MANAGER_VERSION=v1.21.0` against the existing `release-notes-1.21.md` file that was created manually — the tool inserted the community markers and generated the full changelog, including PRs merged up to the time of the run.

PR #2291, opened 2026-09-15
v1.20.4: patch release notes

wallrj:
**Preview:**
- https://deploy-preview-2291--cert-manager.netlify.app/docs/releases/release-notes/release-notes-1.20#v1204

## Summary

- Add v1.20.4 release notes to `release-notes-1.20.md`. This is a CVE-only release, so the notes are written by hand: only #9315 and #9152 carry a release-note block, and the rest are Renovate and base-image bumps that the generator from #1785 skips
- Add a section explaining the two `golang.org/x/crypto/ssh` findings (CVE-2026-56855, CVE-2026-78662) and GO-2026-5932 that image scanners will still report on v1.20.4, why they do not affect cert-manager, and why we are not raising the Go language version in a patch release to take the fix
- `variables.json` is unchanged. v1.21.2 stays `cert_manager_latest_version`
- Add `multiplexer` and `openpgp` to `.spelling`

## Motivation

Part of the cert-manager v1.20.4 release process.

- Tracking issue: cert-manager/cert-manager#9357

## Test plan

- [ ] Netlify deploy preview renders the v1.20.4 section correctly
- [x] `cspell` passes on `release-notes-1.20.md`

/hold until v1.20.4 is published

[Claude Fable 5.1]


wallrj-cyberark:
**I am self-approving this and merging it without a second reviewer.** Flagging that openly rather than doing it quietly.

v1.20.4 is already published. The images, the GitHub release and the Helm chart are all live, so these docs are the last piece and every hour they are not merged is an hour the installation page points at v1.20.3. The change is release notes for a release that has already shipped, so a reviewer could only check it against artefacts I have already verified. That did not seem worth another maintainer's time. Happy to take post-merge review comments.

What I checked before approving:

- **The three residual findings are real and correctly described.** A trivy scan of the five published v1.20.4 images, with `--ignorefile /dev/null` so the repository's own `.trivyignore` is not applied, reports exactly CVE-2026-56855, CVE-2026-78662 and GO-2026-5932, all `golang.org/x/crypto` v0.55.0. Nothing extra, nothing missing.
- **"Only the controller links the `ssh` package" is accurate.** `go list -deps` on each binary module shows `golang.org/x/crypto/ssh` in `cmd/controller` alone, imported by `github.com/Venafi/vcert/v5/pkg/util`. cainjector, startupapicheck and webhook pull in x/crypto but not the `ssh` package; acmesolver has no x/crypto at all. Trivy flags four binaries because it matches on module metadata, not linked packages.
- **"`govulncheck` confirms the vulnerable functions are not called" is backed by a run.** [Run 34996988453](https://github.com/cert-manager/cert-manager/actions/runs/34996988453) against `release-1.20` at 0b9448974 scanned all eight Go modules and reported no vulnerabilities in every one.
- The changelog entries match the PRs merged to `release-1.20` since v1.20.3, and the attributions are correct.

One fix pushed in fefab8ea: the section heading said "two findings" where the body lists three. The published GitHub release notes say three, so the heading was the odd one out.

Deliberately not changed: `content/docs/variables.json` stays at v1.21.2, because v1.20.4 is a patch on the older supported minor and is not the latest release. Separately, `content/v1.20-docs/variables.json` still reads v1.20.3 and has not been touched since the 1.20 docs were frozen. Worth deciding later whether the frozen archive should track patch releases, since as it stands those pages hand people a version with 15 HIGH findings. Out of scope here.

